PR #1, opened 2017-11-01
Adds ranking table below the sunburst charts

alannahw:
* to display the top five offenders from the usage distribution
* made arc responsive to hover for visibility
* moved metrics number from under sunburst to center of diagram

### DIFF
--- a/promvt/src/Sunburst.css
+++ b/promvt/src/Sunburst.css
@@ -31,32 +31,16 @@
   width: 100%;
   opacity: 0.7;
   text-align: center;
-  border-top: 1px solid #ccc;
-  border-bottom: 1px solid #ccc;
-}
-.sunburst .count {
-  text-align: center;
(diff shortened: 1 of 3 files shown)


davkal:
The code to the generate the tables is great.

The CSS needs adapting though: This project uses [basscss](http://basscss.com/) classes to express styles, and only does minimal adjustments with local classes (ideally there would be zero). The goal is maximum reusability in cases where sub-components are included into other projects. The fewer explicit styles are in this repo, the more freedom the super-project has in laying things out. 

Here is a screenshot on how the table looks when included in a super-project:

screen shot 2017-11-02 at 15 29 09

Concretely, I would ask you to look through the CSS changes of this PR and see if they can also be done via bass classes. Let's take it from there.

Last thing: there needs to be a separation between the label and the metric name in each table row.

PR #32, opened 2018-01-30
add support for multiple queries in Grafana queryPanel

msiuts:
This adds support for multiple queries, like it was already supported in the tablePanel.


davkal:
Works like a charm! Thanks @msiuts

PR #33, opened 2018-02-01
make --web.route-prefix configurable

msiuts:
In order to be able to use prometheus via kubectl proxy you need to configure
the `web.route-prefix`.

Example:
```
prometheus_external_hostname: "http://localhost:8001",
prometheus_path: "/api/v1/namespaces/default/services/prometheus:80/proxy/",
prometheus_web_route_prefix: "/",
```

### DIFF
--- a/prometheus-ksonnet/lib/prometheus.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus.libsonnet
@@ -5,7 +5,7 @@ k {
     prometheus_external_hostname: "http://prometheus.%s.svc.cluster.local" % $._config.namespace,
     prometheus_path: "/",
     prometheus_port: 80,
-    prometheus_web_route_prefix: "/",
+    prometheus_web_route_prefix: $._config.prometheus_path,
   },
 
   local policyRule = $.rbac.v1beta1.policyRule,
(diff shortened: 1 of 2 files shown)


prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -5,6 +5,7 @@ k {
     prometheus_external_hostname: "http://prometheus.%s.svc.cluster.local" % $._config.namespace,
     prometheus_path: "/",
     prometheus_port: 80,
+    prometheus_web_route_prefix: "/",
```
tomwilkie: To make this compatible with what we do, could you set this to default to `$._config.prometheus_path`?

prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -40,7 +42,7 @@ k {
     container.withArgs([
       "-v", "-t", "-p=/etc/prometheus",
       "curl", "-X", "POST", "--fail", "-o", "-", "-sS",
-      "http://localhost:%s%s-/reload" % [$._config.prometheus_port, $._config.prometheus_path],
+      "http://localhost:%s/-/reload" % [$._config.prometheus_port],
```
tomwilkie: Instead of removing this, could you set it to `$._config.prometheus_web_route_prefix` (and remove the slash)?  That should have the same effect.

tomwilkie:
Thanks @msiuts!  It looks good, couple of tweaks needed to make it compatible with us too.

msiuts:
I applied the suggestions, but I also discovered that the scrapping now tries to get the prometheus metrics via the proxy which is not allowed. So I will need to do some changes there too before you merge. So this is currently WIP.

msiuts:
I used the web_route_prefix also for scrapping and hope that works for your environments too.

msiuts:
Is this now compatible to your use case @tomwilkie ?

tomwilkie:
Perfect, thanks @msiuts

PR #36, opened 2018-02-19
klumps: fix recording rule record name, namespace_name

accnops:
The record name prefix "namespace" does not match with the prefix used in other records in the same set "namespace_name" . Additionally, in the Grafana dashboards "namespace_name" is used instead of "namespace". I guess this was a simple typo.

### DIFF
--- a/klumps/recording_rules.jsonnet
+++ b/klumps/recording_rules.jsonnet
@@ -12,13 +12,13 @@
       name: "k8s.rules",
       rules: [
         {
-          record: "namespace:container_cpu_usage_seconds_total:sum_rate",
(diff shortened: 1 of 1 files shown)


tomwilkie:
Hi Arthur - not sure this is right.  There is already a `namespace_name:container_cpu_usage_seconds_total:sum_rate` rule defined in the file, and the one you've changed really does only aggregate by namespace (the prefix on recording rules is used to indicate the labels on the output - see https://prometheus.io/docs/practices/rules/#naming-and-aggregation).

accnops:
Didn't see they already existed! Trying to add KLUMPS, but most of record rules don't get data...

PR #39, opened 2018-03-15
Update alertmanager to 0.14.0

msiuts:


tomwilkie:
Wow we were running an old version...

Thanks!

PR #41, opened 2018-03-23
Amend KubePodsCrashing Looping Alert

jdbaldry:
- Should use kube_pod_container_status_restarts_total
- https://github.com/kubernetes/kube-state-metrics/blob/master/Documentation/pod-metrics.md


tomwilkie:
Thanks Jack!

PR #42, opened 2018-03-26
Make Prometheus Resources Limits configurable

WolfgangMau:
since we discovered a usage of 87% in K8S , we  needed to increase the Limits.


tomwilkie:
Hi Wolfgang; with ksonnet its not necessary to expose every configuration variable .  You can do something like this in your main.jsonnet to simply override the things you don't like:

```
local prometheus = import "prometheus-ksonnet/prometheus-ksonnet.libsonnet";

prometheus {
  prometheus_container+::
     $.util.resourcesRequests("1", "2Gi") +
     $.util.resourcesLimits("2", "4Gi"),
}
```

This is the whole beauty of jsonnet!

PR #45, opened 2018-03-27
run node-exporter as root

msiuts:
This fixes

```
Error on statfs() system call for "/rootfs/var/lib/kubelet/pods/XXXXX/volumes/XXXXXXX": permission denied"
```
on newer kernel versions.


tomwilkie:
LGTM!  Thanks

PR #46, opened 2018-03-28
Grafana admin password as parameter

WolfgangMau:
makes it possible to overwrite the grafana admin-password


davkal:
The change looks good to me. 

The conflict is about the hostname of the webserver: we've changed this to `nginx` because that will do the routing based on the path `/grafana` instead of the host.

Will this work for you still @WolfgangMau ?

tomwilkie:
This has gone very stale.

PR #51, opened 2018-04-05
Update volume mount jsonnet, to avoid  diffs.

tomwilkie:
Also, bring in some updates and cleanups from Grapeshot.

@jdbaldry PTAL


jdbaldry:
Looks good, love the new diffs

PR #53, opened 2018-04-12
Set homepage to "." to allow deployment at any URL

dobesv:
This makes it so `yarn build` generates HTML with relative paths to the css / js files, which makes it easier to deploy somewhere without running the webpack-dev-server.

### DIFF
--- a/promvt/package.json
+++ b/promvt/package.json
@@ -6,6 +6,7 @@
   "repository": "https://github.com/kausalco/public",
   "bugs": "https://github.com/kausalco/public/issues",
   "proxy": "http://localhost:9090",
+  "homepage": ".",
   "dependencies": {
     "d3-color": "^1.0.3",
     "d3-hierarchy": "^1.1.5",


davkal:
This is going to mess with any feature additions that rely on browser history, e.g.: `/details/foo`, because then it will try loading the assets from `/details/static...`.
I'd rather that this gets baked in at build time from an environment variable.

dobesv:
Fair point.

dobesv:
By the way, is there currently a way to specify this using an environment variable?  That would be very handy.

davkal:
This project uses create-react-app, which uses react-scripts under the hood. I've taken a quick look and perhaps there is a way via setting `PUBLIC_URL`, see
https://github.com/facebook/create-react-app/blob/next/packages/react-scripts/config/paths.js#L34

Have not tried it yet.

dobesv:
It works.  Thanks!

PR #54, opened 2018-04-12
Include credentials in prometheus API fetches

dobesv:
I deployed promvt as a subfolder of the same domain I access prometheus from.  That subdomain was protected by an authentication mechanism.  I had to enable authentication on this `fetch` call so that promvt could load from the prometheus API from my browser.


davkal:
Thanks, no reason not to include those.

PR #57, opened 2018-04-21
Use std.manifestIni to the Grafana config is extensible.

tomwilkie:
@sparkprime finally got round to it!

@msiuts we discussed this a few days ago.


sparkprime:
LGTM :)

PR #58, opened 2018-05-08
Add configVolumeMountSubPath util function

jdbaldry:
Gives user possibility of mounting configmap to a file in a directory rather than mounting the whole directory. e.g to /etc/config/config.file rather than to /etc/config/.

Is this something you would want? Would the better implementation be to have logic in configVolumeMount that uses subpath if there is a third argument (I'm not sure yet how to do that with jsonnet)?

### DIFF
--- a/prometheus-ksonnet/lib/kausal.libsonnet
+++ b/prometheus-ksonnet/lib/kausal.libsonnet
@@ -284,11 +284,6 @@ k {
         volume.fromEmptyDir(name),
       ]),
 
-      deployment.mapContainers(addMount) +
-      deployment.mixin.spec.template.spec.withVolumesMixin([
-        volume.fromEmptyDir(name),
-      ]),
-
     manifestYaml(value):: (
       local f = std.native('manifestYamlFromJson');
       f(std.toString(value))
--- a/prometheus-ksonnet/lib/kausal.libsonnet
+++ b/prometheus-ksonnet/lib/kausal.libsonnet
@@ -233,6 +233,21 @@ k {
       deployment.mixin.spec.template.spec.withVolumesMixin([
         volume.fromConfigMap(name, name),
       ]),
+      
+    configVolumeMountSubPath(name, path, subpath)::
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/lib/kausal.libsonnet
```diff
@@ -233,6 +233,21 @@ k {
...
       ]),
+      
+    configVolumeMountSubPath(name, path, subpath)::
+      local container = $.core.v1.container,
+            deployment = $.extensions.v1beta1.deployment,
+            volumeMount = $.core.v1.volumeMount,
+            volume = $.core.v1.volume,
+            addMount(c) = c + container.withVolumeMountsMixin(
+        volumeMount.new(name, path) +
+        volumeMount.withSubPath(subpath)
```
tomwilkie: You could add an optional param to `configVolumeMount(..., volumeMountMixin={})` and then pass `withSubPath` there.  Not sure thats better though.  WDYT?

tomwilkie: Sorry for delay!  If you have a minute, make it a optional parameter.  If not let me know.

jdbaldry:
Implemented an optional mixin solution that works for me. Does it look okay to you?

prometheus-ksonnet/lib/kausal.libsonnet
```diff
@@ -263,20 +269,26 @@ k {
...
 
       deployment.mapContainers(addMount) +
       deployment.mixin.spec.template.spec.withVolumesMixin([
         volume.fromEmptyDir(name),
       ]),
 
+      deployment.mapContainers(addMount) +
+      deployment.mixin.spec.template.spec.withVolumesMixin([
+        volume.fromEmptyDir(name),
+      ]),
```
tomwilkie: This look like its been accidentally repeated here @jdbaldry?

tomwilkie:
Thanks Jack!

PR #60, opened 2018-05-25
add hostname as instance label for node exporter

jtlisi:


prometheus-ksonnet/lib/prometheus-config.libsonnet
```diff
@@ -164,6 +164,39 @@
...
+        relabel_configs: [
+
+          // Drop anything who's service is not kube-state-metrics
+          // Rename jobs to be <namespace>/<name, from pod name label>
+          {
+            source_labels: ['__meta_kubernetes_namespace', '__meta_kubernetes_pod_label_name'],
+            separator: '/',
+            regex: '%s/node-exporter' % $._config.namespace,
+            action: 'keep',
+          },
```
tomwilkie: You can limit the SD to just one namespace  (see https://prometheus.io/docs/prometheus/latest/configuration/configuration/#%3Ckubernetes_sd_config%3E), and then make this just match on name.

tomwilkie:
One nit, otherwise LGTM!

prometheus-ksonnet/lib/prometheus-config.libsonnet
tomwilkie: If you fancy doing that with kube-state-metrics too, that'd be awesome.

tomwilkie:
(Merged in manually with a few fixes.)

PR #64, opened 2018-06-20
make name label default for statefulSet consistent with deployment

jtlisi:


jtlisi:
@tomwilkie when you get a chance can you look at this. I was trying to figure out why statefulsets of prometheus weren't scraping themselves and realized they are dropped because they don't have the name label.

tomwilkie:
LGTM

PR #67, opened 2018-06-26
update node exporter job to keep instance label as pod name and node …

jtlisi:
This PR reverts the instance label on the node exporter job to the pod name and adds a node label with the name of the node. This fixes the following rule which currently is not being evaluated correctly.

```
{
  record: 'node:node_num_cpu:sum',
  expr: |||
    count by (node) (sum by (node, cpu) (
      node_cpu{%(nodeExporterSelector)s}
    * on (namespace, %(podLabel)s) group_left(node)
      node_namespace_pod:kube_pod_info:
   ))
}
```

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -198,11 +198,18 @@
             action: 'keep',
           },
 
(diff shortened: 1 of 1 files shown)


jtlisi:
@tomwilkie could you take a look at this when you have a chance

PR #74, opened 2018-07-24
Add option for best effort anti-affinity

gouthamve:
Most applications don't require strict anti-affinity, only best effort. If I were to use the strict version on a local minikube setup, I wouldn't be able to deploy new version.


tomwilkie:
This has gone very stale.

Plus, kubernetes does best effort anti affinity to begin with.

PR #83, opened 2018-09-26
All statefulsets should use RollingUpdate by default.

tomwilkie:

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -125,7 +125,8 @@ k {
 
     statefulSet+: {
       new(name, replicas, containers, volumeClaims, podLabels={})::
-        super.new(name, replicas, containers, volumeClaims, podLabels { name: name }),
+        super.new(name, replicas, containers, volumeClaims, podLabels { name: name }) +
+        super.mixin.spec.updateStrategy.withType("RollingUpdate"),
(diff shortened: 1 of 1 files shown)


ksonnet-util/kausal.libsonnet
```diff
@@ -125,7 +125,8 @@ k {
 
     statefulSet+: {
       new(name, replicas, containers, volumeClaims, podLabels={})::
-        super.new(name, replicas, containers, volumeClaims, podLabels { name: name }),
+        super.new(name, replicas, containers, volumeClaims, podLabels { name: name }) +
+        super.mixin.spec.updateStrategy.withType("RollingUpdate"),
```
woodsaj: If this was not already the default, this would be a breaking change for many users.

tomwilkie: https://kubernetes.io/docs/tutorials/stateful-application/basic-stateful-set/#updating-statefulsets

> RollingUpdate update strategy is the default for StatefulSets.

I don't know why its not being set by default for us, but I thinks its because we're on 1.8.  I suspect most others users (if there are any..) are surprised by this, like we are.

tomwilkie: Its seems to think thats the default, but I can't find where its setting it.

PR #94, opened 2018-11-26
remove config.file from container.withArgs for prometheus

cstyan:
The Prometheus Dockerfile was changed recently (prometheus/prometheus#4796) to use `Entrypoint` only instead of `Entrypoint` and `Cmd`. This meant that args we passed in here were appended to the end of the Prometheus command inside the container, rather than overriding the defaults that were there previously via `Cmd`, so `--config.file` was being passed twice.

We don't pass a value for config file that's any different than the default anyways.

cc @tomwilkie 


cstyan:
looks like @gouthamve fixed this :)

PR #95, opened 2018-11-28
update cadvisor scrape config

woodsaj:
Since k8s 1.7.3 the cadvisor metrics are available via kubelet using the /metrics/cadvisor path.  From k8s 1.11 cadvisor metrics are no longer available on port :4194

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -246,12 +246,13 @@
         ],
       },
 
-      // As of k8s 1.7, cAdvisor is on a port 4194.
+      // As of k8s 1.7.3, cAdvisor metrics are available via kubelet using the /metrics/cadvisor path
       {
         job_name: 'kube-system/cadvisor',
         kubernetes_sd_configs: [{
           role: 'node',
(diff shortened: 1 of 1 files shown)


tomwilkie:
We need to make this optional, so this doesn't break our non-upgraded clusters.  Let me know if you want more exposition.

woodsaj:
As indicated in the comment, this will work on k8s >= 1.7.3
The oldest version we have running in is 1.8.8

I have tested these older nodes to ensure that metrics are returned when sending a  query to `https://kubernetes.default.svc/api/v1/nodes/$NODE/proxy/metrics/cadvisor` 

eg.  
```
$ kube proxy &
$ curl -s http://localhost:8001/api/v1/nodes/$NODE/proxy/metrics/cadvisor |grep version 
# HELP cadvisor_version_info A metric with a constant '1' value labeled by kernel version, OS version, docker version, cadvisor version & cadvisor revision.
# TYPE cadvisor_version_info gauge
cadvisor_version_info{cadvisorRevision="",cadvisorVersion="",dockerVersion="17.03.2-ce",kernelVersion="4.4.111+",osVersion="Container-Optimized OS from Google"} 1
```

PR #96, opened 2018-12-06
Prometheus puts config file in the entrypoint now

gouthamve:
https://github.com/prometheus/prometheus/blob/master/Dockerfile#L22

Repeated flags means Prometheus will fail to start. This should not
break anything as the flag value is the default in the Dockerfile also

Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>


tomwilkie:
Do you need to update the image version too?

gouthamve:
Nope, not required. It's still just rc now.

tomwilkie:
Ah, so probably don't merge this then?  Otherwise anyone trying to use it will break, no?

PR #98, opened 2018-12-08
Revert "Prometheus puts config file in the entrypoint now"

gouthamve:
Reverts grafana/jsonnet-libs#96

Ughh, I don't know what I was thinking back then. This is why I should stop working when tired. This does break things when we deploy using `2.6`


tomwilkie:
Lgtm

PR #99, opened 2018-12-19
WIP Use __meta_kubernetes_pod_controller_name instead of pod name label.

tomwilkie:
This means we don't need to rely on there being a name label on the pods, should be more generic across different clusters.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -97,16 +97,9 @@
             replacement: '$1:$3',
           },
 
-          // Drop pods without a name label
+          // Rename jobs to be <namespace>/<name, pod controller name>
           {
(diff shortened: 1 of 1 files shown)


tomwilkie:
(Needs testing)

tomwilkie:
@beorn7 I'd like to resurrect this and unify it with the scrape config we use for Loki: https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet#L4

beorn7:
Did you do any testing? Was anything not working?

tomwilkie:
I didn't yet, no.  Plus its really out of date with the Loki config, where a bunch of effort went into dealing with replica sets etc.

beorn7:
Is this still feasible @tomwilkie ?

tomwilkie:
It would be nice to unify the prometheus and loki scrape configs, but I doubt this PR is useful for that, its so out of date.

PR #100, opened 2019-01-08
Initial Jaeger mixin 

gouthamve:
Signed-off-by: Goutham Veeramachaneni <gouthamve@gmail.com>

### DIFF
--- a/jaeger-mixin/alerts.libsonnet
+++ b/jaeger-mixin/alerts.libsonnet
@@ -1,0 +1,140 @@
+local percentErrs(metric, errSelectors) = '100 * sum(rate(%(metric)s{%(errSelectors)s}[1m])) by (instance, job, namespace) / sum(rate(%(metric)s[1m])) by (instance, job, namespace)' % {
+  metric: metric,
+  errSelectors: errSelectors,
+};
+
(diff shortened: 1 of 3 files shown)


jaeger-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,140 @@
+local percentErrs(metric, errSelectors) = '100 * sum(rate(%(metric)s{%(errSelectors)s}[1m])) by (instance, job, namespace) / sum(rate(%(metric)s[1m])) by (instance, job, namespace)' % {
+  metric: metric,
+  errSelectors: errSelectors,
+};
```
tomwilkie: Why `by (instance)`?

gouthamve: Included `instance` because of the loadbalancing, I think. Because the balancing is not uniform, we might be losing all the traces from an instance but won't know it if it's getting a small number of traces.

gouthamve: Well if one instance is dropping them, I don't think there is any way to indicate backpressure for example. And there is a `1 <--> 1` connection b/w agent and collector. There is no HA there, if the collector is dropping some spans, the agent doesn't change anything, it just keeps sending the spans to the collector. Hopefully that'll change with gRPC but that's not implemented as of now.

tomwilkie:
Merge for now, we'll have to revisit how HA our jaeger setup really is later.

tomwilkie:
@gouthamve ping

PR #103, opened 2019-01-16
Add common proxy headers to nginx location stanzas

jdbaldry:


prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -65,7 +65,7 @@
   local pvc = $.core.v1.persistentVolumeClaim,
 
   prometheus_pvc::
-    if ! $._config.stateful
+    if !$._config.stateful
```
tomwilkie: I believe this is a syntax error introduce by an old version of jsonnet fmt.  See https://github.com/google/jsonnet/issues/572

There is a workaround using brackets on that issue.

tomwilkie:
Thanks for the PR @jdbaldry! One issue otherwise LGTM.

jdbaldry:
No problem @tomwilkie. Thanks for the tip. I probably should change my workflow from running a `make fmt` at the end without retesting. Should be all good now.

tomwilkie:
Thanks Jack!  Sorry for being slow.

PR #115, opened 2019-03-25
Better jaeger alerts

gouthamve:
This expands on the alerts we already have.

### DIFF
--- a/jaeger-mixin/alerts.libsonnet
+++ b/jaeger-mixin/alerts.libsonnet
@@ -14,7 +14,19 @@ local percentErrsWithTotal(metric_errs, metric_total) = '100 * sum(rate(%(metric
       {
         name: 'jaeger_alerts',
         rules: [{
-          alert: 'JaegerHTTPServerErrs',
+          alert: 'JaegerAgentUDPPacketsBeingDropped',
+          expr: 'rate(jaeger_agent_thrift_udp_server_packets_dropped_total[1m]) > 1',
+          'for': '15m',
+          labels: {
+            severity: 'warning',
+          },
+          annotations: {
+            message: |||
+              {{ $labels.job }} {{ $labels.instance }} is dropping {{ printf "%.2f" $value }} UDP packets per second.
+            |||,
+          },
+        }, {
+          alert: 'JaegerAgentHTTPServerErrs',
           expr: percentErrsWithTotal('jaeger_agent_http_server_errors_total', 'jaeger_agent_http_server_total') + '> 1',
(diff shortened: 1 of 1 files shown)


jaeger-mixin/alerts.libsonnet
```diff
@@ -14,27 +14,27 @@ local percentErrsWithTotal(metric_errs, metric_total) = '100 * sum(rate(%(metric
       {
         name: 'jaeger_alerts',
         rules: [{
-          alert: 'JaegerHTTPServerErrs',
-          expr: percentErrsWithTotal('jaeger_agent_http_server_errors_total', 'jaeger_agent_http_server_total') + '> 1',
+          alert: 'JaegerAgentUDPPacketsBeingDropped',
+          expr: 'rate(jaeger_agent_thrift_udp_server_packets_dropped_total[1m]) > 1',
```
tomwilkie: Can this be done as a % of total packed?  I don't like absolute thresholds.

jaeger-mixin/alerts.libsonnet
```diff
@@ -14,27 +14,27 @@ local percentErrsWithTotal(metric_errs, metric_total) = '100 * sum(rate(%(metric
...
             message: |||
-              {{ $labels.job }} {{ $labels.instance }} is experiencing {{ printf "%.2f" $value }}% HTTP errors.
+              {{ $labels.job }} {{ $labels.instance }} is dropping {{ printf "%.2f" $value }} UDP packets per second.
             |||,
           },
         }, {
-          alert: 'JaegerRPCRequestsErrors',
-          expr: percentErrs('jaeger_client_jaeger_rpc_http_requests', 'status_code=~"4xx|5xx"') + '> 1',
+          alert: 'JaegerAgentHTTPServerErrs',
+          expr: percentErrsWithTotal('jaeger_agent_http_server_errors_total', 'jaeger_agent_http_server_total') + '> 1',
```
tomwilkie: IMO its borderline whether the benefits of functions like this are worth it.  Makes queries more consistent vs not obvious what the actual query is.

jaeger-mixin/alerts.libsonnet
gouthamve: The main reason I did it this way because with UDP server dropping, it's because the UDP queue is full, and just checking for dropping makes more sense than a percentage here.

I'd ideally like to know packets are being dropped vs if only `>1%` of the packets are being dropped no? Atleast for the case of UDP?

tomwilkie:
JFDI

PR #118, opened 2019-04-11
Add _config in to prometheus.libsonnet 

cstyan:
Add `_config` the same way we add `prometheus_config` and other objects so that we can overwrite
certain fields when creating new prometheus instances, such as
changing the web prefix and hostname to differentiate between
instances with different purposes.

@jtlisi note that I also added `_config` in to the watch container so that we have the right port and web route prefix there if they've been overwritten for the prometheus container.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus.libsonnet
@@ -78,7 +78,9 @@
     local deployment = $.apps.v1beta1.deployment,
 
     prometheus_deployment:
-      if $._config.stateful
+      local _config = self._config;
+
+      if _config.stateful
       then {}
       else (
         deployment.new(self.name, 1, [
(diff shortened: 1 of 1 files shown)


jtlisi:
@cstyan you need to add it to the statefulset deploy as well
```
    prometheus_statefulset:
      if !($._config.stateful)
      then {}
      else (
        statefulset.new(self.name, 1, [
          self.prometheus_container.withVolumeMountsMixin(
            volumeMount.new('%s-data' % self.name, '/prometheus')
          ),
          self.prometheus_watch_container,
        ], self.prometheus_pvc) +
        $.util.configVolumeMount('%s-config' % self.name, '/etc/prometheus') +
        statefulset.mixin.spec.withServiceName('prometheus') +
        statefulset.mixin.spec.template.metadata.withAnnotations({ 'prometheus.io.path': '%smetrics' % $._config.prometheus_web_route_prefix }) +
        statefulset.mixin.spec.template.spec.securityContext.withRunAsUser(0) +
        if $._config.enable_rbac
        then statefulset.mixin.spec.template.spec.withServiceAccount(self.name)
        else {}
      ),
```

PR #119, opened 2019-04-12
helper to set up cron job with proper name label on pod

DanCech:
This maintains backward compatibility with the standard library, but sets the name down into the pod labels if it is specified.  It also accepts schedule and container(s) similar to the other helpers.

Not sure whether we'd like to be a bit more prescriptive about defaults for things like successful/failed jobs history or parallelism?

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -88,6 +88,30 @@ k {
     },
   },
 
+  batch+: {
+    v1beta1+: {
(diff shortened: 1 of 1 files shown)


tomwilkie:
I'd probably make name a required parameter (for consistency with others).  This will break backwards compatibility, but we've changes constructor args in this overlay in the past.  Don't feel strongly about this though, will leave it up to you.  LGTM.

PR #120, opened 2019-04-12
update alertmanager image to 16.2

jtlisi:
Signed-off-by: Jacob Lisi <jacob.t.lisi@gmail.com>


cstyan:
Don't know if there's a reason we've explicitly not upgrade AM yet, otherwise lgtm.

tomwilkie:
Minor changes like this should be safe. LGTM.

PR #122, opened 2019-04-17
Support Grafana 6.1 Notification channel provisioning

malcolmholmes:
This allows for the provisioning of notification channels as well as dashboards/alerts.

### DIFF
--- a/prometheus-ksonnet/lib/grafana.libsonnet
+++ b/prometheus-ksonnet/lib/grafana.libsonnet
@@ -101,6 +101,32 @@
 
   grafanaNotificationChannels+:: {},
 
+  /*
+    to add a notification channel:
+
+    grafanaNotificationChannels+:: {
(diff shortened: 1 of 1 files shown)


woodsaj:
I really dont think that "**prometheus**-ksonnet" needs support for adding non-prometheus datasources.
The notification-channels make sense as they require their own configMap and you might want to use Grafana alerting (eg to monitor prometheus and alert manager)

Right now, the non-prometheus support is not even used.  We just provision our graphite datasources with:
```
  // Tell Grafana about metrictank DS.
  grafana_datasource_config_map+:
    configMap.withDataMixin({
      'metrictank.yml': $.util.manifestYaml({
        apiVersion: 1,
        datasources: [{
          name: 'REDACTED',
          type: 'graphite',
          access: 'proxy',
....
```

PR #136, opened 2019-06-13
Don't let diff show empty configmaps as differences

malcolmholmes:
Without this change, we see a diff if a configmap is empty. This prevents change ksonnet from creating empty configmap entries, and keeps ksonnet diff happy.

Will this create any issues for applications that expect a config map entry to be present, but empty?

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -12,6 +12,12 @@ k {
       configMap+: {
         new(name)::
           super.new(name, {}),
+        withData(data)::
+          if (data == {}) then {}
+          else super.withData(data),
+        withDataMixin(data)::
+          if (data == {}) then {}
+          else super.withData(data),
       },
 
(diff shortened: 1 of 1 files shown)


ksonnet-util/kausal.libsonnet
```diff
@@ -12,6 +12,12 @@ k {
       configMap+: {
         new(name)::
           super.new(name, {}),
+        withData(data)::
+          if (data == {}) then {}
+          else super.withData(data),
+        withDataMixin(data)::
+          if (data == {}) then {}
+          else super.withData(data),
```
woodsaj: shouldn't this be `super.withDataMixin(data)` ?

malcolmholmes: eeks, thanks for spotting @woodsaj .

malcolmholmes: fix pushed

PR #137, opened 2019-06-25
Add option to rewrite paths in nginx for services that don't support serving from a subpath.

tomwilkie:

### DIFF
--- a/prometheus-ksonnet/lib/nginx.libsonnet
+++ b/prometheus-ksonnet/lib/nginx.libsonnet
@@ -12,13 +12,13 @@
             proxy_set_header    Upgrade $http_upgrade;
             proxy_set_header    Connection "Upgrade";
     ||| else '' + if subfilter then |||
-      sub_filter 'href="/' 'href="/%(path)s/';
-      sub_filter 'src="/' 'src="/%(path)s/';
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/lib/nginx.libsonnet
```diff
@@ -11,13 +11,25 @@
             # Allow websocket connections https://www.nginx.com/blog/websocket-nginx/
             proxy_set_header    Upgrade $http_upgrade;
             proxy_set_header    Connection "Upgrade";
-    ||| else '',
+    ||| else '' + if subfilter then |||
+      sub_filter 'href="/' 'href="/%(path)s/';
+      sub_filter 'src="/' 'src="/%(path)s/';
+      sub_filter 'endpoint:"/' 'endpoint:"/%(path)s/';  # for XHRs.
```
woodsaj: i dont think that a pagerbeuty specific filter should be in this ksonnet library.
Can we change the lib so that you can just pass a custom nginx config snippet instead?

tomwilkie: These aren't pagerbeautry specific (even if thats the reason we added them).  This should work for anything using react AFAICT.

PR #138, opened 2019-06-27
Hide main_prometheus since we're optionally deploying replicas now.

cstyan:
Since we're deploying replicas now, and only deploying main_prometheus if the config is set to not deploy replicas, main_prometheus should be hidden by default.


tomwilkie:
Erm, if we hire this it won't get deployed?

PR #147, opened 2019-07-18
Add critical pod priority to monitoring pods.

tomwilkie:


ksonnet-util/kausal.libsonnet
```diff
@@ -391,5 +392,11 @@ k {
           podAntiAffinity.requiredDuringSchedulingIgnoredDuringExecutionType.withTopologyKey('kubernetes.io/hostname'),
         ]).spec,
       },
+
+    podPriority(p):
```
beorn7: Perhaps add a comment that this only sets the pod priority if `enable_pod_priorities` is set to `true`.

PR #149, opened 2019-07-24
Add node-mixin to prometheus-ksonnet

beorn7:
There is a non-empty intersection between the kubernetes-mixin and the
node-mixin in terms of rules and dashboards. However, the names are
actually all different (kind of deliberate for the dashboards - for
the rules we are "lucky" that the kubernetes-mixin is totally not
following the Prometheus rule naming conventions, while the node-mixin
is by now).

So this should just play well along each other for now. Of course, we
want to deduplicate the more or less similar dashboards eventually. We
still have to figure out on which level this should happen. We either
kick out some dashboards here in prometheus-ksonnet or, probably
better, make upstream kubernetes-mixin use the node-mixin instead of
doing their own stuff.

Historically, I guess, the node-mixin parts that overlap with the
kubernetes-mixin were copied over from or at least inspired by the
kubernetes-mixin. It probably makes a lot of sense to let the
kubernetes-mixin use the node-mixin maintained in the
prometheus/node_exporter repo directly.

But let's first play with it and then make that call later.

### DIFF
--- a/prometheus-ksonnet/jsonnetfile.json
+++ b/prometheus-ksonnet/jsonnetfile.json
@@ -31,6 +31,16 @@
             "version": "mixin"
         },
         {
+            "name": "node-mixin",
+            "source": {
+                "git": {
+                    "remote": "https://github.com/prometheus/node_exporter",
+                    "subdir": "docs/node-mixin"
+                }
+            },
+            "version": "master"
+        },
+        {
             "name": "oauth2_proxy",
             "source": {
                 "git": {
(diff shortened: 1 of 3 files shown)


beorn7:
NB: That will be a lot of layers if indirection:

- The local config/ksonnet setup imports prometheus-ksonnet.
- prometheus-ksonnet imports kubernetes-mixin.
- kubernetes-mixin imports node-mixin.

And there are some more questions, but let's discuss that later. I should stop brainstorming here. This is just about getting the node-mixin going. Once we get a feeling how it works, we can tackle upstreaming to the kubernetes-mixin.

beorn7:
@cstyan perhaps you can rubberstamp^Wreview this while Tom is gone?

PR #156, opened 2019-08-10
Only deploy Prometheus as statefulset, deploy replicas.

cstyan:
This PR removes the ability to deploy Prometheus as a deployment, it will always deploy Prom as a statefulset. Also adds Prometheus replicas/removes `main_prometheus`.

If we want I can make it so the statefulset/replicas are default but deployment/`main_prometheus` are still available via config vars.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus.libsonnet
@@ -2,6 +2,15 @@
   prometheus:: {
     name:: error 'must specify name',
 
+    _config:: $._config,
+
+    // We bounce through various layers of indirection so user can:
+    // a) override config for all Prometheus' by merging into $.prometheus_config,
+    // b) override config for a specific Prometheus instance by merging in here.
+    prometheus_config:: $.prometheus_config,
+    prometheusAlerts:: $.prometheusAlerts,
+    prometheusRules:: $.prometheusRules,
+
     local policyRule = $.rbac.v1beta1.policyRule,
 
     prometheus_rbac:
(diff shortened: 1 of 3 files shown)


prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -77,71 +77,92 @@
       ]),
 
     local deployment = $.apps.v1beta1.deployment,
```
tomwilkie: This should be remove too, no?

prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -77,71 +77,92 @@
...
+      pvc.new() +
+      pvc.mixin.metadata.withName('%s-data' % (self.name)) +
+      pvc.mixin.spec.withAccessModes('ReadWriteOnce') +
+      pvc.mixin.spec.resources.withRequests({ storage: '300Gi' }),
 
     local statefulset = $.apps.v1beta1.statefulSet,
     local volumeMount = $.core.v1.volumeMount,
 
     prometheus_statefulset:
       local _config = self._config;
```
tomwilkie: I think this can be removed too.

tomwilkie:
I'd make this only do that stateful change, and move the replicas logic here in the next PR.  Small changes.

prometheus-ksonnet/lib/prometheus.libsonnet
tomwilkie: No it cannot.

PR #159, opened 2019-08-16
Optionally shard dashboards into multiple config maps to avoid size limits.

tomwilkie:


presidenten:
@tomwilkie Can I use this if I want to export a dashboard as json and create my own configmaps to be imported with the grafana sidecar?
Are there any instructions anywhere?

malcolmholmes:
@presidenten we don't do that. We mount these configmaps into a pod and then use Grafana provisioning to pull them in to Grafana.

Having said that, if the sidecar can consume a directory containing json dashboards, then there's no reason this logic couldn't be used.

PR #161, opened 2019-08-19
memcached: add more tunables

rfratto:
Adding more tunables to the upstream memcached.libsonnet so Loki and Cortex can share the same memcached libsonnet config.


rfratto:
@tomwilkie PTAL, doing this as part of the new Cortex changes.

PR #163, opened 2019-08-20
Update node_exporter to v0.18.1

beorn7:
This has been canaried in ops-tools1 already.


beorn7:
Should this first be rolled out to all prod environments before pushing it up to the library?
This has worked very smoothly in ops-tools1.

tomwilkie:
> Should this first be rolled out to all prod environments before pushing it up to the library?
This has worked very smoothly in ops-tools1.

Its a tricky one that we need a better answer to.  Lets merge this, update our vendor and deploy to 2-3 clusters, wait a few days, then deploy to all.  The outstanding diff is undesireable, but I can't think of a better way.

beorn7:
And I guess the right crowd of people to keep informed is #infra.

PR #165, opened 2019-08-20
Add another dashboard to override

beorn7:
I also used a different syntax. The effect on ks-diff seems to be the
same, but this syntax appears "more correct" to me.

Still, in the ks-diff, the `null` shows up as `"null"`. The jsonnet
tutorial promises that `null` elements will not be rendered. Hmmm...


tomwilkie:
Does tanka show them?

beorn7:
Yeah, same effect with tanka. I assume it will still work. Will roll back if not.

beorn7:
Now I have found the “proper” way, see #168 .

The promise about kicking out `null` fields is about the _name_ of the field, not the _value_.

PR #169, opened 2019-09-04
Add service name to alertmanager's statefulset

beorn7:
My K8s foo is too weak to be confident that this is the right
approach. Please doublecheck.

If it works, it should fix https://github.com/raintank/deployment_tools/issues/1455 .

### DIFF
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -79,6 +79,7 @@
       $.alertmanager_container,
       $.alertmanager_watch_container,
     ], self.alertmanager_pvc) +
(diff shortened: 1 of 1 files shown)


beorn7:
Will cautiously try it out in ops-tools1 first.

beorn7:
Result: This cannot simply be applied: `The StatefulSet "alertmanager" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', and 'updateStrategy' are forbidden`

I guess I have to delete the statefulset for alertmanager and then apply.

beorn7:
That seems to have worked. The new statefulset is even using the old PVC, i.e. pre-existing silences survived.

beorn7:
Merging and working on the vendor update.

PR #177, opened 2019-10-16
Defaulting to dark mode

joe-elliott:
Default to dark mode as it is the superior method of viewing graphs.

### DIFF
--- a/prometheus-ksonnet/lib/grafana.libsonnet
+++ b/prometheus-ksonnet/lib/grafana.libsonnet
@@ -13,7 +13,7 @@
         reporting_enabled: false,
       },
       users: {
-        default_theme: 'light',
+        default_theme: 'dark',
       },
       explore+: {
         enabled: true,


tomwilkie:
Nooooo!

gouthamve:
Agree with Tom. I like the light mode :P

tomwilkie:
“Manager mode” :-)

rfratto:
👍 for dark mode. Maybe we could have a separate link for Dark Grafana on the NGINX admin services page?

Klaven:
:+1: for dark mode as default mode.

tomwilkie:
Its super easy for users of this module to specify dark mode (just merge a new value info the grafana_config struct).

This kinda thing is always going to be contentious, and there is not right answer.  Grafana will eventually let you choose defaults per user and persist them with cookies.

Until then, I want to leave this as is.  Sorry.

PR #178, opened 2019-10-25
[WIP] Add clustered alertmanager

beorn7:
Signed-off-by: beorn7 <beorn@grafana.com>


beorn7:
This should work "in principle". Still has to be "tested in production". See https://github.com/raintank/deployment_tools/pull/1752 for the PR to apply this in production but without any effective change yet. (We can then switch over to a global AM gradually and see how it goes.)

beorn7:
There is one known problem: 
```jsonnet
          $.core.v1.containerPort.newUDP('gossip-udp', $._config.alertmanager_gossip_port),
          $.core.v1.containerPort.new('gossip-tcp', $._config.alertmanager_gossip_port),
```

This _should_ create two named ports, both pointing to the same port number but with a different protocol. However, both of the lines are creating only one port, possibly with mixed information from the above (like name `gossip-tcp` but protocol being UDP). I believe K8s allows this creation, and I assume this is either a problem in ksonnet or in the kausal util lib. I'm still investigating. If you know anything about it, please let me know.

beorn7:
A bit of research tells me that this is probably a problem of the ksonnet flattening: It uniquely identifies ports by their number, ignoring the protocol. Any ideas how to proceed from here?

beorn7:
More research results: I have prematurely blamed ksonnet. `ks` is actually doing the right thing. This is a problem with Tanka. @sh0rez I'll file an issue over there.

beorn7:
To follow up here: The port issue is caused by a bug in `kubectl diff`. The ports defined here should work as advertised.

PR #179, opened 2019-10-25
Remove redundant anchoring from Prometheus regexp's

beorn7:
Regexp's in Prometheus are always anchored by default. Adding
redundant anchoring is not only a humungus waste of bytes, but also
confuses people into believing the regexp's are not anchored by
default (which can backfire badly elsewhere).

(Same thing as https://github.com/raintank/deployment_tools/pull/1755 , just in this repo.)


beorn7:
See, we are already reaping the benefits of this…

PR #183, opened 2019-11-01
Switch from DNS-SRV to plain DNS for non-local Alertmanagers

beorn7:
I assumed, the K8s-provided DNS-SRV records for named ports would
resolve into all the individual pods having that port. Instead, it
resolves to the load-balanced service IP number. That doesn't help us
because Prometheus servers have to send their alerts to _each_
Alertmanager, not just one of them.

So we have to hardwire the DNS names of each AM pod. The only downside
is that a change of the replica count is not autopropagated but
creates a `tk diff` that needs to be applied explicitly.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -40,13 +40,16 @@
               }],
             },
         ] + if $._config.alertmanager_cluster_self.global then [{
-          // For non-local instances, use DNS-SRV SD.
+          // For non-local instances, use static DNS entries.
+	  // Sadly, the K8s-provided DNS-SRV records only point to the service IP,
+	  // but we need to send alerts to every Alertmanager instance individually.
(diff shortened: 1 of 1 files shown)


woodsaj:
> I assumed, the K8s-provided DNS-SRV records for named ports would
resolve into all the individual pods having that port

you can configure a k8s service to do this.  You just set `ClusterIP: None` in the service spec.

https://kubernetes.io/docs/concepts/services-networking/service/#headless-services

beorn7:
Good to know.

However, we actually _need_ the service IP number (for the web UI).
If there is no way of resolving to pod IPs while at the same time having a service IP, I think we still have to go down the path in this PR.

beorn7:
FTR: We had a bit more discussion on slack, considering the alternative of creating a 2nd service for AM, this time as headless, for the usecase of hitting every single instance. But we decided that the way in this PR is not much better or worse technically, but much simpler to configure and to understand.

PR #189, opened 2019-11-27
feat(ksonnet-util): k.libsonnet using jb package

### LINKED ISSUE #188: k.libsonnet using jsonnet-bundler
At the moment, `ksonnet-util` is using the naked `k.libsonnet` import Ksonnet used to promote:
https://github.com/grafana/jsonnet-libs/blob/8f9d72b2e35b5f3cc1b7c2a8af9bbae7658804e2/ksonnet-util/kausal.libsonnet#L2

This however is not really possible with Tanka anymore, because of two reasons:

1. Tanka does not do magic, so we do not magically add a library to the import paths
2. `jb` does not support installing naked files (it did by accident), so you cannot `jb install github.com/ksonnet/ksonnet-lib/ksonnet.beta.3/k.libsonnet`

Our best solution yet is to use the actual package of this library, which is `ksonnet.beta.3`, located at `github.com/ksonnet/ksonnet-lib/ksonnet.beta.3`.  
Doing this would involve adding the package prefix to the above mentioned import. Furthermore, it would be cool to add a `jsonnetfile.json` (no `.lock.json` though) to the `ksonnet-util` folder, so that `jb` automatically takes care of providing the library when `ksonnet-util` is installed :D

Opinions?

/cc @tomwilkie @malcolmholmes @rfratto @gouthamve

sh0rez:
Switches from the "magic" `k.libsonnet` import of `ksonnet` to the more future
proof approach of using the actual package installed using `jb`.

Fixes #188

### DIFF
--- a/ksonnet-util/.gitignore
+++ b/ksonnet-util/.gitignore
@@ -1,0 +1,2 @@
+vendor/
+jsonnetfile.lock.json
--- a/ksonnet-util/jsonnetfile.json
+++ b/ksonnet-util/jsonnetfile.json
@@ -1,0 +1,14 @@
+{
+  "dependencies": [
+    {
+      "name": "ksonnet.beta.3",
+      "source": {
+        "git": {
+          "remote": "https://github.com/ksonnet/ksonnet-lib",
+          "subdir": "ksonnet.beta.3"
+        }
+      },
+      "version": "master"
+    }
+  ]
+}
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -1,5 +1,5 @@
 // Override defaults paramters for objects in the ksonnet libs here.
-local k = import 'k.libsonnet';
+local k = import 'ksonnet.beta.3/k.libsonnet';
 
 k {
   _config+:: {


sh0rez:
Actually thinking about this a little bit more, I am not sure if this is really a good idea.

We would effectively force everyone that uses this to stick to `ksonnet.beta.3` - which is incompatible with Kubernetes 1.16 and up. 

What about not including `k.libsonnet` at all and letting the users choose their own version?

```jsonnet
(import "ksonnet.beta.X/k.libsonnet") +
(import "ksonnet-util/kausal.libsonnet) + 
...
```

sh0rez:
The actual issue lies in the fact that `k.libsonnet` had a magic meaning in `ksonnet` (was always matching the cluster version), which is against fundamental design decision of `tk` and `jb` (read no magic).

Still, we have the issue that libraries must be able to depend on `k.libsonnet` without stating the exact version number. 

Maybe it would be better to have a generic import path (`ksonnet-lib/k.libsonnet`) and a git tag for each version of the Kubernetes API. This would allow to leverage the version support of `jb` very well:

Shared libraries could just add `ksonnet-lib/k.libsonnet@master` to their `jsonnetfile.json`, so you have a flawless install experience.  
When a user has special needs, one could just manually `jb install` the desired version, which would take precedence over all other transient versions, because it is in the top level `jsonnetfile.json`

PR #190, opened 2019-12-04
Fix name variable in memcached dashboard

pracucci:
The `memcached` jsonnet is configured to run memcached as a stateful set, so the `owner_name` is the name of the stateful set. Because of this, the current regex used to extract the `owner_name` - while fetching values for the `$name` dashboard variable - is not correct, because strips away the last part of the stateful set name.

Before this PR:

Screen Shot 2019-12-04 at 17 05 51

After this PR:

Screen Shot 2019-12-04 at 17 05 41


gouthamve:
This was last changed here: https://github.com/grafana/jsonnet-libs/pull/166/files But we now use memcached as a statefulset

tomwilkie:
Is there a way we can make this work for both deployments and statefulsets?

pracucci:
> Is there a way we can make this work for both deployments and statefulsets?

@tomwilkie Please see https://github.com/grafana/jsonnet-libs/pull/191

PR #191, opened 2019-12-05
Fix memcached dashboard name variable picking it from a memcached metric

pracucci:
Following up [Tom's comment](https://github.com/grafana/jsonnet-libs/pull/190#issuecomment-561765030) we could make the name variable work in any context simply picking it from a memcached metric instead of a K8S metric (where mangling it is more complicated). All in all, we'll use it to filter memcached metrics by `job` label.


tomwilkie:
Good idea! The “namespace/job” format is our opinion and I’d like the mixin to work without it. So I’d drop the separate namespace selector and just select the job.

pracucci:
@tomwilkie I'm not sure I'm getting it. All metrics in this dashboard are filtered by `job=~"($namespace)/($name)"` which already implies that the job is composed by namespace + name. It looks the issue you're talking about is not a regression of this PR.

tomwilkie:
Yes I'm suggesting you get rid of that, and filter based on `job=~"$job"` and then have a job selector at the top.  Much simpler IMO - I was try to overcomplicate this originally.

pracucci:
@tomwilkie Thanks for your reply. I've applied the changes as you suggested and tested it in Grafana (looks working as expected).

PR #192, opened 2019-12-17
fix(ksonnet-util): Kubernetes 1.16 deployments

sh0rez:
With Kubernetes 1.16, the `extensions/v1beta1` and `apps/v1beta` have been
removed in favor of `apps/v1`.

While `ksonnet.beta.4` supports this version, our helper did not.

Most severe issues:
- by not using named arguments to `port.newNamed()` we were suffering from
a change in the API which flipped the parameters order, effectively generating
invalid yaml (port name numeric, port number being the actual name)
- `apps.v1` was not overridden with our enhancements at all, generated
deployment was lacking labels, etc.

Furthermore, the `apps/v1` group requires a `matchLabels` or similar selector,
which previous ones did not. Added this.

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -160,16 +160,8 @@ k {
 
     statefulSet+: {
       new(name, replicas, containers, volumeClaims, podLabels={})::
-        { kind: 'StatefulSet'} + 
-        { apiVersion: 'apps/v1beta1' } + 
-        self.mixin.metadata.withName(name) + 
-        self.mixin.spec.withReplicas(replicas) + 
-        self.mixin.spec.template.spec.withContainers(containers) + 
-        self.mixin.spec.template.metadata.withLabels(podLabels { name: name }) +
-        (if std.length(volumeClaims)>0
-        then self.mixin.spec.withVolumeClaimTemplates(volumeClaims)
-        else {}) +
-        super.mixin.spec.updateStrategy.withType('RollingUpdate'),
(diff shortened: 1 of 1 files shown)


ksonnet-util/kausal.libsonnet
```diff
@@ -143,29 +143,25 @@ k {
...
       new(name, replicas, containers, volumeClaims, podLabels={})::
-        { kind: 'StatefulSet'} + 
-        { apiVersion: 'apps/v1beta1' } + 
-        self.mixin.metadata.withName(name) + 
-        self.mixin.spec.withReplicas(replicas) + 
-        self.mixin.spec.template.spec.withContainers(containers) + 
-        self.mixin.spec.template.metadata.withLabels(podLabels { name: name }) +
-        (if std.length(volumeClaims)>0
-        then self.mixin.spec.withVolumeClaimTemplates(volumeClaims)
-        else {}) +
```
malcolmholmes: you've just removed some important code here.

If we have a statefulset that has no volumes (for some apps, we use a statefulset for the consistent pod naming, not its volume persistence), then by default `k.libsonnet` will generate a resource that will **always** generate a diff (IIRC it generates an empty list, which Kubernetes drops, thus Jsonnet and K8s are always diferent). These three lines (which necessitate this whole block) cause Jsonnet to render a statefulset that does NOT include the volumeclaim block if no volumes are used.

sh0rez: Fixed in 332f7eb4e62aabe93caeb53e0a755d4cd69dd356

PR #193, opened 2019-12-26
Split prometheus/grafana config maps into lib

Duologic:
This will allow us to import the config maps logic into another library
and provision dashboard, notification channels, alert rules and other config maps
separate from (Grafana) deployments. The config maps labels will allow us
to query the API and aggregate them later on in the process.

### DIFF
--- a/prometheus-ksonnet/lib/config.libsonnet
+++ b/prometheus-ksonnet/lib/config.libsonnet
@@ -110,5 +110,8 @@
     // Optionally shard dashboards into multiple config maps.
     // Set to the number of desired config maps.  0 to disable.
     dashboard_config_maps: 0,
+
+    // Optionally add labels to grafana config maps.
+    grafana_config_maps_labels: {},
   },
 }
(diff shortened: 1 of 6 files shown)


prometheus-ksonnet/lib/config.libsonnet
```diff
@@ -38,6 +38,7 @@
     prometheus_path: '/prometheus/',
     prometheus_port: 80,
     prometheus_web_route_prefix: $._config.prometheus_path,
+    prometheus_config_dir: '/etc/prometheus',
```
woodsaj: This only changes the path that prometheus looks in for the config, but does not change the path that the configMap is mounted into.  I understand this is the intent here, but i feel users will see the config option and not understand the impact of changing it.

can we not just override the prometheus container args or configMap mount path using jsonnet?

Duologic: Agreed, this is a bit confusing, reverting that.

PR #201, opened 2020-01-15
upgrade apiVersion of deployment/daemonset/statefulset from v1beta1 to v1 for k8s v1.16+

benjaminhuo:
Signed-off-by: Benjamin <benjamin@yunify.com>

### DIFF
--- a/consul/consul.libsonnet
+++ b/consul/consul.libsonnet
@@ -294,7 +294,7 @@ k {
     ]) +
     container.withPorts(containerPort.new('http-metrics', 9107)),
 
-  local deployment = $.apps.v1beta1.deployment,
+  local deployment = $.apps.v1.deployment,
 
   consul_deployment:
     deployment.new('consul', $._config.consul_replicas, [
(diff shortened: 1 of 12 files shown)


andrew-waters:
@benjaminhuo I'm pretty sure this has broken the memcache lib.

v1 statefulset requires a selector: if you don't add it (and you don't have it in the memcache code here), you'll get a validation error:

```
kubectl failed exit status 1: error validating data: ValidationError(StatefulSet.spec): missing required field "selector" in io.k8s.api.apps.v1.StatefulSetSpec
```

I've had a world of pain with Loki today that seems to root from this - and rolling back to `$.apps.v1beta1.statefulSet` fixes my deployment issues. I think the kube API is automatically adding it in v1beta1 but not in v1.

I may have missed something more obvious though

sh0rez:
Very good catch, apologies for having you to figure this out.

I reviewed based on the kubernetes migration doc and it didn't mention this, so I didn't think of this.

You are right, the `v1` API requires an explicit selector.  Opened an issue: https://github.com/grafana/jsonnet-libs/issues/205

andrew-waters:
No worries, thanks for the quick fix. I was wondering how to persist my vendor change 😂

benjaminhuo:
@andrew-waters Sorry for overlooking this and thanks @sh0rez 's quick fix

PR #203, opened 2020-01-15
Allow plugins

malcolmholmes:
The current Grafana cannot be configured to install plugins. The entrypoint of the Docker container, `run.sh` can do this, but we avoid it so as to set our own config file.

This change switches to running Grafana with the `run.sh` entrypoint, and passes the needed config file in as an environment variable.

It also creates a `grafana_plugins: [ ]` element, and then joins names found in this array and hands them to `run.sh` via `GF_INSTALL_PLUGINS`.

This means that we can add plugins to Grafana by adding config such as this:

```
grafana_plugins+:: ['grafana-piechart-panel'],
```
Which I think is rather cool :-)

### DIFF
--- a/prometheus-ksonnet/lib/grafana.libsonnet
+++ b/prometheus-ksonnet/lib/grafana.libsonnet
@@ -161,10 +161,10 @@
   grafana_container::
     container.new('grafana', $._images.grafana) +
     container.withPorts($.core.v1.containerPort.new('grafana', 80)) +
-    container.withEnv([
-      container.envType.new('GF_PATHS_CONFIG', '/etc/grafana-config/grafana.ini'),
-      container.envType.new('GF_INSTALL_PLUGINS', std.join(',', $.grafana_plugins)),
-    ])+
+    container.withEnvMap({
+      'GF_PATHS_CONFIG': '/etc/grafana-config/grafana.ini',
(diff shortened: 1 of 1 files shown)


woodsaj:
@sh0rez  I am not fan of your commit to move grafana dashboards into $._config

a) it is inconsistent with how everything else is provisioned, eg, prometheus_alerts and prometheus_rules.
b) $._config should be limited to fields that a user might want to set.  Dashboards are resources that are created by other mixins.

sh0rez:
Fair enough, didn't think too much about this specific field.

The rationale behind this change in general was to stick to the pattern we are trying to popularize with the upcoming Tanka guide on creating libraries:

- all parameters in `$._config` under a key-name that describes what is to be set
- using nesting instead of underscores
- no "free" (top level) settable variables

This structure should it make the config parameters more discoverable after all by putting those under well defined places.

Considering this, do you still prefer the `dashboards` to be moved back to the top level? Happy with both solutions! What about `dashboardShards`? Seems to be a user parameter, should that stay in `_config`?

woodsaj:
@sh0rez i think you should revert your commit and so the grafana Plugin support can be added.

changing how $.config is used should be discussed in a separate PR.

sh0rez:
@woodsaj done. However, we now introduce `grafana_plugins`, which needs to deprecated soon when we switch to `_config`

PR #204, opened 2020-01-17
Use `default` as a picker value for datasource variable

navi7:
Per discussion in https://github.com/kubernetes-monitoring/kubernetes-mixin/pull/307, this changes the selector to go for the 'default' datasource instead of the hardcoded `Prometheus`.

### DIFF
--- a/grafana-builder/grafana.libsonnet
+++ b/grafana-builder/grafana.libsonnet
@@ -88,8 +88,8 @@
       list: [
         {
           current: {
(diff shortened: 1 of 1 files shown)


beorn7:
I think this makes sense (but I was part of the above linked discussion). Would be great if another Grafanista could have a look.

beorn7:
Thanks @sh0rez . Nobody else seems to have any objections, so I'm merging.

PR #210, opened 2020-01-22
serviceFor() port naming breaks Istio enabled services

prune998:
# Description 
For Istio to work, port name in Services need to be formed with `<protocol>[-<suffix>]`
https://istio.io/docs/ops/configuration/traffic-management/protocol-selection/

When using `util.serviceFor()`, the ports are named with the container name first : 
```
servicePort.newNamed(c.name + '-' + port.name, port.containerPort, port.containerPort) +
```

# Solution
I propose to change the naming to the other way around:
```
servicePort.newNamed(port.name + '-' + c.name, port.containerPort, port.containerPort) +
```

As far as I know, nothing else is dependent on this naming beside Istio...

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -231,7 +231,7 @@ k {
       local service = $.core.v1.service;
       local servicePort = service.mixin.spec.portsType;
       local ports = [
-        servicePort.newNamed(c.name + '-' + port.name, port.containerPort, port.containerPort) +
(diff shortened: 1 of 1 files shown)


sh0rez:
This would be a breaking change for existing users ... I'd suggest an alternative solution as in #212, which maintains the current behavior by default, but allows to adapt the name to your liking by changing the `printf` template

prune998:
closed in favor of https://github.com/grafana/jsonnet-libs/pull/212

PR #213, opened 2020-01-29
Remove RBAC helpers.

mplzik:
These helpers are already implemented in k8s.libsonnet and have
equivalent functionality (adding a convenient `self` to the return
value, allowing to chain the `with` statements).


tomwilkie:
Does this have any effect on the generated yaml?

mplzik:
@tomwilkie none, as far as I can tell. If I understand correctly, the only case where this could be problematic is manifests that were malfunctioning in the first place (that is, were missing some parameters due to e.g. `withKind` being called as a last in the call chain, originally discarding any values returned by prior `with...` functions and now including them.

tomwilkie:
@mplzik as long as you've checked there are no diff when we use this in deployment tools, LGTM

PR #214, opened 2020-02-07
Make serviceFor inherit the deployment namespace

JeffreyVdb:
The serviceFor utility does not inherit the namespace from the given deployment. With this commit, it will do just that.

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -234,7 +234,11 @@ k {
         labels,  // selector
         ports,
       ) +
-      service.mixin.metadata.withLabels({ name: deployment.metadata.name }),
+      service.mixin.metadata.withLabels({ name: deployment.metadata.name }) + (
+        if 'namespace' in deployment.metadata then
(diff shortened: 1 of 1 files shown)


tomwilkie:
Sorry I missed this! Seems reasonable to me, although the way we use this means the namespace is always set by Tanka.

ksonnet-util/kausal.libsonnet
```diff
@@ -234,7 +234,11 @@ k {
         labels,  // selector
         ports,
       ) +
-      service.mixin.metadata.withLabels({ name: deployment.metadata.name }),
+      service.mixin.metadata.withLabels({ name: deployment.metadata.name }) + (
+        if 'namespace' in deployment.metadata then
+          service.mixin.metadata.withNamespace(deployment.metadata.namespace)
+        else {}
+      ),
```
tomwilkie, line 242:
```suggestion
      service.mixin.metadata.withLabels({ name: deployment.metadata.name }) +
      (
        if 'namespace' in deployment.metadata 
        then service.mixin.metadata.withNamespace(deployment.metadata.namespace)
        else {}
      ),
```

PR #217, opened 2020-02-23
feat(ksonnet-util): proper pvc `new()`

sh0rez:
The previous `new()` constructor for `core.v1.persistentVolumeClaim` was
pretty useless, because it returned an empty object, not even
`apiVersion` or `kind`

Thus I created a new one, that initializes all required fiels of a PVC:

- `metadata.name`
- `spec.resources.requests`: how big the PVC is going to be
- `spec.accessModes` (default `ReadWriteOnce`)

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -56,7 +56,12 @@ k {
       },
 
       persistentVolumeClaim+:: {
-        new():: {},
+        // new() creates a new PVC with the given name (e.g. grafana)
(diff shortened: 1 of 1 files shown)


malcolmholmes:
How back compatible is this? Using new() would create PVCs with empty entries, no?

sh0rez:
It's not backwards compatible because it made no sense. 

The old behavior was broken (it created unusable, incomplete resources).

jalseth:
When will this be merged? I'm currently having to work-around this bug in my resources.

PR #226, opened 2020-03-23
Upgrade kubernetes-mixin

Duologic:
We've updated most clusters to 1.14, should be safe to update kubernetes-mixin now.

### DIFF
--- a/prometheus-ksonnet/jsonnetfile.json
+++ b/prometheus-ksonnet/jsonnetfile.json
@@ -8,7 +8,7 @@
                     "subdir": ""
                 }
             },
-            "version": "release-0.1"
+            "version": "master"
         },
         {
             "name": "prometheus-mixin",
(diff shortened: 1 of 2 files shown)


prometheus-ksonnet/jsonnetfile.json
```diff
@@ -8,7 +8,7 @@
                     "subdir": ""
                 }
             },
-            "version": "release-0.1"
+            "version": "master"
```
mplzik: nit: this sounds like a floating label that might change quite a lot -- is that ok?

Duologic: The point is not a nit. As @malcolmholmes also pointed out to me, we should not force a new version just because we upgraded. Fortunately, this is not much of a problem because of how `jb` (0.2.0 at least) works: it doesn't by default install/vendor dependencies of dependencies, because of this it is likely a conscious decision of downstream to vendor it in or not.

Duologic: As to justify the changes specific to our environment, the overrides and the version 'release-0.1' here were originally because we didn't upgrade. This PR should set things right.

malcolmholmes: Yep, this was a locking back for Grafana Labs. This sets this right.

PR #228, opened 2020-03-24
Update Prometheus to 2.17.0 and Alertmanager to 0.20.0

beorn7:
Both updates are low risk and should "just work".

(Also note that this will only propagate to production with a vendoring update.)


beorn7:
(I'll do the vendoring update and the actual rollout tomorrow.)

PR #234, opened 2020-04-07
Don't merge mixins in the global namespace anymore.

tomwilkie:
Instead, put them in the $.mixins map.

This allows for mixins to be places in specific Grafana directories, and also removes the horrible bug around playbook links.

Still support merging into the global namespace for the time being, but this will be going away.

NB this also removes the old field names (`grafana_dashboards`), I couldn't find a mixin that used this anymore.

### DIFF
--- a/prometheus-ksonnet/grafana/dashboards.libsonnet
+++ b/prometheus-ksonnet/grafana/dashboards.libsonnet
@@ -29,7 +29,7 @@
     function(mixinName, acc)
       local mixin = $.mixins[mixinName] + emptyMixin;
       if std.objectHas(mixin, 'grafanaDashboardFolder')
-      then acc + {
+      then acc {
         [mixin.grafanaDashboardFolder]: mixin.grafanaDashboards,
       }
       else acc,
--- a/prometheus-ksonnet/grafana/dashboards.libsonnet
+++ b/prometheus-ksonnet/grafana/dashboards.libsonnet
@@ -10,11 +10,14 @@
   // New API: Mixins go in the mixins map.
   mixins+:: {},
 
-  // Legacy extension points for you to add your own dashboards.
+  // emptyMixin allows us to reliably do `mixin.grafanaDashboards` without
+  // having to check the field exists first. Some mixins don't declare all
+  // the fields, and thats fine.
   local emptyMixin = {
     grafanaDashboards+: {},
   },
 
+  // Legacy extension points for you to add your own dashboards.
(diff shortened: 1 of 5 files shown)


prometheus-ksonnet/mixins.libsonnet
```diff
@@ -0,0 +1,44 @@
+{
+  // Add you mixins here.
+  mixins+:: {
+    kubernetes+:
```
sh0rez, line 4:
```suggestion
    kubernetes:
```

Don't think we need to merge here. This could cause hard to debug issues in the future

prometheus-ksonnet/grafana/dashboards.libsonnet
```diff
@@ -7,11 +7,35 @@
...
-  grafana_dashboards+:: {},
-  grafanaDashboards+:: $.dashboards + $.grafana_dashboards,
-  dashboardsByFolder+:: {},
+  // New API: Mixins go in the mixins map.
+  mixins+:: {},
+
+  // Legacy extension points for you to add your own dashboards.
+  local emptyMixin = {
+    grafanaDashboards+: {},
+  },
```
sh0rez, line 16: What do we need this for? If you actually added a mixin here, it would be recursively merged with all other mixins, wouldn't it?

tomwilkie: Hence why its a local; prevents people adding any mixins here.

It is here so we can reliably do `mixin.grafanaDashboards` without having to check the field exists first.  Some mixins don't declare all the fields, and thats fine.

sh0rez: Ah fine, the comment confused me

malcolmholmes:
But the `$.mixins` map is still global.

You could say, the current approach is that `prometheus-ksonnet` "reaches out" into the global namespace for its mixins. What I am attempting to move us towards is where we create a Grafana or a "monitoring thing" and pass it our mixins:

```
local mymixin = import 'mymixin.libsonnet';
local grafana = import 'grafana.libsonnet';
{
  grafana: grafana.new()
     + grafana.withMixin(mymixin),
}
```
Which is a more conventional programming model. Because the `grafana` object (or monitoring thingie) holds the mixins within it, we can now have as many as we like, without them overlapping, because they have their resources handed to them, rather than the 'reaching out' that we are currently doing.

tomwilkie:
@malcolmholmes said

>  I don't see this as a need to block that PR

PR #235, opened 2020-04-08
Copy some dashboard opinions from kubernete-mixin and apply to all dashboards.

tomwilkie:
Previously, this would happen automatically as all mixins were merged into one.  Now we keep mixins in different namespaces, this needs to be done by out grafana jsonnet code.

Opinions are:
  // - Dashboard UIDs should be the md5 hash of their filename.
  // - Timezone should be default (ie local).
  // - Tooltip should only show a single value.

### DIFF
--- a/prometheus-ksonnet/grafana/dashboards.libsonnet
+++ b/prometheus-ksonnet/grafana/dashboards.libsonnet
@@ -10,17 +10,48 @@
   // New API: Mixins go in the mixins map.
   mixins+:: {},
 
-  // emptyMixin allows us to reliably do `mixin.grafanaDashboards` without
+  // mixinProto allows us to reliably do `mixin.grafanaDashboards` without
   // having to check the field exists first. Some mixins don't declare all
   // the fields, and thats fine.
-  local emptyMixin = {
-    grafanaDashboards+: {},
+  //
+  // We also use this to add a little "opinion":
+  // - Dashboard UIDs should be the md5 hash of their filename.
+  // - Timezone should be "default" (ie local).
+  // - Tooltip should only show a single value.
+  local mixinProto = {
+    grafanaDashboards+:: {},
+  } + {
+    local grafanaDashboards = super.grafanaDashboards,
+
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/grafana/dashboards.libsonnet
```diff
@@ -10,17 +10,48 @@
...
-    grafanaDashboards+: {},
+  //
+  // We also use this to add a little "opinion":
+  // - Dashboard UIDs should be the md5 hash of their filename.
+  // - Timezone should be "default" (ie local).
+  // - Tooltip should only show a single value.
+  local mixinProto = {
+    grafanaDashboards+:: {},
+  } + {
+    local grafanaDashboards = super.grafanaDashboards,
```
sh0rez, line 24: just to clarify, `super` here is the already merged `$.mixins[mixinName] + { grafanaDashboards+:: {} }`?

prometheus-ksonnet/grafana/dashboards.libsonnet
```diff
@@ -10,17 +10,48 @@
...
+  // mixinProto allows us to reliably do `mixin.grafanaDashboards` without
   // having to check the field exists first. Some mixins don't declare all
   // the fields, and thats fine.
-  local emptyMixin = {
-    grafanaDashboards+: {},
+  //
+  // We also use this to add a little "opinion":
+  // - Dashboard UIDs should be the md5 hash of their filename.
+  // - Timezone should be "default" (ie local).
+  // - Tooltip should only show a single value.
```
sh0rez, line 20: `should` sounds like we do validation here – but we actually just force this behaviour

prometheus-ksonnet/grafana/dashboards.libsonnet
tomwilkie: Yes, thats just there so we can reliably reference super.grafanaDashboards (you can do std.hasField(super, ...).

Issue #160, opened 2019-08-17
Tidy up Prometheus config


tomwilkie:
- [x] remove option to run as a deployment. (#156)
- [x] remove old non-sharded dashboard config map.
- [x] remove support for old mixin field names (`grafana_dashboards` vs `grafanaDashboards`) #234
- [x] scrape grafana metrics (#239)
- [x] stop running grafana as root (#240)
- [x] stop running prometheus, alertmanager etc as root (but keep them serving on port 80 if possible) (#244)
- [ ] get rid of the lib folder, put them all in the base folder like the rest of our modules.
- [x] move HA pair handling here.
- [ ] move the config fields into their respective files, and remove unused ones.
- [ ] audit and update node-exporter flags (we have ones left over from 0.16)
- [x] add silence link to alert template
- [x] make a file per-mixins' alerts, not one big file - makes diffing easier when order changes.

beorn7:
I guess @cstyan and I can divide this among the two of us.

malcolmholmes:
Grafana does not reload on yaml file changes, just json (i.e. dashboards). This catches users out as they expect dashboards and notification channels to reload but they dont.

The yaml configmaps should have hashes to trigger reloads to fix this.

beorn7:
Note: We have solved the configmap layout in a different way, and tanka made diffing easy in yet another way.

beorn7:
I think we can consider the config related tasks as done or obsolete. @malcolmholmes / @Duologic you did so much there, could you verify?

Which leaves us with node_exporter, which is indeed running in an ancient version. That upgrade with all the ramifications might be a good starter project for someone on the cloud-platform team?

beorn7:
I take silence as a yes. Closing this, as I have created a separate issue for the node-exporter update, #527 .

PR #240, opened 2020-04-13
Don't run Grafana as root anymore,

tomwilkie:
Instead use the service to map port 80 to port 3000.

Also, include the config hash so Grafana gets restarted when the config changes.

/cc @slim-bean @cyriltovena I couldn't get the capabilities thing to work, rummaging around on github issues seems to imply is won't work with some setuid bit on the binary, and I think thats probably a little far.  Hence, I remapped port 80 to 3000 using the service instead.

### DIFF
--- a/prometheus-ksonnet/grafana/config.libsonnet
+++ b/prometheus-ksonnet/grafana/config.libsonnet
@@ -6,7 +6,7 @@
         org_role: 'Admin',
       },
       server: {
-        http_port: 80,
+        http_port: 3000,
         root_url: $._config.grafana_root_url,
       },
       analytics: {
--- a/prometheus-ksonnet/grafana/deployment.libsonnet
+++ b/prometheus-ksonnet/grafana/deployment.libsonnet
@@ -86,7 +86,7 @@
     $.util.serviceFor($.grafana_deployment) +
     service.mixin.spec.withPortsMixin([
       servicePort.newNamed(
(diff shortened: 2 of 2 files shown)


malcolmholmes:
Is there a specific reason not to run Grafana as root? (Other than root being a bad idea in general?)

And why not do the hash thing on all configmaps? (Except dashboards, which reload already). it would be useful (and less confusing) for it to reload datasources/channels too.

tomwilkie:
> Is there a specific reason not to run Grafana as root? (Other than root being a bad idea in general?)

No, just the root being a bad idea.  The grafana container actually doesn't run as root by default - we were hacking it up to make it run as root here.  So really this is removing that hack..

> And why not do the hash thing on all configmaps? (Except dashboards, which reload already). it would be useful (and less confusing) for it to reload datasources/channels too.

I was under the impression everything provisioned via the grafana provisioning YAMLs should automatically get loaded in to Grafana - not just dashboards.  Could be wrong though.

malcolmholmes:
Re root : makes sense.

Re reloading, datasources and notification channel yamls dont reload on change. I presume it is that Grafana watches json but not yaml files. Would be good to reload these too, but doesnt need to be in this PR.

tomwilkie:
Ah, cool - mind adding that to #160 so we don't forget?

Thanks!

PR #243, opened 2020-04-14
Add library for termination handler

Duologic:
Lib for [Kubernetes on GCP Node Termination Event Handler](https://github.com/GoogleCloudPlatform/k8s-node-termination-handler)

This library will provide a DaemonSet and a ClusterRole/ServiceAccount to run it.

*not tested yet, created from [example](https://github.com/GoogleCloudPlatform/k8s-node-termination-handler/tree/master/deploy)*

### DIFF
--- a/k8s-node-termination-handler/handler.libsonnet
+++ b/k8s-node-termination-handler/handler.libsonnet
@@ -1,0 +1,104 @@
+{
+  new(config):: {
+    local k = import 'ksonnet-util/kausal.libsonnet',
+
+    local _config = {
+      namespace: 'kube-system',
+      slack_webhook: '',
+    } + config,
+
+
+    _images+:: {
+      node_termination_handler: 'k8s.gcr.io/gke-node-termination-handler@sha256:aca12d17b222dfed755e28a44d92721e477915fb73211d0a0f8925a1fa847cca',
+    },
+
+
+    local container = k.core.v1.container,
+    container::
+      container.new('node-termination-handler', self._images.node_termination_handler) +
+      container.withCommand(['./node-termination-handler']) +
+      container.withArgsMixin([
+        '--logtostderr',
+        '--exclude-pods=$(POD_NAME):$(POD_NAMESPACE)',
+        '-v=10',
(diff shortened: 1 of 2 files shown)


k8s-node-termination-handler/handler.libsonnet
```diff
@@ -0,0 +1,104 @@
...
+      container.mixin.securityContext.capabilities.withAdd(['SYS_BOOT']) +
+      k.util.resourcesLimits('150m', '30Mi'),
+
+    local daemonSet = k.apps.v1.daemonSet,
+    local tolerations = daemonSet.mixin.spec.template.spec.tolerationsType,
+    local nodeAffinity = daemonSet.mixin.spec.template.spec.affinity.nodeAffinity,
+    local b = nodeAffinity.requiredDuringSchedulingIgnoredDuringExecutionType,
+    daemonset:
+      daemonSet.new('node-termination-handler', [self.container]) +
+      daemonSet.mixin.spec.template.spec.affinity.nodeAffinity.requiredDuringSchedulingIgnoredDuringExecution.mixinInstance(
```
woodsaj, line 40: Are these nodeAffinity rules needed?   is there any harm to the node-termination-handler just running on every node in the cluster?

Duologic: I followed the example, the README of the project says it is not intended for other VMs: 

> Do not consume this project unless you are managing k8s clusters that run non migratable VM types.

I'd say we start with this and evaluate. The project does a delete instead of eviction, we might want to change that behavior before running it with other workloads.

Duologic: Evict vs. Delete is being discussed here: https://github.com/GoogleCloudPlatform/k8s-node-termination-handler/issues/16

woodsaj: Delete makes the most sense.  Evict would block to honor Pod Disruption Budgets,  but if the node is powering off, blocking is not an option.  When the node is shutting down, you just want to send a signal to every to POD to tell it to shutdown, which is what delete does.

PR #244, opened 2020-04-14
Don't run Prometheus and Alertmanager as root.

tomwilkie:

### DIFF
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -127,8 +127,20 @@
     $.util.podPriority('critical')
   else {},
 
+  local service = $.core.v1.service,
+  local servicePort = service.mixin.spec.portsType,
+
   // Do not create service in clusters without any alertmanagers.
-  alertmanager_service: if replicas > 0 then
-    $.util.serviceFor($.alertmanager_statefulset)
-  else {},
+  alertmanager_service:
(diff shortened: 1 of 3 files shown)


tomwilkie:
> I'm not familiar with those port configuration, but if it works as intended, I'm all for it.

I'm putting the jobs back to their default ports, then using the k8s service to map port 80 to those ports.  At least, thats what I'm trying to do...

I double checked prometheus can talk to alert manager, and can't think of anything else I might be breaking.  So lets find out!

prometheus-ksonnet/lib/alertmanager.libsonnet
```diff
@@ -121,14 +121,24 @@
...
   // Do not create service in clusters without any alertmanagers.
-  alertmanager_service: if replicas > 0 then
-    $.util.serviceFor($.alertmanager_statefulset)
-  else {},
+  alertmanager_service:
+    if replicas == 0
+    then {}
+    else
+      $.util.serviceFor($.alertmanager_statefulset) +
+      service.mixin.spec.withPortsMixin([
```
sh0rez, line 137: You're using `withPortsMixin` here – wouldn't that add another port to the service, along the one inferred from the deployment? Is this intentional?

tomwilkie: Yes, I wanted to leave all the existing ports there just in case they are being used by name.

PR #245, opened 2020-04-15
Ensure Grafana Dashboard Provisioner Names and Paths are Unique

trotttrotttrott:
Grafana will not start if there are duplicate providers in dashboards.yml.

I came across this with [loki_mixin](https://github.com/grafana/loki/blob/f86785b68e52866259b294bcba0e6bcc1c90572b/production/loki-mixin/mixin.libsonnet#L4) and [promtail_mixin](https://github.com/grafana/loki/blob/f86785b68e52866259b294bcba0e6bcc1c90572b/production/promtail-mixin/mixin.libsonnet#L4). They both use `grafanaDashboardFolder: 'Loki'` which was resulting in duplicate providers named "dashboards-loki". Grafana fails with:

```
t=2020-04-14T22:55:06+0000 lvl=eror msg="Server shutdown" logger=server reason="Failed to create provisioner: Failed to read dashboards config: could not parse provisioning config file: dashboards.yml error: dashboard name \"dashboards-loki\" is not unique"
```

### DIFF
--- a/prometheus-ksonnet/grafana/dashboards.libsonnet
+++ b/prometheus-ksonnet/grafana/dashboards.libsonnet
@@ -151,8 +151,10 @@
               path: '/grafana/dashboards-%s' % $.folderID($.mixins[mixinName].grafanaDashboardFolder),
             },
           }
-          for mixinName in std.objectFields($.mixins)
-          if $.isFolderedMixin($.mixins[mixinName])
+          for mixinName in std.set(
+            std.filter(function(n) $.isFolderedMixin($.mixins[n]), std.objectFields($.mixins)),
+            keyF=function(n) $.mixins[n].grafanaDashboardFolder
+          )
         ],
       }),
     }),


prometheus-ksonnet/grafana/dashboards.libsonnet
```diff
@@ -151,8 +151,10 @@
               path: '/grafana/dashboards-%s' % $.folderID($.mixins[mixinName].grafanaDashboardFolder),
```
DanCech, line 151: If the problem is that the paths on disk need to be unique, and we're already keying mixins by `mixinName` (so that must be unique), then wouldn't it make more sense to use `mixinName` to name the folders so that they're also going to be unique?  AFAIK grafana is fine with multiple providers putting dashboards into the same folder in Grafana, it just doesn't like multiple providers with the same path on disk.  Wouldn't hurt to update line 144 as well: `name: 'dashboards-%s' % $.folderID(mixinName),`
```suggestion
              path: '/grafana/dashboards-%s' % $.folderID(mixinName),
```

trotttrotttrott: Thanks, @DanCech. Much better.

tomwilkie:
The intention was to have the loki and promtail dashboards in the same folder.  

In other places I've used a map to dedupe the entries (see #241) - I think thats what we should do here, instead of using mixin name as the folder name.

DanCech:
Yeah, the way this is handled in Grafana is a bit weird, each provider must have its own path on disk but it is possible to have multiple providers pushing dashboards into the same folder in Grafana.  If you do that however, Grafana will emit error log lines though it does work.  The intent of my suggestion was to keep things simple and easy to reason about by having a provider for each mixin (with multiple providers potentially provisioning dashboards into the same folder in Grafana) but it seems like that is discouraged so it may make more sense to re-arrange things so that we end up with a provider for each Grafana folder rather than each mixin.

tomwilkie:
@trotttrotttrott @Duologic this is what I was talking about - sorry I couldn't explain it more clearly.

tomwilkie:
Output after running this:

```diff
diff -u -N /var/folders/g5/3pb5_td927l61zm9xfmmtcp80000gn/T/LIVE-983987820/v1.ConfigMap.default.grafana-dashboard-provisioning /var/folders/g5/3pb5_td927l61zm9xfmmtcp80000gn/T/MERGED-115563227/v1.ConfigMap.default.grafana-dashboard-provisioning
--- /var/folders/g5/3pb5_td927l61zm9xfmmtcp80000gn/T/LIVE-983987820/v1.ConfigMap.default.grafana-dashboard-provisioning 2020-04-15 15:55:06.000000000 +0100
+++ /var/folders/g5/3pb5_td927l61zm9xfmmtcp80000gn/T/MERGED-115563227/v1.ConfigMap.default.grafana-dashboard-provisioning       2020-04-15 15:55:06.000000000 +0100
@@ -75,14 +75,6 @@
             path: /grafana/dashboards-prometheus
         orgId: 1
         type: file
-      - disableDeletion: true
-        editable: false
-        folder: Loki
-        name: dashboards-promtail
-        options:
-            path: /grafana/dashboards-promtail
-        orgId: 1
-        type: file
 kind: ConfigMap
 metadata:
   annotations:
```

Duologic:
The solution looks really clever. I do worry that if I come back to this in a week or a month, if I still understand what is going on here.

trotttrotttrott:
I _think_ my initial solution/commit does basically the same thing - https://github.com/grafana/jsonnet-libs/pull/245/commits/582bea4ae98ce7c85f81285b96652b356ed4acd5. Instead of pushing it through a map it uses `std.set` to ensure uniqueness.

Only difference is the array order in the `grafana-dashboard-provisioning` configmap since `std.set` sorting behavior puts lowercase characters after any uppercase.

tomwilkie:
> I think my initial solution/commit does basically the same thing - 582bea4. Instead of pushing it through a map it uses std.set to ensure uniqueness.

You're 100% right, I don't know how I missed that.  And that is much simpler.  I'll back out my changes.  Sorry.

tomwilkie:
Ah I see I didn't look at the individual commits, just the total diff.  Do you want to remove the last three commits and go with the original way?

PR #246, opened 2020-04-15
Prevent conflicts with grafanaDashboardFolder naming

Duologic:
`name` should be a unique key in dashboards.yaml, this made Grafana crash (after update).

`folder` should not conflict with Dashboard names (`title`), this throws an error and prevents conflicting dashboard from being loaded.


Duologic:
Looks like I did nearly the same as @trotttrotttrott in #245

tomwilkie:
Lets close in favour of #245

PR #247, opened 2020-04-15
Add ' folder' suffix to prevent naming conflict

Duologic:
Resolves:

name=dashboards-prometheus error="Dashboard name cannot be the same as folder"

### DIFF
--- a/prometheus-ksonnet/grafana/dashboards.libsonnet
+++ b/prometheus-ksonnet/grafana/dashboards.libsonnet
@@ -143,7 +143,7 @@
           {
             name: 'dashboards-%s' % $.folderID($.mixins[mixinName].grafanaDashboardFolder),
             orgId: 1,
-            folder: $.mixins[mixinName].grafanaDashboardFolder,
+            folder: '%s folder' % $.mixins[mixinName].grafanaDashboardFolder,
             type: 'file',
             disableDeletion: true,
             editable: false,


tomwilkie:
I'd prefer if didn't mutate the dashboard name - esp if someone calls their folder "Foo folder", this will end up calling it "Foo folder folder".

Can we instead change the dashboard name?  Maybe add `(Overview)` to the end of the dashboard name?

Duologic:
I agree it would look silly to have 'Foo folder folder'. However, these might be 3rd party dashboards that we do not control, implementing overrides when they occur would be a hassle. I think it should be either be handled by the logic implementing that or by changing the name of the folder causing the issues.

We could do this, wdyt?

```diff
diff --git a/prometheus-ksonnet/mixins.libsonnet b/prometheus-ksonnet/mixins.libsonnet
index 9906947..645eddd 100644
--- a/prometheus-ksonnet/mixins.libsonnet
+++ b/prometheus-ksonnet/mixins.libsonnet
@@ -21,7 +21,7 @@

     prometheus:
       (import 'prometheus-mixin/mixin.libsonnet') {
-        grafanaDashboardFolder: 'Prometheus',
+        grafanaDashboardFolder: 'Prometheus folder',

         _config+:: {
           prometheusSelector: 'job="default/prometheus"',
```

Duologic:
@tomwilkie The Prometheus dashboard that conflicts with this folder is from [prometheus-mixin](https://github.com/prometheus/prometheus/tree/master/documentation/prometheus-mixin). We can't expect external mixins to follow suit.

tomwilkie:
I chatted with Carl and there are no plans to fix https://github.com/grafana/grafana/issues/15642 any time soon.

I'm still averse to adding a "folder" prefix to the end of the folder name (for the same reason I don't call by variables "fooStr" or "barInt"). 

I think we can rename clashing dashboards in upstream repos to fix this.

tomwilkie:
I've opened a PR to fix the Prometheus dashboard name: https://github.com/prometheus/prometheus/pull/7489

tomwilkie:
And a PR in deployment tools to update this: https://github.com/grafana/deployment_tools/pull/4117

tomwilkie:
The renamed prometheus dashboard is merged now.

PR #248, opened 2020-04-15
Move contents from lib to base folder

hoenn:
Addresses #160 
>get rid of the lib folder, put them all in the base folder like the rest of our modules.

Should the `prometheus-ksonnet/grafana/` folder also be extracted to the base folder?


tomwilkie:
I've been avoiding doing this as @malcolmholmes has a bunch of outstanding refactors.  In particular, we've been discussing split this up into separate libraries, not simply moving all the files.

> Should the prometheus-ksonnet/grafana/ folder also be extracted to the base folder?

We've only recently split that out, I think that is a nice self-contained library and should stay as such.

malcolmholmes:
> We've only recently split that out, I think that is a nice self-contained library and should stay as such.

I want to see Grafana consumable without Prometheus - we can't assume all Grafana users also use Prometheus. And the Grafana lib has some useful stuff in it. Making sure the Grafana lib is consumable as a subdir of prometheus-ksonnet actually achieves most of this, actually. (other than the "why would I get a Grafana lib from inside a prometheus thing?" question).

So equally, nginx could go into its own folder, prometheus too - perhaps people might want to use prometheus but something other than Grafana. Then, this 'prometheus-ksonnet' thing becomes an aggregation of these powerful libraries.

tomwilkie:
@hoenn please have a go at the feedback or close the PR if you don't have a chance - trying to reduce number of pending PRs in this repo.

PR #250, opened 2020-04-17
Copy of ksonnet.beta.4

Duologic:
Internally Grafana uses a slightly adapted version of ksonnet-lib derived from ksonnet.beta.4. To use this lib, we must follow a process of `tk init` and manually copying over the altered version. This is cumbersome and holds us back from publicly developing against ksonnet-lib.

A 'fork' like this will allow us to threat this lib as any other lib and reference it in other libraries and projects. I've provided a clean history by copying the lib and applying the changes maded internally, it might be useful to keep this git tree and not squash it in.


Duologic:
Replaced by #252

PR #251, opened 2020-04-17
Add smoke test for termination handler lib

Duologic:
Depends on #252


Duologic:
This might become an example for a bigger effort on jsonnet testing.

PR #253, opened 2020-04-17
Update all jsonnetfile.json to jb v0.3.1 format

Duologic:
Not sure if we want this already?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,10 @@ This repository contains various Jsonnet libraries we use at Grafana Labs:
 * [`prometheus-ksonnet`](prometheus-ksonnet/): A set of extensible configurations
   for running Prometheus on Kubernetes.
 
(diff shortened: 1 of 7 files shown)


sh0rez:
This breaks pre-0.3 versions, they won't be able to install these packages anymore then. Unfortunately, jb only has config schema versions from 0.3 and up, so older jb's will just panic.

As we didn't upgrade all environment vendors yet,  we need to do this first

Duologic:
@sh0rez I added a note in the README, we still need to tag master before continuing.

(Ah, it seems I already merged ksonnet-util with an updated jsonnetfile, so we have to tag a version prior to that, that was commit `b7d0399`)

Duologic:
Not needed now, will pick this up later (with newer version of jb).

PR #254, opened 2020-04-17
fix: remove broken persistentVolumeClaim that breaks the resource

jalseth:
The `persistentVolumeClaim+` resource in `kasual.libsonnet` does not add anything and breaks the upstream `new()` in `k8s.libsonnet`. As this did not add any features, removing this resource from the kasual library makes sense.


jalseth:
Duplicates #217

PR #256, opened 2020-04-20
support for generating slack receivers with buttons for alerts

sandeepsukhani:
Signed-off-by: Sandeep Sukhani <sandeep.d.sukhani@gmail.com>

### DIFF
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -18,7 +18,7 @@
   else [],
 
   slackAlertTitle:: '{{ template "__alert_title" . }}',
-  slackAlertText:: {{ template "__alert_text" . }},
+  slackAlertText:: '{{ template "__alert_text" . }}',
 
   build_slack_receiver(name, slack_channel)::
     {
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -17,15 +17,8 @@
     ]
   else [],
 
-  slackAlertTitle:: '[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.cluster }}: {{ .GroupLabels.alertname }} ({{ .GroupLabels.namespace }})',
-  slackAlertText:: |||
-    {{ .CommonAnnotations.summary }}
-    {{ if .Alerts.Firing | len }}Firing alerts:
-    {{ range .Alerts.Firing }}- {{ .Annotations.message }}{{ .Annotations.description }}
-    {{ end }}{{ end }}{{ if .Alerts.Resolved | len }}Resolved alerts:
-    {{ range .Alerts.Resolved }}- {{ .Annotations.message }}{{ .Annotations.description }}
-    {{ end }}{{ end }}
-  |||,
+  slackAlertTitle:: '{{ template "__alert_title" . }}',
+  slackAlertText:: {{ template "__alert_text" . }},
 
   build_slack_receiver(name, slack_channel)::
     {
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -17,20 +17,58 @@
     ]
   else [],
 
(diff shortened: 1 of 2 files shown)


prometheus-ksonnet/lib/alertmanager.libsonnet
```diff
@@ -17,20 +17,58 @@
...
 
+  slackAlertTitle:: '[{{ .Status | toUpper }}{{ if eq .Status "firing" }}:{{ .Alerts.Firing | len }}{{ end }}] {{ .GroupLabels.cluster }}: {{ .GroupLabels.alertname }} ({{ .GroupLabels.namespace }})',
+  slackAlertText:: |||
+    {{ .CommonAnnotations.summary }}
+    {{ if .Alerts.Firing | len }}Firing alerts:
+    {{ range .Alerts.Firing }}- {{ .Annotations.message }}{{ .Annotations.description }}
+    {{ end }}{{ end }}{{ if .Alerts.Resolved | len }}Resolved alerts:
+    {{ range .Alerts.Resolved }}- {{ .Annotations.message }}{{ .Annotations.description }}
+    {{ end }}{{ end }}
+  |||,
```
tomwilkie, line 28: Can we move this into the template too?

prometheus-ksonnet/lib/files/alertmanager_config.tmpl
```diff
@@ -0,0 +1,12 @@
+# This builds the silence URL.  We exclude the alertname in the range
```
tomwilkie, line 1: Might be worth adding a comment saying where this was inspired by.

prometheus-ksonnet/lib/alertmanager.libsonnet
```diff
@@ -17,20 +17,51 @@
     ]
   else [],
 
+  slackAlertTitle:: '{{ template "__alert_title" . }}',
+  slackAlertText:: '{{ template "__alert_text" . }}',
```
tomwilkie, line 21: Now these are much simpler, is there a reason not to inline them below?

prometheus-ksonnet/lib/alertmanager.libsonnet
```diff
@@ -17,20 +17,51 @@
...
+        api_url: $._config.slack_url,
+        channel: slack_channel,
+        send_resolved: true,
+        title: $.slackAlertTitle,
+        text: $.slackAlertText,
+        actions: [
+          {
+            type: 'button',
+            text: 'Runbook :green_book:',
+            url: '{{ (index .Alerts 0).Annotations.runbook_url }}',
```
tomwilkie, line 36: A thought: what happens when there is no runbook url on the alert?

prometheus-ksonnet/lib/alertmanager.libsonnet
sandeepsukhani: It is to let the users override the title and text if they want. I can move it below if we do not want to allow that or most of the users would be happy with what we are offering here. What do you think?

prometheus-ksonnet/lib/alertmanager.libsonnet
sandeepsukhani: I have verified it with our alerts which do not have a runbook url, the button simply does not show up. I guess Slack takes care of it when URL is empty.

prometheus-ksonnet/lib/alertmanager.libsonnet
tomwilkie: Do we actually override it though?  I don't think this library has many users other than us.  Plus, they can specify their own template if need be.

prometheus-ksonnet/lib/alertmanager.libsonnet
tomwilkie: Nice, thanks.

prometheus-ksonnet/lib/alertmanager.libsonnet
sandeepsukhani: We will be using it as is since this is what we already have for our usage. Own template sounds good as well. Will move usage of templates below. Thanks!

sandeepsukhani:
@tomwilkie I have addressed all the feedbacks. Do you want to do another pass or this is good to merge now?

tomwilkie:
That looks really good, thanks for sticking with this Sandeep.  LGTM!

PR #274, opened 2020-06-23
Update prometheus and alertmanager to recent releases

beorn7:


beorn7:
I'll vendor this in in the next days.

PR #280, opened 2020-06-25
HA prometheus statefulset mixin

malcolmholmes:
Adds a mixin that will adapt a prometheus instance to have two replicas, and each has the ability to have its own config.

This PR is for review only - it has not yet been tried out.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
@@ -1,0 +1,42 @@
+{
+  prometheus_config_0: $.prometheus_config,
+  prometheus_config_1: $.prometheus_config,
+
+  prometheusAlerts_0: $.prometheusAlerts,
+  prometheusAlerts_1: $.prometheusAlerts,
+
+  prometheusRules_0: $.prometheusRules,
+  prometheusRules_1: $.prometheusRules,
+
+  prometheus+:: {
+    local configMap = $.core.v1.configMap,
+
+    prometheus_config_map:: {},
+
+    prometheus_config_maps: [{
+      configMap.new('%s-config-0' % self.name) +
+      configMap.withData({
+        'prometheus.yml': $.util.manifestYaml($.prometheus_config_0),
+        'alerts.rules': $.util.manifestYaml($.prometheusAlerts_0),
+        'recording.rules': $.util.manifestYaml($.prometheusRules_0),
+      }),
+      configMap.new('%s-config-1' % self.name) +
+      configMap.withData({
+        'prometheus.yml': $.util.manifestYaml($.prometheus_config_1),
+        'alerts.rules': $.util.manifestYaml($.prometheusAlerts_1),
+        'recording.rules': $.util.manifestYaml($.prometheusRules_1),
+      }),
+    ]},
+
+  prometheus_config_mount:: {},
+
+  prometheus_init_container::
+    container.new('prometheus-init', 'busybox') +
+    container.withCommand(["ln", "-s", "/etc/$HOSTNAME", "/etc/prometheus"]),
+
+  prometheus_statefulset:
+    $.util.configVolumeMount('%s-config-0' % self.name, '/etc/prometheus-0') +
+    $.util.configVolumeMount('%s-config-1' % self.name, '/etc/prometheus-1') +
+    statefulset.mixin.spec.template.spec.withInitContainers(self.prometheus_init_container) +
+    statefulset.mixin.spec.withReplicas(2),
+}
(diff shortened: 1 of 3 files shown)


prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -0,0 +1,42 @@
...
+        'alerts.rules': $.util.manifestYaml($.prometheusAlerts_1),
+        'recording.rules': $.util.manifestYaml($.prometheusRules_1),
+      }),
+    ]},
+
+  prometheus_config_mount:: {},
+
+  prometheus_init_container::
+    container.new('prometheus-init', 'busybox') +
+    container.withCommand(["ln", "-s", "/etc/$HOSTNAME", "/etc/prometheus"]),
```
tomwilkie, line 35: Does this work?  I'd be surprised it the `/etc` directory is shared between the init container and the main container.

That being said, you could mount a tmp volume that is shared...

malcolmholmes: As I said above, I havent tried this :-) that comes next, and there may well be issues like the one you mention.

tomwilkie: Give us a ping when you've tested it and I'll do another review.

malcolmholmes:
This now passes a `tk show` / `tk diff`.

Adding this:
```
  prometheus_ha: $.prometheus
                 { name: 'prometheus-ha' } +
                 (import 'prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet'),
```
will add an entirely new Prometheus pair as a single statefulset, each with individually configurable Prometheus configs.

Now need to work out how we would migrate to this, and actually try it for real.

beorn7:
I think this will work. @malcolmholmes and I came up with a migration plan, see #4078. We'll experiment with it in one of the dev clusters next week. Which will also exercise the code in this PR. @tomwilkie we'd still appreciate your detailed review.

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -0,0 +1,46 @@
...
+
+{
+  prometheus_config_0: $.prometheus_config,
+  prometheus_config_1: $.prometheus_config,
+
+  prometheusAlerts_0: $.prometheusAlerts,
+  prometheusAlerts_1: $.prometheusAlerts,
+
+  prometheusRules_0: $.prometheusRules,
+  prometheusRules_1: $.prometheusRules,
```
tomwilkie: The mix of camelCase and underscores is a little ugly, we should pick one.

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -0,0 +1,46 @@
+local container = $.core.v1.container;
+local statefulset = $.apps.v1.statefulSet;
+
+{
+  prometheus_config_0: $.prometheus_config,
+  prometheus_config_1: $.prometheus_config,
```
tomwilkie: It might be worth adding a cluster and replica label at this point, as per:

https://github.com/grafana/deployment_tools/blob/master/ksonnet/lib/default/prometheus.libsonnet#L59

https://github.com/grafana/deployment_tools/blob/master/ksonnet/lib/default/prometheus.libsonnet#L100

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -0,0 +1,46 @@
...
+        'alerts.rules': $.util.manifestYaml($.prometheusAlerts_1),
+        'recording.rules': $.util.manifestYaml($.prometheusRules_1),
+      }),
+    ]},
+
+  prometheus_config_mount:: {},
+
+  prometheus_config_file:: '/etc/$(POD_NAME)/prometheus.yml',
+
+  prometheus_container+:: container.withEnv([
```
tomwilkie, line 39:
```suggestion
  prometheus_container+:: 
    container.withEnv([
```

prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -30,13 +30,15 @@
...
 
+    prometheus_config_file:: '/etc/prometheus/prometheus.yml',
+
     prometheus_container::
       local _config = self._config;
 
       container.new('prometheus', $._images.prometheus) +
       container.withPorts($.core.v1.containerPort.new('http-metrics', _config.prometheus_port)) +
       container.withArgs([
-        '--config.file=/etc/prometheus/prometheus.yml',
```
tomwilkie, line 39: Would be good to turn these into a map as we do in cortex: https://github.com/grafana/cortex-jsonnet/blob/master/cortex/distributor.libsonnet#L45

malcolmholmes: I like that, but.... it changes the order of the args, causing a redeploy. Would cause a re-deploy of *both* the old and the new prometheus servers! And given that redeploying prometheus is what we're trying to make smoother, this isn't the time to make this change, I'm afraid.

PR #283, opened 2020-07-13
Add session affinity to Alertmanager service

beorn7:


beorn7:
This will prevent the occasional 404s that can be seen after creating a silence.

PR #285, opened 2020-07-14
Point the watcher for the Prometheus config file to its directory

beorn7:
With the hardcoding of `/etc/prometheus`, any changes to the location
of the Prometheus config file would make it unwatched. This is
essentially applying `dirname` to the Prometheus config file and
therefore watches the directory in which the config file is located.

Please let me know if there is a less horrible way of doing `dirname`.

Another way would be to to a more invasive refactoring and define a `_config.prometheus_dir` variable and refer to it everywhere instead of hardcoding `/etc/prometheus`. But I first wanted to know your opinion before doing that more invasive refactoring.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus.libsonnet
@@ -52,7 +52,8 @@
       container.withArgs([
         '-v',
         '-t',
-        '-p=/etc/prometheus',
+        // Watch the directory in which the Prometheus config file is located.
+        '-p=' + std.substr(_config.prometheus_config_file, 0, std.reverse(std.findSubstr('/', _config.prometheus_config_file))[0]),
         'curl',
         '-X',
         'POST',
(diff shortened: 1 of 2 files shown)


prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -52,7 +52,8 @@
       container.withArgs([
         '-v',
         '-t',
-        '-p=/etc/prometheus',
+        // Watch the directory in which the Prometheus config file is located.
+        '-p=' + std.substr(_config.prometheus_config_file, 0, std.reverse(std.findSubstr('/', _config.prometheus_config_file))[0]),
```
malcolmholmes, line 56: you could do this more easily. What you want is the config dir. Do this in `lib/config.libsonnet`:
```
-    prometheus_config_file: '/etc/prometheus/prometheus.yml',
+    prometheus_config_dir: '/etc/prometheus',
+    prometheus_config_file: self.prometheus_config_dir + '/prometheus.yml',
```
Then you can refer to `_config.prometheus_config_dir` instead, and then, in `prometheus-ha-mixin.libsonnet` set `prometheus_config_dir` instead of `prometheus_config_file`. Reasonable?

beorn7:
Agreed. I'll do the refactoring with a prometheus_config_dir variable. Closing this.

PR #290, opened 2020-07-19
refactor: kausal.libsonnet to match k8s-alpha

Duologic:
Essentially a refactoring of kausal.libsonnet to work with k8s-alpha. This might break compat with ksonnet-lib, therefore moved to a new path/file.

ref: #215

Might want to diff locally against the original file: `diff ksonnet-util/kausal.libsonnet k8s-util/k8s.libsonnet`

Issue #215, opened 2020-02-12
ksonnet-util refactoring for clarity


malcolmholmes:
Having given some talks recently demonstrating Tanka, at the start, I needed to add: `(import 'ksonnet-util/kausal.libsonnet')`. This makes no sense - what is "ksonnet" and what is "kausal"? (Other than a cool company that has now merged with Grafana Labs!!).

I suggest we make some adjustments to `ksonnet-util` to make it cleaner for newcomers. These suggestions currently include:
- [x] move `jaeger.libsonnet` somewhere more sensible - it is not relevant where it is #291 
- [ ] rename `ksonnet-util/kausal.libsonnet` to ~`tanka-util/util.libsonnet`~ `k8s-mixin/defaults.libsonnet` #293 
- [ ] make the addition of labels (a) optional (b) configurable
- [ ] make `prometheus-ksonnet` use the defaulted config value for label name

sh0rez:
Counterproposal!

* ~~rename ksonnet-util/kausal.libsonnet to tanka-util/util.libsonnet~~
* integrate most of kausal.libsonnet into `kube-jsonnet/k`

malcolmholmes:
> Counterproposal!
> 
> * ~rename ksonnet-util/kausal.libsonnet to tanka-util/util.libsonnet~
> * integrate most of kausal.libsonnet into `kube-jsonnet/k`

Valid proposal. What would happen to the bits that aren't part of 'most'?

And how soon can this merge in be done?

ilpianista:
@sh0rez may you please clarify how this should work? Should the generator generate a kausal file for each version? Is there already something like that? If not, can you please provide the steps needed?

sh0rez:
Hi, generation of kausal.libsonnet is most probably not needed, as the file is only a patch.

Currently, there are three code generators around:
* `ksonnet/ksonnet-gen`, old and deprecated but still in use, up to k8s 1.14
* `kube-jsonnet/k`, fork of `ksonnet-gen`
  * maintained by us and some folks at RedHat
  * up to k8s 1.18
  * unhealthy codebase, hard to maintain
* `jsonnet-libs/k8s`: my very alpha, experimental generator, hopefully the future

All of these have in common, that they _generate_ the overall library, but rely on _hand-written_ patches to provide extra features, such as constructors.

`k.libsonnet` is such a hand written file, wrapping around `k8s.libsonnet`, the generated one.

[...]

PR #293, opened 2020-07-20
refactor: kausal.libsonnet to match k8s-alpha

Duologic:
Essentially a refactoring of kausal.libsonnet to work with k8s-alpha. This might break compat with ksonnet-lib, therefore moved to a new path/file.

ref: #215

(replaces #290)

Some background information: https://gist.github.com/Duologic/875c374cd0df5a04d059fc3e5de9a3dd#gistcomment-3384919

Depends on: https://github.com/jsonnet-libs/k8s/pull/8


k8s-mixin/defaults.libsonnet
```diff
@@ -33,8 +33,14 @@
       // Expose volumes type.
       volume+:: {
         // Make items parameter optional from fromConfigMap
-        fromConfigMap(name, configMapName, configMapItems={})::
-          super.fromConfigMap(name, configMapName, configMapItems),
+        fromConfigMap(name, configMapName, configMapItems=[])::
+          {
+            configMap+:
+              if configMapItems == [] then { items:: '' }
```
malcolmholmes, line 39: `items:: null` might say more clearly that this will not be a thing

k8s-mixin/defaults.libsonnet
```diff
@@ -233,7 +233,8 @@
         local serviceAccount = $.core.v1.serviceAccount,
 
         service_account:
-          serviceAccount.new(name),
+          serviceAccount.new(name) +
+          serviceAccount.metadata.withNamespace($._config.namespace),
```
malcolmholmes, line 248: is `$._config.namespace` an expectation of this `k8s-mixin`? It isn't an expectation of `k8s-alpha`, is it?

Duologic: yeah, that would be part of splitting out the utils into a separate util lib that needs its own config

Duologic:
replaced by #426

PR #295, opened 2020-07-21
altered alertmanager dns naming logic for clusters, added a gossip service

### LINKED ISSUE #294: prometheus-ksonnet/alertmanager needs a headless service if replicas > 1
Alertmanager configures its peer addresses when the replica count is > 1.  It follows the pod discovery convention used by Statefulsets.   The processes themselves use this convention to gossip to their peers.  

However, without a Headless Service named `alertmanager`, they can't reach one another.  

By default, the existing service is already named `alertmanager`.  It also exposes ports 9094 and 9093, which it probably doesn't need to.  It should probably only expose 80. 

Maybe we could name the new service `alertmanager-gossip` which has `ClusterIp` set to `None`, and only expose 9094 and 9093 on it,  when replicas is > 1.

rrichardson:
There are some opinionated changes in here.  This is driven from several points: 

1. The "cluster" name used by alertmanager to identify monitoring domains shouldn't be the same as the "cluster" that is found in the local DNS of Kubernetes. In fact, I'm pretty sure there is no way that the previous version could work without a single cluster named "cluster"
1. The service that alertmanager exposes probably shouldn't be the same one that the statefulset uses to gossip amongst itself. 

I'm not 100% clear on what we should do for the peers in the case of multiple AlertManager clusters . 
If AlertManager does actually need separate gossip pools for separate clusters, then I think we should have separate statefulsets. 

This config assumes that a single gossip pool can serve multiple AM clusters, which I thought was the case. 

/fixes #294

### DIFF
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -2,17 +2,9 @@
   local replicas = self._config.alertmanager_cluster_self.replicas,
   local isGlobal = self._config.alertmanager_cluster_self.global,
   local isGossiping = replicas > 1 || isGlobal,
-  local peers = if isGlobal
-  then
+  local peers = if isGossiping then
     [
-      'alertmanager-%d.alertmanager.%s.svc.%s.%s:%s' % [i, $._config.namespace, cluster, $._config.cluster_dns_tld, $._config.alertmanager_gossip_port]
-      for cluster in std.objectFields($._config.alertmanager_clusters)
-      if $._config.alertmanager_clusters[cluster].global
-      for i in std.range(0, $._config.alertmanager_clusters[cluster].replicas - 1)
-    ]
-  else if isGossiping then
-    [
-      'alertmanager-%d.alertmanager.%s.svc.%s.%s:%s' % [i, $._config.namespace, $._config.cluster_name, $._config.cluster_dns_tld, $._config.alertmanager_gossip_port]
+      'alertmanager-%d.alertmanager-gossip.%s.svc.%s:%s' % [i, $._config.namespace, $._config.cluster_dns_suffix, $._config.alertmanager_gossip_port]
       for i in std.range(0, replicas - 1)
     ]
   else [],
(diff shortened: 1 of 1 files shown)


beorn7:
Thanks for your contribution, and my apologies for the waiting time. I've just returned from vacations. I'd love the clearer and simpler version you are proposing here, but I remember vaguely that I ran into some problems that made me do things in the way I did. I'll discuss this with my peers here at Grafana to find out if there is something weird in the way we run things internally or if everything will just work as you propose it. Stay tuned…

prometheus-ksonnet/lib/alertmanager.libsonnet
```diff
@@ -2,17 +2,9 @@
...
     [
-      'alertmanager-%d.alertmanager.%s.svc.%s.%s:%s' % [i, $._config.namespace, cluster, $._config.cluster_dns_tld, $._config.alertmanager_gossip_port]
-      for cluster in std.objectFields($._config.alertmanager_clusters)
-      if $._config.alertmanager_clusters[cluster].global
-      for i in std.range(0, $._config.alertmanager_clusters[cluster].replicas - 1)
-    ]
-  else if isGossiping then
-    [
-      'alertmanager-%d.alertmanager.%s.svc.%s.%s:%s' % [i, $._config.namespace, $._config.cluster_name, $._config.cluster_dns_tld, $._config.alertmanager_gossip_port]
+      'alertmanager-%d.alertmanager-gossip.%s.svc.%s:%s' % [i, $._config.namespace, $._config.cluster_dns_suffix, $._config.alertmanager_gossip_port]
```
beorn7, line 7: This list comprehension looks like it won't iterate anymore over Alertmanagers in different clusters. But that's the whole idea of an HA Alertmanager cluster: Have instances in different zones, together creating a global AM cluster.

rrichardson: I think that incorporating multiple clusters in an HA arrangement is a good goal. 
However, it is not clear to me how the previous code would  achieve the intended result. Unless I'm missing something, it was attempting to use the `internal DNS` designation of each cluster, e.g. `cluster.local` as the differentiator of each cluster.  Maybe Grafana uses some mechanism for exposing the internal DNS tld for each of your federated clusters?  In my experience, the internal DNS identifiers in each cluster are private to the cluster, and different mechanisms are used for aggregating services across clusters.   Am I just getting the intent of that code wrong?

beorn7: Maybe. I'll have to discuss this with my infrastructure team mates.

To clarify: The code here was written specifically to run a cross-colo AM cluster, and said cluster has been running at Grafana with this code for quite some time.

Duologic: Hey @rrichardson

Beorn is right, we have CoreDNS configured that passes DNS queries to other clusters. I just merged https://github.com/grafana/jsonnet-libs/pull/472 that tries to create more manageable building blocks for a highly available alertmanager setup, please have a look if that can help you.

Maybe we can come up with a mixin for the service-based? For example a `withGossipService` function that modifies the alertmanager setup to match your use case? Then it doesn't break existing setups.

PR #297, opened 2020-07-22
Include config hashes for Grafana datasources and notification channels

robx:
Grafana only picks up changes to provisioned dashboards automatically.

For reference: https://grafana.com/docs/grafana/latest/administration/provisioning/#dashboards
```
# <int> how often Grafana will scan for changed dashboards
updateIntervalSeconds: 10
```

Also I ran into this myself, with Grafana not picking up reconfigured datasources.

### DIFF
--- a/prometheus-ksonnet/grafana/deployment.libsonnet
+++ b/prometheus-ksonnet/grafana/deployment.libsonnet
@@ -75,13 +75,12 @@
 
   grafana_deployment:
     deployment.new('grafana', 1, [$.grafana_container]) +
-    // Use configMapVolumeMount to automatically include the hash of the config
-    // as an annotation.  No need to use for others, Grafana will pick up
-    // changes there.
+    // Use configMapVolumeMount to automatically include the hash of the config.
+    // No need to use for dashboards, Grafana will pick up changes there.
     $.util.configMapVolumeMount($.grafana_config_map, '/etc/grafana-config') +
+    $.util.configMapVolumeMount($.grafana_datasource_config_map, '%(grafana_provisioning_dir)s/datasources' % $._config) +
+    $.util.configMapVolumeMount($.notification_channel_config_map, '%(grafana_provisioning_dir)s/notifiers' % $._config) +
     $.util.configVolumeMount('grafana-dashboard-provisioning', '%(grafana_provisioning_dir)s/dashboards' % $._config) +
-    $.util.configVolumeMount('grafana-datasources', '%(grafana_provisioning_dir)s/datasources' % $._config) +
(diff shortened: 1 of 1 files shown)


Duologic:
I've resolved this in the past with sidecar that watches for changes and let it eload through an API call. This would be much less disruptive then restarting the container.

You can use this as a mixin:

```jsonnet
{
  grafana_watch:: {
    reload_type:: error 'must specify reload_type',
    watch_dir:: error 'must specify watch_dir',

    local container = $.core.v1.container,
    local images = $._images,

    container::
      container.new('watch-%s' % self.reload_type, $._images.watch) +
      container.withArgs([
        '-v',
        '-t',
        '-p=%s' % self.watch_dir,
        'curl',
        '-X',
        'POST',
        '--fail',
        '-o',
        '-',
        '-sS',
        'http://admin:admin@localhost/api/admin/provisioning/%s/reload' % self.reload_type,
      ]),
  },

  grafana_datasources_watch:: $.grafana_watch {
    reload_type:: 'datasources',
    watch_dir:: '%s/datasources' % provisioningDir,
  },

  grafana_notifiers_watch:: $.grafana_watch {
    reload_type:: 'notifications',
    watch_dir:: '%s/notifiers' % provisioningDir,
  },

  grafana_deployment+:
    deployment.spec.template.spec.withContainersMixin([
      self.grafana_datasources_watch.container,
      self.grafana_notifiers_watch.container,
    ]),
}
```

robx:
@Duologic interesting! Do you suggest including that in jsonnet-libs, or doing this "by hand" on a case-by-case basis? If the latter, I'd close this and make a small PR to just reword the comment that currently suggests datasources would be picked up automatically.

Duologic:
I think it might be interesting as a PR here, but not change the default behavior but as a mixin, like so:

```jsonnet
local grafana_watch_sidecar = import 'prometheus-ksonnet/grafana/watch_sidecar.libsonnet';
{
  grafana_deployment+: grafana_watch_sidecar,
}
```

This is just an idea.

Duologic:
https://github.com/grafana/jsonnet-libs/blob/24760f1f722d105e246e9112ea807c14f6c08457/prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet is intended to work like that.

malcolmholmes:
I don't see why it shouldn't be included in the raw, grafana lib. The behaviour of not reloading datasources is counter-intuitive, and likely wastes peoples' time until they eventually realise they need to restart Grafana.

We have a watcher for Prometheus, so it seems totally reasonable to have this in prometheus-ksonnet itself, for all users.

PR #299, opened 2020-07-23
fix(prometheus-ksonnet): add ns/pod/container

Duologic:
This adds labels for namespace, pod and container to metrics coming from
job=default/kubernetes.

The etcd-mixin uses `grpc_server_*` to calculate some of its alerts,
these metrics originate from job=default/kubernetes and we are missing the
namespace label. In our current routing setup, we need the namespace
label to route our alerts to the right team.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -364,11 +364,30 @@
           insecure_skip_verify: $._config.prometheus_insecure_skip_verify,
         },
 
-        relabel_configs: [{
-          source_labels: ['__meta_kubernetes_service_label_component'],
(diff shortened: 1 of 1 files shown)


Duologic:
Just  realized this won't work, right? :-|  It will just get the labels from default/kubernetes.

Duologic:
Just  realized this won't work, right? :-|  It will just get the labels from default/kubernetes.

beorn7:
There is only one target found in this scrape config. It has namespace="default" (which has the same routing as no namespace), and no pod name and no container name.

The etcd alerts are based on metrics with a correct namespace label, but the alerts themselves aggregate away the namespace label. The fix is https://github.com/etcd-io/etcd/pull/12122

PR #317, opened 2020-08-06
Configmap per mixin for Prometheus

malcolmholmes:
Makes a config map per mixin (and a single one for 'root' configs).

This appears to work (not yet applied to a cluster though).

It does not (yet) work against an HA cluster. That's the next task.


prometheus-ksonnet/lib/prometheus-configmap.libsonnet
```diff
@@ -1,28 +1,47 @@
...
+                              configMap.new('%s-config' % self.name) +
+                              configMap.withData({
+                                'prometheus.yml': $.util.manifestYaml(prometheus_config),
+                                'alerts.rules': $.util.manifestYaml(prometheusAlerts),
+                                'recording.rules': $.util.manifestYaml(prometheusRules),
+                              }),
+                            ] +
+                            [
+                              local mixin = $.mixins[mixinName] + util.emptyMixin;
+                              configMap.new('%s-%s-config' % [self.name, util.normalise(mixinName)]) +
```
Duologic, line 39: Instead of looping over the mixins twice (here and in the configVolumeMount) and normalizing the name, you could do it once in a list and and use that list as an input for both? I assume it would cost less computing power and thus run faster.

Duologic: Aha, you already have that, in the `rule_files` on line 11 here.

tomwilkie:
We already have per-mixin config maps for dashboards - this is just for rules and alerts, right?

tomwilkie:
I think you're missing a change to https://github.com/grafana/jsonnet-libs/blob/master/prometheus-ksonnet/lib/prometheus-config.libsonnet#L7 to tell Prometheus that these new files exist.

malcolmholmes:
> I see it now. Perhaps we could move it to prometheus-config.libsonnet so its all in one place?

Makes sense. Did it.

And yes, this is for rules and alerts only.

prometheus-ksonnet/lib/prometheus-config.libsonnet
```diff
@@ -7,7 +8,19 @@
...
+        '%s/recording.rules' % util.normalise(mixinName),
+      ]
+      for mixinName in std.objectFields($.mixins)
+    ]) + std.flattenArrays([
+      [
+        '%s/alerts.rules' % util.normalise(mixinName),
+        '%s/recording.rules' % util.normalise(mixinName),
+      ]
+      for mixinName in std.objectFields($.mixins)
+    ]),
```
tomwilkie, line 23: So good you did it twice?

prometheus-ksonnet/lib/prometheus-configmap.libsonnet
```diff
@@ -9,20 +11,28 @@
...
+                              configMap.withData({
+                                'alerts.rules': $.util.manifestYaml(mixin.prometheusAlerts),
+                                'recording.rules': $.util.manifestYaml(mixin.prometheusRules),
+                              },)
+                              for mixinName in std.objectFields($.mixins)
+                            ],
```
tomwilkie, line 36: The indentation is a little ugly - can you add a new line after the `:`?

beorn7:
I studied the `tk diff`. I believe this is doing what is intended, but I have seen two weird issues:

- The billing Prometheus gets all the rules from the mixins now because we only override `prometheusRules` and `prometheusAlerts`. We need to adjust that, ideally so that the billing Prometheus doesn't even see all the mixins. (We should not install all the then-unused config maps for the billing Prometheus, either.) This will most likely happen in `lib/default/prometheus.libsonnet` (in deployment_tools).
- The `runbookURLPattern` override in deployment_tools is just for the kubernetes mixin. But somehow, it also affects alerts from the Prometheus mixin (which I don't understand in the first place). With the change here, some alerts from the Prometheus mixin still seem to show up in the top-level `alerts.rules`, but without the runbook. I might misread something there.

The diff is very hard to read because it essentially dumps everything twice, once as a deletion in the old config map, and then as an addition in all the now config maps. I suggest to apply this to only one of two HA instances and do a live comparison of the two Prometheus servers in a dev cluster.

prometheus-ksonnet/lib/prometheus-config.libsonnet
malcolmholmes: oops, fixed :-)

malcolmholmes:
This proved excessively complex to get to be a no-op. Closing in favour of newer efforts.

PR #319, opened 2020-08-07
Don't wrap datasources in an array

malcolmholmes:
At present, we are placing datasource json into an object that pre-supposes we will use Provisioning to deploy. It puts the datasource into an array, and that means that you cannot patch the datasource (because patching arrays is tough).

This PR moves the boilerplate needed for the provisioning API to right before the JSON is rendered as YAML, meaning any datasources in `grafanaDatasources` will be patchable.

This should effectively be a no-op. It isn't long ago that everything in `grafanaDatasources` was rendered YAML strings. So the json objects in `grafanaDatasources` option that we are changing here has only been in the library for a week or so, and is thus likely unused.

This is attempting to move us to an agnostic representation of Grafana resources, such that code specifying datasources can be consumed outside of `prometheus-ksonnet`, e.g. by [Grizzly](https://github.com/grafana/grizzly) to push datasources directly to Grafana, e.g. by people not even using Kubernetes, or for people who'se Grafana instance isn't Kubernetes managed (e.g. Grafana Cloud).

### DIFF
--- a/prometheus-ksonnet/grafana/datasources.libsonnet
+++ b/prometheus-ksonnet/grafana/datasources.libsonnet
@@ -34,7 +34,10 @@
   */
   grafana_add_datasource(name, url, default=false, method='GET')::
     configMap.withDataMixin({
-      ['%s.yml' % name]: $.util.manifestYaml($.grafana_datasource(name, url, default, method)),
+      ['%s.yml' % name]: $.util.manifestYaml({
+        apiVersion: 1,
+        datasources: [$.grafana_datasource(name, url, default, method)],
+      }),
     }),
 
   // Generates yaml string containing datasource config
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/grafana/datasources.libsonnet
```diff
@@ -38,29 +34,28 @@
...
-        version: 1,
-        editable: false,
-        basicAuth: true,
-        basicAuthUser: username,
-        basicAuthPassword: password,
-        jsonData: {
-          httpMethod: method,
-        },
-      }],
+  grafana_datasource_with_basicauth(name, url, username, password, default=false, method='GET', type='prometheus'):: {
```
Duologic, line 44: Can't this call `grafana_datasource()` and mixin the auth items?

malcolmholmes: that comes next. One thing at a time.

PR #320, opened 2020-08-10
cert-manager through std.native('helmTemplate')

Duologic:
ref: https://github.com/grafana/tanka/pull/336


Duologic:
Replaced by https://github.com/grafana/jsonnet-libs/pull/330

PR #324, opened 2020-08-13
refactor(etcd-cluster): use core.v1 resources

Duologic:
Uses:
core.v1.podAffinityTerm
core.v1.envVar
rbac.v1.policyRule

removes .mixin

works with k8s-alpha


Duologic:
Not needed, kausal.libsonnet deals with that.

PR #328, opened 2020-08-14
feat: add helm-util

Duologic:
Requires: https://github.com/grafana/tanka/pull/336

### DIFF
--- a/helm-util/README.md
+++ b/helm-util/README.md
@@ -1,0 +1,50 @@
+# helm-util
+
+Provides utilities for using helm in jsonnet.
+
+## Helmraiser
+
+Helmraiser refers to Go package that provides helmTemplate native function through Tanka for generating a Jsonnet
+library from any Helm chart. The goal of this function is to keep track of upstream changes in a Helm chart while
+retaining Jsonnet changes (read: mixins) on top of them.
+
+## Example usage
+
+```
+jb install github.com/grafana/jsonnet-libs/helm-util
+```
+
+```
+local helm = import 'github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet';
+
+{
+    version:: 'v0.16.0',
+
+    values:: {
+      installCRDs: true,
+      global: {
+        podSecurityPolicy: {
+          enabled: true,
+          useAppArmor: false,
+        },
+      },
+    },
+
+    labels:: {
+      'custom_label': 'example',
+    }
+
+    chart: helm.helmTemplate(
+      name='cert-manager',
+      chart='jetstack/cert-manager',
+      $.conf,
+      conf={
+        values: $.values,
+        flags: [
+          '--version=%s % $.version',
+        ],
+      },
+      $.labels
+    ),
+}
+```
--- a/helm-util/README.md
+++ b/helm-util/README.md
@@ -4,7 +4,7 @@ Provides utilities for using helm in jsonnet.
 
 ## Helmraiser
 
-Helmraiser refers to Go package that provides helmTemplate native function through Tanka for generating a Jsonnet
+Helmraiser refers to a Go package that provides helmTemplate native function through Tanka for generating a Jsonnet
 library from any Helm chart. The goal of this function is to keep track of upstream changes in a Helm chart while
 retaining Jsonnet changes (read: mixins) on top of them.
 
@@ -18,33 +18,27 @@ jb install github.com/grafana/jsonnet-libs/helm-util
 local helm = import 'github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet';
 
 {
-    version:: 'v0.16.0',
+  version:: 'v0.16.0',
 
-    values:: {
-      installCRDs: true,
-      global: {
-        podSecurityPolicy: {
-          enabled: true,
-          useAppArmor: false,
-        },
+  values:: {
+    installCRDs: true,
+    global: {
+      podSecurityPolicy: {
+        enabled: true,
+        useAppArmor: false,
       },
     },
+  },
 
-    labels:: {
-      'custom_label': 'example',
-    }
-
-    chart: helm.helmTemplate(
-      name='cert-manager',
-      chart='jetstack/cert-manager',
-      $.conf,
-      conf={
-        values: $.values,
-        flags: [
-          '--version=%s % $.version',
-        ],
-      },
-      $.labels
-    ),
+  chart: helm.template(
+    name='cert-manager',
+    chart='jetstack/cert-manager',
+    conf={
+      values: $.values,
+      flags: [
+        '--version=%s % $.version',
+      ],
+    },
+  ),
 }
 ```
--- a/helm-util/helm.libsonnet
+++ b/helm-util/helm.libsonnet
@@ -1,0 +1,57 @@
+local d = import 'github.com/sh0rez/docsonnet/doc-util';
+{
+  local this = self,
+
+  '#': d.pkg(
+    name='helm-util',
+    url='github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet',
+    help='`helm-util` provides utilities for using helm in jsonnet',
+  ),
+
+  '#helmTemplate': d.fn(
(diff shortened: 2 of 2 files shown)


helm-util/README.md
```diff
@@ -0,0 +1,50 @@
+# helm-util
+
+Provides utilities for using helm in jsonnet.
+
+## Helmraiser
+
+Helmraiser refers to Go package that provides helmTemplate native function through Tanka for generating a Jsonnet
```
malcolmholmes, line 7:
```suggestion
Helmraiser refers to a Go package that provides helmTemplate native function through Tanka for generating a Jsonnet
```

helm-util/README.md
```diff
@@ -0,0 +1,50 @@
...
+    },
+
+    labels:: {
+      'custom_label': 'example',
+    }
+
+    chart: helm.helmTemplate(
+      name='cert-manager',
+      chart='jetstack/cert-manager',
+      $.conf,
```
malcolmholmes, line 40: what does this refer to?

helm-util/README.md
```diff
@@ -0,0 +1,50 @@
...
+      conf={
+        values: $.values,
+        flags: [
+          '--version=%s % $.version',
+        ],
+      },
+      $.labels
+    ),
+}
+```
```
malcolmholmes, line 50: could we document patching the resultant json?

helm-util/helm.libsonnet
```diff
@@ -0,0 +1,57 @@
+local d = import 'github.com/sh0rez/docsonnet/doc-util';
+{
+  local this = self,
+
+  '#': d.pkg(
```
sh0rez, line 5: Docsonnet keys should be private (`::`), because they never need to be evaluated out of docsonnet

helm-util/helm.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+    name='helm-util',
+    url='github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet',
+    help='`helm-util` provides utilities for using helm in jsonnet',
+  ),
+
+  '#helmTemplate': d.fn(
```
sh0rez, line 11: `helm.helmTemplate` is doubled, let's do `helm.template` instead

helm-util/helm.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+    '`helmTemplate` wraps the helmTemplate native function and returns a rendered chart as a map of kubernetes manifests',
```
sh0rez, line 12: As a user, I don't quite care about the native func. I want to know what happens when I call this. Maybe like so?

```markdown
`template` expands the Helm Chart to it's underlying resources and returns them in an `Object`,
so they can be consumed and modified from within Jsonnet.

This functionality requires Helmraiser support in Jsonnet (e.g. using Grafana Tanka) and also
the `helm` binary installed on your `$PATH`.
```

helm-util/helm.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+      d.arg('name', d.T.string),
+      d.arg('chart', d.T.string),
+      d.arg('conf', d.T.object),
+      d.arg('labels', d.T.object),
+    ]
+  ),
+  helmTemplate(name, chart, conf={}, labels={})::
+    this.patchLabels(
+      std.native('helmTemplate')(name, chart, conf),
+      labels { 'app.kubernetes.io/managed-by': 'Helmraiser' }
```
sh0rez, line 23: Do we need to set this? Tanka does not do so usually, because it let's kubectl do it's thing, being transparent to the cluster.

Can't think of reasons Helmraiser resources should be special in any way

helm-util/README.md
```diff
@@ -0,0 +1,50 @@
...
+      'custom_label': 'example',
+    }
+
+    chart: helm.helmTemplate(
+      name='cert-manager',
+      chart='jetstack/cert-manager',
+      $.conf,
+      conf={
+        values: $.values,
+        flags: [
```
sh0rez, line 43: Let's instead accept an Object here and use the `mapToFlags` code (copy from `kausal.libsonnet`) to convert to flags.

Flags are an implementation detail after all and if we ever move off them (e.g. vendoring Helm code instead of shelling out) it's far easier working with objects than having to translate flags into variables

Also doesn't need string templating :)

helm-util/helm.libsonnet
```diff
@@ -0,0 +1,57 @@
...
+  '#helmTemplate': d.fn(
+    '`helmTemplate` wraps the helmTemplate native function and returns a rendered chart as a map of kubernetes manifests',
+    [
+      d.arg('name', d.T.string),
+      d.arg('chart', d.T.string),
+      d.arg('conf', d.T.object),
+      d.arg('labels', d.T.object),
+    ]
+  ),
+  helmTemplate(name, chart, conf={}, labels={})::
```
sh0rez, line 20: This function signature is already rather long. Given you already have an opts-like object (conf) in here, I think labels should be part of that instead.

Generally spoken once you have a lot of function parameters, I prefer to only have these that are **required** as parameters, and those that are optional in an final object parameter

helm-util/helm.libsonnet
Duologic: It is part of the standard labels (https://helm.sh/docs/chart_best_practices/labels/#standard-labels), I just want to make clear this is not managed-by: helm, because it is not.

helm-util/README.md
Duologic: As mentioned earlier, I don't want to implement it now, having the ability to play with the raw flags at this stage of development is far more valuable to me than abstracting it away.

helm-util/helm.libsonnet
Duologic: I'd rather drop the `labels` parameter completely than put it into `conf`, these arguments have nothing to do with each other in lower level code.

Duologic: Maybe we can have a `withLabels` function on here?

helm-util/helm.libsonnet
Duologic: Just mimicking the docs as written in k8s-alpha ;)

helm-util/helm.libsonnet
sh0rez: Not having this label at all should be a good start. If you wan't more clearness, Tanka should label everything it manages.

helm-util/README.md
sh0rez: We are creating an API here, which should stay as stable as possible.

Having those flags here is not acceptable for a good API IMO. Furthermore, `mapToFlags` does not abstract things away, it's a more generic way to express the same. There is no functional difference, you still have access to all flags you might want

helm-util/helm.libsonnet
sh0rez: Right, then let's remove labels entirely from `helmTemplate` and have it as a separate func

helm-util/helm.libsonnet
sh0rez: In k8s-alpha, these are already created as private by the generator. `+:` maintains previous visibility, so this is not a problem there.

Here you create those keys the first time, it inherits public visibility, which is an issue when importing the library like so:

```
(import "helm-util/helm.libsonnet") {
}
```

Because now downstream code needs access to `doc-util` and also the docsonnet keys will be in the output

helm-util/README.md
sh0rez: To not hold this any longer, let's go ahead with this version and deprecate it before releasing Tanka 1.0

Duologic: I'm 100% sure flags/command line arguments can be represented as an array of strings (for reference: the k8s container spec, this is also represented args like this). If we want to do more specific things and use a map in future releases, we can either parse the flag strings or introduces a new map that holds specific options as part of the 'conf' map argument in the native command, it should leave us enough room for growth.

helm-util/helm.libsonnet
Duologic: I've moved this up into a variable that can be overwritten easily.

malcolmholmes:
The last commit addresses my comment, which was to expose 'labels' so that users can override them should they wish to.

PR #329, opened 2020-08-17
Split prom configmap into three

malcolmholmes:
This patch puts prometheus.yml, alerts.rules and recording.rules into their own configmaps.

My previous attempt (#317) at fixing the config-map full issue for Prometheus was great in theory, but our config consuming it was too complicated for me to roll this out in short order.

Therefore, this is a much simpler version. It does not buy us as much time, but it allows for a doubling of the size of our configs, which is worth having.

This has been tested to the point of `tk show` looking correct, so far.

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -5,8 +5,8 @@
     },
 
     rule_files: [
-      'alerts.rules',
-      'recording.rules',
+      'alerts/rules',
+      'recording/rules',
     ],
 
     alerting: {
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -5,8 +5,8 @@
     },
 
     rule_files: [
-      'alerts/rules',
-      'recording/rules',
+      'alerts/alerts.rules',
+      'recording/recording.rules',
     ],
 
     alerting: {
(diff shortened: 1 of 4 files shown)


prometheus-ksonnet/lib/prometheus.libsonnet
```diff
@@ -75,7 +75,10 @@
     local volumeMount = $.core.v1.volumeMount,
 
     prometheus_config_mount::
-      $.util.configVolumeMount('%s-config' % self.name, _config.prometheus_config_dir),
+      $.util.configVolumeMount('%s-config' % self.name, _config.prometheus_config_dir)
+      + $.util.configVolumeMount('%s-alerts' % self.name, _config.prometheus_config_dir + '/alerts')
+      + $.util.configVolumeMount('%s-rules' % self.name, _config.prometheus_config_dir + '/recording')
```
beorn7, line 80: Shouldn't this be `/rules` is in the HA setup above?

prometheus-ksonnet/lib/prometheus-config.libsonnet
```diff
@@ -5,8 +5,8 @@
     },
 
     rule_files: [
-      'alerts.rules',
-      'recording.rules',
+      'alerts/rules',
+      'recording/rules',
```
beorn7, line 9: Shouldn't this be `alerts/alerts.rules` and `recording/recording.rules`?

prometheus-ksonnet/lib/prometheus.libsonnet
malcolmholmes: yes!

prometheus-ksonnet/lib/prometheus-config.libsonnet
malcolmholmes: right again! Had to think about this one though.

tomwilkie:
Beorn and Jeroen has reviewed, I don't think you need one from me, so removing this to stop the nag.

PR #333, opened 2020-08-31
Configure securityContext for AlertManager statefulset.

mplzik:
AlertManager became a non-root image some time ago. Since in StatefulSet
variant it writes some state to an external PV, which is usually
ext4-formatted and root-owned, this prevents AM from writing to the PV.

Prometheus had a similar issue which has been resolved by adding
`securityContext` to the StatefulSet in https://github.com/grafana/jsonnet-libs/pull/244; let's do the same for
AlertManager.


sh0rez (approved):
Don't know the full context of our alertmanager deployment, but also looking at https://github.com/grafana/jsonnet-libs/pull/244 this seems reasonable. 

LGTM

beorn7:
Just curious: How did you notice that AM couldn't write to the PV? Was there any alert firing for it? (I suspect it didn't, but we should have one…)

PR #337, opened 2020-09-10
Configure Grafana outside grafana lib

malcolmholmes:
We have a library here for installing Grafana that has some really neat features in it. However, this library contains a couple of things that make it specific to a Prometheus case. Therefore, to use it without Prometheus, you have to know that you have to remove things.

This PR switches those to sensible defaults for Grafana, and then moves the Prometheus specific configs into Prometheus specific locations. If you want to use Grafana on its own, you can. If you want to use Grafana with Prometheus, you can. If you want to use Prometheus without Grafana, you can.

This also can act as a precursor to moving the Grafana lib one level up, so that it isn't in the `prometheus-ksonnet` package - it no-longer needs to be there and has a more general purpose value.

This is a no-op on all the clusters I tested. This should also help @jdbaldry I believe!

### DIFF
--- a/prometheus-ksonnet/grafana/config.libsonnet
+++ b/prometheus-ksonnet/grafana/config.libsonnet
@@ -1,4 +1,23 @@
 {
+  _config+:: {
+    admin_services+:: [
+      {
+        title: 'Grafana (Light)',
+        path: 'grafana',
+        params: '/?search=open&theme=light',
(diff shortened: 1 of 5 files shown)


jdbaldry:
Thanks Malcolm, looks like a clean first step towards separating grafana and moving towards prometheus-ksonnet being an opinionated composition of Prometheus, Alertmanager, and Grafana.

jdbaldry (approved):

PR #339, opened 2020-09-14
Make grafana a library in its own right

malcolmholmes:
This has to be the simplest ever PR. Given that `prometheus-ksonnet` imports `grafana/grafana.libsonnet`, moving it from inside the `prometheus-ksonnet` directory to `vendor/grafana` works just the same.

Moving this code outside of its parent has the benefit of allowing us to offer a generic "mixin-aware" Grafana library, independent of Prometheus, whilst continuing to support the `prometheus-ksonnet` use-case as before.

N.B. The vendoring is not tested yet, as it requires it to be merged first. We could either do that in quick succession, or do this in two PRs (copy/delete instead of move).


sh0rez (approved):
LGTM.

You can test vendoring by locally changing the version to this branch and running install. If that works, it's good to go

jdbaldry (approved):
LGTM

tomwilkie (approved):

malcolmholmes:
Finally got around to confirming this branch as @sh0rez suggested (after much needed rebasing). Worked fine.

tomwilkie (approved):

malcolmholmes:
Superceded by #449 which does the separation in a much more intelligent way.

PR #342, opened 2020-09-15
Add the polystat panel to grafana instances

tomwilkie:
Want to use this in the node-mixin.

### DIFF
--- a/prometheus-ksonnet/grafana/deployment.libsonnet
+++ b/prometheus-ksonnet/grafana/deployment.libsonnet
@@ -15,7 +15,7 @@
   },
 
   grafana_plugins+:: [
-    "grafana-polystat-panel",
+    'grafana-polystat-panel',
   ],
 
   local container = $.core.v1.container,
--- a/prometheus-ksonnet/grafana/deployment.libsonnet
+++ b/prometheus-ksonnet/grafana/deployment.libsonnet
@@ -14,7 +14,9 @@
     grafana_notification_channel_labels: {},
   },
 
-  grafana_plugins+:: [],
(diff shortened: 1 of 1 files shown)


malcolmholmes:
This is tough. Would rather not add it to prometheus-ksonnet itself, just to support a single mixin.

It seems like this is a valid extension to the mixin model: add `grafana-plugins` to the supported options, then can add this directly to the `node-mixin`. Would want to support the new plugin provisioning method too. Does that work?

tomwilkie:
Yeah, I think I can make that work.

malcolmholmes:
https://grafana.com/docs/grafana/latest/administration/provisioning/#plugins for the new way of provisioning that we should ideally support.

robx:
> https://grafana.com/docs/grafana/latest/administration/provisioning/#plugins for the new way of provisioning that we should ideally support.

Note that plugin provisioning is only necessary for app plugins; datasource and panel plugins only need to be installed. So adding support for plugin provisioning shouldn't be necessary here.

(I agree that it would be good to support though. An example of doing it explicitly is the [alerting UI on dev-us-central1](https://github.com/grafana/deployment_tools/blob/00569f2cf9633cf4bbc1bb8915fb3805c00cef99/ksonnet/environments/default/dev-us-central1.default/main.jsonnet#L126).)

tomwilkie:
Yes, thank you Rob - this is intended for panel plugins right now.

@malcolmholmes this seems to work in my testing, ready for another look if you don't mind?

malcolmholmes (approved):
Looks good to me.

PR #343, opened 2020-09-18
A generic way of mapping k8s label/annotations to prom labels

bergquist:
This will allow devs to add Prometheus label by exposing k8s labels/annotations in a generic fashion rather than having unique scrape configs. 

ref https://github.com/grafana/hosted-grafana/pull/216

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -171,7 +171,8 @@
             replacement: '__param_$1',
           },
 
-          // Generic mapping of k8s label/annotation to prometheus labels.
+          // Map all K8s labels/annotations starting with
+          // 'prometheus.io/label-' to Prometheus labels.
           {
             regex: '__meta_kubernetes_pod_label_prometheus_io_label_(.+)',
             action: 'labelmap',
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -171,6 +171,17 @@
             replacement: '__param_$1',
           },
 
+          // Generic mapping of k8s label/annotation to prometheus labels.
+          {
+            regex: '__meta_kubernetes_pod_label_prometheus_io_label_(.+)',
+            action: 'labelmap'
+          },
+
+          {
+            regex: '__meta_kubernetes_pod_annotation_prometheus_io_label_(.+)',
+            action: 'labelmap'
+          },
+
           // Drop pods with phase Succeeded or Failed
           {
             source_labels: ['__meta_kubernetes_pod_phase'],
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/lib/prometheus-config.libsonnet
```diff
@@ -171,6 +171,17 @@
             replacement: '__param_$1',
           },
 
+          // Generic mapping of k8s label/annotation to prometheus labels.
+          {
+            regex: '__meta_kubernetes_pod_label_prometheus_io_label_(.+)',
```
bergquist, line 176: I found a request on the mailing list and It didn't include the`replacement` field.
https://github.com/prometheus/prometheus/blob/86426c05669417cf1012fbc3a40fb7e0f5a8a101/documentation/examples/prometheus-kubernetes.yml#L113-L114

beorn7: Indeed. That's because `$1` is the default value for `replacement`. \o/

prometheus-ksonnet/lib/prometheus-config.libsonnet
```diff
@@ -171,6 +171,17 @@
             replacement: '__param_$1',
           },
 
+          // Generic mapping of k8s label/annotation to prometheus labels.
```
beorn7, line 174: Perhaps mention the precise prefix requirement here?

E.g. "Map all K8s labels/annotations starting with 'prometheus.io/label-<name>' to Prometheus labels."

beorn7 (commented):
Since this file is the brainchild of @tomwilkie , it would be great if he could sanity-check this.

tomwilkie:
In Loki we just add all labels (not those prefixed with io.prometheus.): https://github.com/grafana/loki/blob/master/production/ksonnet/promtail/scrape_config.libsonnet#L26

Should we do that same here?

bergquist:
> Should we do that same here?

I don't mind it but we would have to go thru all current deployments and make sure it works for them. 

This seems like a generic solution that moves us forward.

tomwilkie:
Yeah thats fair.  I'd be tempted to JFDI (we can roll it back if we break something). 

But LGTM either way, up to you Carl.

beorn7:
I think we have quite a bit of label proliferation at this point. Not sure how much this will confuse people at this point. But I'm fine either way. I like if labels are equivalent between K8s and P8s.

In either case, annotations have to go with the prefix solution, I think, because we are already doing something like the param thing there. (And Loki is not doing the annotations at all, which is also telling us something.)

bergquist:
> I like if labels are equivalent between K8s and P8s.

I do agree that this sounds like a very good idea going forward. Since I'm not in a rush to start scraping metrics we should try to get to this point

>  I'd be tempted to JFDI

I am more than happy to stand next to whoever does that as I'm not super comfortable deploying such change :)

beorn7:
So then let's change this PR to directly map all labels but keep the way annotations are handled. I'll leave comments to illustrate what I mean.

beorn7 (approved):

beorn7:
This should now be very low-risk to roll-out. So I suggest to merge and roll-out at your descretion.

PR #346, opened 2020-09-22
Revert "fix(cert-manager): set calledFrom explicitly"

sh0rez:
This reverts commit 8d04b4834a47006016a23717a1f72fcca7196836.

The issue this commits fixes was caused by a broken dependency setup, not because of a problem with this library.

`(import "helm...").new(std.thisFile)` is the user-facing API, and our libraries should use that. `calledFrom` is an implementation detail and Tanka gives no stability guarantee on that.

### DIFF
--- a/cert-manager/main.libsonnet
+++ b/cert-manager/main.libsonnet
@@ -1,4 +1,5 @@
-local helm = (import 'github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet');
+local helm = (import 'github.com/grafana/jsonnet-libs/helm-util/helm.libsonnet').new(std.thisFile);
+
 {
   values:: {
     installCRDs: if $._config.custom_crds then false else true,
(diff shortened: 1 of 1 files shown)


Duologic (approved):
It is a bit ridiculous that we have to pass the calledFrom from Jsonnet, can't we do something similar to jpath.resolve for this?

Anyway, PR is probably fine.

sh0rez:
@Duologic not to my knowledge.

By design, we require Helm Charts to be exclusively loaded from the local filesystem, to keep the configuration tree self contained. Futhermore, these must be referred to using a path relative to the consumer (the one calling `helm.template`) to keep it independent from the current working directory.
The Jsonnet API does not provide us the information where the native func was called from, and even if it did, this would always yield `helm.libsonnet`, which is pointless.
As a workaround, we require users to instantiate the helm util once using their "own location", so that chart resolution works.
This allows libraries to vendor their own charts in tree.

PR #347, opened 2020-09-27
Allow patching by name/kind after helm.template

malcolmholmes:
Currently, `patchKubernetesObjects` patches *all* objects. I want to add a node-selector to all StatefulSets. This PR adds an optional `kind` and `name` elements which allow us to patch specific resources rather than all of them.

### DIFF
--- a/helm-util/helm.libsonnet
+++ b/helm-util/helm.libsonnet
@@ -51,7 +51,7 @@ local d = import 'github.com/sh0rez/docsonnet/doc-util/main.libsonnet';
     then
       // a Kubernetes object is characterized by having an apiVersion and Kind
       if std.objectHas(object, 'apiVersion') && std.objectHas(object, 'kind')
-         && (kind == null || object.kind == std.trace('%s: %s' % [object.kind, kind], kind)) && (name == null || object.metadata.name == name)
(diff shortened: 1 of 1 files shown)


sh0rez (changes requested):
I feel like this is slowly creeping the scope of the package.

imo helm-util is supposed to provide helm rendering support, and not much beyond that.

While the code for object mutation, including tree traversal is cool, this really should be a package in its own right, or part of kausal or k8s-util

malcolmholmes:
that's fair enough. Could be a dependency of helm-util too. `object-util` for example.

Duologic:
I think we can add this here for now, but yes, ideally this moves out of helm-util, sounds like it might belong in k8s-util once we get a move on that.

Duologic:
@malcolmholmes Is this going anywhere?

malcolmholmes:
Yes, I for one need it. I totally agree that there's a heap of useful functionality (e.g. deep patching functions, generic functions to patch objects in lists, etc) that would make sense inside a more generic `object-utils` or `tanka-utils`. But that's all for the future.

malcolmholmes:
i.e. can we approve/merge this?

Duologic (approved):
Agreed, let's merge this with the prospect of a more generalized library.

sh0rez (approved):

PR #344, opened 2020-09-18
Add a dedicated scrape config for kube-dns

beorn7:

PR #348, opened 2020-09-28
fix(prometheus-ksonnet): kube-dns is now prefixed

Duologic:
The scrape config was changed (#344), this makes sure the
KubeVersionMismatch warning stops firing.

### DIFF
--- a/prometheus-ksonnet/mixins.libsonnet
+++ b/prometheus-ksonnet/mixins.libsonnet
@@ -16,6 +16,7 @@
           kubeControllerManagerSelector: 'job="kube-system/kube-controller-manager"',
           kubeApiserverSelector: 'job="kube-system/kube-apiserver"',
           podLabel: 'instance',
+          notKubeDnsCoreDnsSelector: 'job!~"kube-system/kube-dns|coredns"',
         },
       },
 


prometheus-ksonnet/mixins.libsonnet
```diff
@@ -16,6 +16,7 @@
           kubeControllerManagerSelector: 'job="kube-system/kube-controller-manager"',
           kubeApiserverSelector: 'job="kube-system/kube-apiserver"',
           podLabel: 'instance',
+          notKubeDnsCoreDnsSelector: 'job!~"kube-system/kube-dns|coredns"',
```
mplzik, line 19: nit: Is coredns running anywhere in our clusters? Depending on the configuration this might be also:
'job!~"kube-system/(kube-dns|coredns)'...

mplzik (approved):
LGTM

prometheus-ksonnet/mixins.libsonnet
Duologic: That was not changed in #344 and simply taken from the original value from kubernetes-mixin, just changing what is strictly necessary here.

PR #350, opened 2020-09-30
Memcached: reduce cpu request

c0ffeec0der:
Hello,

I would like to request to reduce cpu request from 500m to 100m.

500m is too much as starting point for small servers. 100m is also sufficient to keep alive most programs

Thank you

### DIFF
--- a/memcached/memcached.libsonnet
+++ b/memcached/memcached.libsonnet
@@ -42,7 +42,7 @@ k {
         '-c %(connection_limit)s' % self,
         '-v',
       ]) +
-      $.util.resourcesRequests('500m', $.util.bytesToK8sQuantity(self.memory_request_bytes)) +
+      $.util.resourcesRequests('100m', $.util.bytesToK8sQuantity(self.memory_request_bytes)) +
       $.util.resourcesLimits(self.cpu_limits, $.util.bytesToK8sQuantity(self.memory_limits_bytes)),
 
     memcached_exporter::


c0ffeec0der:
Hello, anyone to comment? Thanks

malcolmholmes:
There will never be a sensible default value for such a thing. Are you aware of how you can 'patch' the library and set your own values?

Add this to your main.jsonnet:

```
 memcached_container+:: .util.resourcesRequests('100m', $.util.bytesToK8sQuantity(self.memory_request_bytes))
```

And you will have set the value for yourself.

c0ffeec0der:
Hello,

Yes I am aware of it. The issue is that this library is being used by loki 

https://github.com/grafana/loki/blob/master/production/ksonnet/loki/memcached.libsonnet

and it seems that changing loki library to expose direct memcached resource change is not preffered

https://github.com/grafana/loki/pull/2695#discussion_r496865690

I am proposing to the loki contributor to expose available memcached args. However, memcached cpu request is not available in args.

My first solution is to reduce the memcached cpu request directly as in this PR, this is simpler as it will not affect other libary using it.

If this solution is not preffered, then can we expose the minimum cpu in args? So that user of loki library can adjust memcached cpu request?

What are your thoughts?

jtlisi:
@c0ffeec0der 

The recommended solution is still to modify this variable in your `environments/loki/main.jsonnet` file. You should be able to add the following snippet as @malcolmholmes said

```
memcached+::
  memcached_container+::
    $.util.resourcesRequests('100m', $.util.bytesToK8sQuantity(self.memory_request_bytes)),
```

It is unlikely we will expose the minimum CPU request as an argument unless there is a compelling/intractable reason the previous recommendation does not work. It looks like Loki imports the memcached library as a top-level object so you should be able to directly modify the container being used by the memcached library.

malcolmholmes:
As @jtlisi says, what you want is possible in jsonnet without modifying the library.

I'm gonna close this issue, feel free to reopen if this doesn't solve it for you.

c0ffeec0der:
Thank you @jtlisi 

I could do it with:

  memcached_chunks+: {
    memory_limit_mb: 1024,
    memcached_container+::
      $.util.resourcesRequests('10m', $.util.bytesToK8sQuantity(self.memory_request_bytes)) +
      $.util.resourcesLimits('20m',$.util.bytesToK8sQuantity(self.memory_limits_bytes)),
  },

PR #351, opened 2020-10-01
Fix pvc creation

nolith:
On a clean tanka project persistentVolumeClaim.new('name') is broken because parent implementation takes no parameters.

```
evaluating jsonnet: RUNTIME ERROR: function expected 0 positional argument(s), but got 1
```

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -66,7 +66,7 @@ k {
         new(name='')::
           if name != '' then
             if 'new' in super
-            then super.new(name)
+            then super.new() + super.mixin.metadata.withName(name)
             else
               {} + super.mixin.metadata.withName(name)
           else


beorn7:
Perhaps one of the Toms could have a look here?

Duologic:
There is a conflict here as we are trying to use this library with both ksonnet and k8s-jsonnet (https://github.com/jsonnet-libs/k8s-alpha, soon to be GA), for which the latter requires the name argument and the other doesn't support it. I wonder if we can support both here. (The current proposal doesn't work with k8s-jsonnet).

tomwilkie:
Sounds like we need to fix k8s-jsonnet to make it backwards compatible here?

nolith:
Thank you all for your comments.

I wasn't aware of the expected dual compatibility with ksonnet and k8s-jsonnet.

I'll stick with my fork on my project for the time being.

Duologic:
This will be resolved by #426

PR #352, opened 2020-10-02
Typo?  I suspect this works with more modern k.libsonnet.

tomwilkie:
Without this I get:

```
% tk diff environments/default
evaluating jsonnet: RUNTIME ERROR: Field does not exist: spec
	/Users/twilkie/Documents/repos/home/tanka/vendor/prometheus-ksonnet/lib/alertmanager.libsonnet:176:7-19	
	During manifestation	
```

### DIFF
--- a/prometheus-ksonnet/lib/alertmanager.libsonnet
+++ b/prometheus-ksonnet/lib/alertmanager.libsonnet
@@ -173,5 +173,5 @@
           targetPort=$._config.alertmanager_port,
         ),
       ]) +
-      service.spec.withSessionAffinity('ClientIP'),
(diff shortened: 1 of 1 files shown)


malcolmholmes (approved):
Yes, the [k8salpha library](https://github.com/jsonnet-libs/k8s-alpha) drops those `mixin` components, so technically, this is rolling back a change.

However, given we've not yet told the world to use k8salpha, we should maintain back compat, so so long as this works with the new library too, it is fine.

Is this install somewhere other than on our infra?

tomwilkie:
Yeah, this is my home k8s cluster.

PR #354, opened 2020-10-09
Introduce operator for percona-server-mongodb

bewest:
Here is an operator for percona server mongodb.  It includes a way to use the default cluster, a way to use the default cluster but with just the credentials changed, and an expressive way to create new mongodb clusters.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,1 @@
-.DS_Store
-vendor
-jsonnetfile.lock.json
-*.zip
+.*.sw[op]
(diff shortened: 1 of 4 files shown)


bewest:
@rahulmalik87, @blez, you percona folks might like tanka/jsonnet way to install the operator?

malcolmholmes:
@bewest thx for this. As we at Grafana Labs don't use percona-server-mongodb, it doesn't make so much sense for us to host the lib (we would then need to review all PRs and would lack the context to do so).

How about rather than this, you host it in your own GitHub account, and make a PR here adding an "other similar libraries" section containing a link to yours?

Duologic:
> How about rather than this, you host it in your own GitHub account, and make a PR here adding an "other similar libraries" section containing a link to yours?

+1

PR #357, opened 2020-10-10
Use mixtool to lint mixins

tomwilkie:


Makefile
```diff
@@ -12,4 +13,17 @@ lint:
 					RESULT=1; \
 				fi; \
 		done; \
+		for d in $$(find . -name '*-mixin' -a -type d -print); do \
+			if [ -e "$$d/jsonnetfile.json" ]; then \
+				echo "Installing dependancies for $$d"; \
```
sh0rez, line 18:
```suggestion
				echo "Installing dependencies for $$d"; \
```

In case you care about this ;)

jaeger-mixin/dashboards.libsonnet
```diff
@@ -30,7 +30,7 @@ local g = (import 'grafana-builder/grafana.libsonnet') + {
 {
   grafanaDashboards+: {
     'jaeger-write.json':
-      g.dashboard('Jaeger / Write')
+      (g.dashboard('Jaeger / Write') + { uid: '5f26222aa7a3fb734f0cd4072cab43b3' })
```
sh0rez, line 33: I've seen us using hashed uids in multiple places. Why? If they were human readable, they could be easy identifiers for dashboards, considering you might want to do something as `mixtool apply -t` someday

sh0rez (approved):

jaeger-mixin/dashboards.libsonnet
tomwilkie: The UIDs generated by grafana are random alpha-numeric IIRC; this was just a convenient way to get it to work, and is here to mirror what we do in prometheus-ksonnet.  We can also change this later.

malcolmholmes: what @sh0rez is saying is that we can actually use human readable UIDs and push them to Grafana. Grafana only generates a random alphanum if it isn't provided with one. Providing a UID can give more human readable URLs too which is useful.

PR #358, opened 2020-10-14
A Promsonnet library for creating Prometheus Rules

malcolmholmes:
A simple library for rendering Prometheus rules. Provides a workaround for the issue that patching Prometheus alerts/recording rules is hard in Jsonnet, as individual elements in a list are not addressable. Under the bonnet, this stores rules in a map, then renders the required list late. The entries in the map can be patched, as demonstrated in the example `patch.jsonnet`.

Comments/thoughts/approvals welcome.

### DIFF
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -15,9 +15,8 @@ local promRuleGroup = prom.v1.ruleGroup;
 {
   prometheus_metamon::
     promRuleGroup.new('prometheus_metamon')
-    + promRuleGroup.rule.new(
+    + promRuleGroup.rule.newAlert(
       'PrometheusDown', {
-        alert: 'PrometheusDown',
         expr: 'up{job="prometheus"}==0',
         'for': '5m',
         labels: {
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -3,8 +3,8 @@
 `PromSonnet` is intended as a very simple library for creating Prometheus
 alerts and rules. It is 'patching friendly', as in, it maintains the
 rules internally as a map, which allows users to easily patch alerts and
-recording rules after the fact. (With lists, this requires iteration, and
-is complex.
+recording rules after the fact. In contrast, lists require complex
+iteration logic.
 
 Take this example:
 
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -1,0 +1,65 @@
+#PromSonnet
+
+`PromSonnet` is intended as a very simple library for creating Prometheus
+alerts and rules. It is 'patching friendly', as in, it maintains the
+rules internally as a map, which allows users to easily patch alerts and
+recording rules after the fact. (With lists, this requires iteration, and
+is complex.
+
+Take this example:
+
+```
+local prom = import 'prom.libsonnet';
+local promRuleGroupSet = prom.v1.ruleGroupSet;
+local promRuleGroup = prom.v1.ruleGroup;
+{
+  prometheus_up::
+    promRuleGroup.new('prometheus_up')
+    + promRuleGroup.rule.new(
+      'PrometheusUp', {
+        alert: 'PrometheusUp',
+        expr: 'up',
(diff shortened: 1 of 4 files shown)


promsonnet/README.md
```diff
@@ -0,0 +1,65 @@
...
+local prom = import 'prom.libsonnet';
+local promRuleGroupSet = prom.v1.ruleGroupSet;
+local promRuleGroup = prom.v1.ruleGroup;
+{
+  prometheus_up::
+    promRuleGroup.new('prometheus_up')
+    + promRuleGroup.rule.new(
+      'PrometheusUp', {
+        alert: 'PrometheusUp',
+        expr: 'up',
```
beorn7, line 21: This should be something like `up{job="prometheus"}==0`, as already changed in example.jsonnet.

promsonnet/README.md
```diff
@@ -0,0 +1,65 @@
...
+
+```
+local prom = import 'prom.libsonnet';
+local promRuleGroupSet = prom.v1.ruleGroupSet;
+local promRuleGroup = prom.v1.ruleGroup;
+{
+  prometheus_up::
+    promRuleGroup.new('prometheus_up')
+    + promRuleGroup.rule.new(
+      'PrometheusUp', {
```
beorn7, line 19: I'd recommend `PrometheusDown`, as already changed in example.jsonnet. Needs to be updated in this file where applicable.

promsonnet/example.jsonnet
```diff
@@ -0,0 +1,24 @@
+local prom = import 'prom.libsonnet';
+local promRuleGroupSet = prom.v1.ruleGroupSet;
+local promRuleGroup = prom.v1.ruleGroup;
+{
+  prometheus_down::
+    promRuleGroup.new('prometheus_down')
```
beorn7, line 6: Perhaps name the group different from the rule? For example `prometheus_metamon`.

promsonnet/prom.libsonnet
```diff
@@ -0,0 +1,39 @@
...
+        name: name,
+        rules_map:: {},
+        rules_order:: [],
+        local rules_map = self.rules_map,
+        local rules_order = self.rules_order,
+        rules: [rules_map[rule] for rule in rules_order],
+      },
+
+      rule: {
+        new(name, rule):: {
```
beorn7, line 30: How about having a `newAlert` and a `newRecording` constructor instead of this `new`? They could then verify that only the allowed fields are set, and they could provide a default naming from the `name` key to not accidentally have them inconsistent.

`newAlert` would set the `alert` field from the `name` key (but give an explicitly specified field precedence)p

`newRecording` would set the `record` field from the `name` key (but give an explicitly specified field precedence).

beorn7 (changes requested):

promsonnet/prom.libsonnet
malcolmholmes: I'm not sure how we'd do validation - that's not something Jsonnet naturally lends itself to. However, a `newAlert` and `newRecording` would allow for setting the correct `name` fields as you suggest.

malcolmholmes: done as suggested.

beorn7 (approved):
Looks good from my side. Jsonnet and general config language experts might still have something to say about this.

promsonnet/README.md
```diff
@@ -0,0 +1,64 @@
+#PromSonnet
+
+`PromSonnet` is intended as a very simple library for creating Prometheus
+alerts and rules. It is 'patching friendly', as in, it maintains the
+rules internally as a map, which allows users to easily patch alerts and
+recording rules after the fact. (With lists, this requires iteration, and
+is complex.
```
Duologic, line 7:
```suggestion
recording rules after the fact, in contrast, lists require complex iteration logic.
```

Duologic (approved):
Concise, love it!

PR #360, opened 2020-10-14
Synthetic Monitoring lib

malcolmholmes:
I originally made a PR with this library against Grizzly, but realised it should be here in jsonnet-libs, so moving it over. This lib makes using Synthetic Monitoring much easier, alongside Grizzly.

### DIFF
--- a/synthetic-monitoring/README.md
+++ b/synthetic-monitoring/README.md
@@ -18,19 +18,16 @@ Here is an example usage:
 **`main.jsonnet:`**
 ```
 local sm = import 'synthetic-monitoring/sm.libsonnet';
+  
 {
   syntheticMonitoring+:: {
-    grafanaHttpCheck: sm.new('grafana', 'https://grafana.com/')
-                      + sm.withHttp()
+    grafanaHttpCheck: sm.http.new('grafana', 'https://grafana.com/')
                       + sm.withProbes('all'),  // enable all probes
-    grafanaPingCheck: sm.new('grafana', 'grafana.com')
-                      + sm.withPing()
+    grafanaPingCheck: sm.ping.new('grafana', 'grafana.com')
                       + sm.withProbes('continents'),  // one check per continent
-    grafanaDnsCheck: sm.new('grafana', 'grafana.com')
-                     + sm.withDns()
+    grafanaDnsCheck: sm.dns.new('grafana', 'grafana.com')
                      + sm.withProbes('europe'),  // just check from Europe
-    grafanaTcpCheck: sm.new('grafana', 'grafana.com:443')
-                     + sm.withTcp()
+    grafanaTcpCheck: sm.tcp.new('grafana', 'grafana.com:443')
                      + sm.withProbes('small'),  // just use a smaller, predefined set of checks
   },
 }
--- a/synthetic-monitoring/README.md
+++ b/synthetic-monitoring/README.md
@@ -1,0 +1,47 @@
+# Synthetic Monitoring Jsonnet Library
+
+This library simplifies interaction with the Grafana Synthetic Monitoring API.
+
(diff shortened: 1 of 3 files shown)


synthetic-monitoring/sm.libsonnet
```diff
@@ -0,0 +1,102 @@
...
+    offset: 0,
+    timeout: 2500,
+    enabled: true,
+    labels: [],
+    target: target,
+    job: name,
+  },
+
+  withHttp():: {
+    settings: {
```
Duologic, line 21:
```suggestion
    settings+: {
```

Duologic: Wouldn't that make more sense?

malcolmholmes: Actually, I believe it is one type per check, I should confirm that. In which case `newHttp` etc would make more sense.

Duologic: If that is the case, nesting it in `settings.http` looks a bit superfluous.

malcolmholmes: exactly.

malcolmholmes: @Duologic I've updated the API to account for this. It is also now slightly more concise. Work for you?

synthetic-monitoring/README.md
```diff
@@ -0,0 +1,44 @@
+# Synthetic Monitoring Jsonnet Library
+
+This library simplifies interaction with the Grafana Synthetic Monitoring API.
```
Duologic, line 3: This reads like the library interacts with the API. Can you rephrase? Like 'generates objects/checks for SM' or something along the lines of that.

synthetic-monitoring/README.md
```diff
@@ -0,0 +1,44 @@
...
+
+Grafana Labs' Synthetic Monitoring is a blackbox monitoring solution provided
+as a part of Grafana Cloud. It provides users with insights into how their
+applications and services are behaving from an external point of view.
+
+It can be configured via an HTTP API. This library allows interaction with
+this API using Jsonnet, making the creation of 'checks' simple.
```
Duologic, line 10: Dito, this reads like it actually interacts with the API, but this is not a client.

synthetic-monitoring/sm.libsonnet
```diff
@@ -0,0 +1,107 @@
...
+      'Frankfurt',
+      'London',
+      'Paris',
+    ],
+    australasia: [
+      'Sydney',
+    ],
+    continents: ['NewYork', 'Paris', 'Singapore', 'Sydney'],
+    all: self.americas + self.asia + self.europe + self.australasia,
+    small: ['SanFrancisco', 'NewYork', 'Singapore', 'London', 'Sydney'],
```
Duologic, line 101: 'small' sounds a bit silly, it implies there is also 'big', is this taken from an external example or did you come up with it?

synthetic-monitoring/sm.libsonnet
```diff
@@ -0,0 +1,107 @@
...
+      },
+    },
+  },
+
+  tcp: {
+    new(name, target):: _new(name, target) + {
+      settings: {
+        tcp: {
+          ipVersion: 'V4',
+          tlsConfig: {},
```
Duologic, line 33: Empty object?

Duologic (approved):
There are a few empty objects/arrays, looks like a documentation example, left a few comments, nothing blocking.

synthetic-monitoring/sm.libsonnet
malcolmholmes: that's what the API returns for a manually created check, so yes!

PR #370, opened 2020-10-21
Add patchGroup and patchAll

malcolmholmes:
Adds two functions that could be useful when patching Prometheus rules.
Means that all rules within mixins can share these, rather than needing
to write their own rule manipulators, which are invariably based upon
lists and thus complex and difficult to reason about.

### DIFF
--- a/promsonnet/example.jsonnet
+++ b/promsonnet/example.jsonnet
@@ -4,7 +4,7 @@ local promRuleGroup = prom.v1.ruleGroup;
 {
   prometheus_metamon::
     promRuleGroup.new('prometheus_metamon')
-    + promRuleGroup.rule.newAlert(
+    + promRuleGroup.newAlertRule(
       'PrometheusDown', {
         expr: 'up{job="prometheus"} == 0',
         'for': '5m',
@@ -15,7 +15,7 @@ local promRuleGroup = prom.v1.ruleGroup;
         },
       }
     )
-    + promRuleGroup.rule.newAlert(
+    + promRuleGroup.newAlertRule(
       'AlertManagerDown', {
         expr: 'left{job="alertmanager"} == 0',
         'for': '5m',
@@ -29,7 +29,7 @@ local promRuleGroup = prom.v1.ruleGroup;
 
   grafana_check::
     promRuleGroup.new('grafana_check')
-    + promRuleGroup.rule.newAlert(
+    + promRuleGroup.newAlertRule(
       'GrafanaDown', {
         expr: 'up{job="grafana"} == 0',
         'for': '5m',
--- a/promsonnet/example.jsonnet
+++ b/promsonnet/example.jsonnet
@@ -10,14 +10,38 @@ local promRuleGroup = prom.v1.ruleGroup;
         'for': '5m',
         labels: {
           namespace: 'prometheus',
-          severity: 'critical',
+        },
+        annotations: {
+        },
+      }
+    )
+    + promRuleGroup.rule.newAlert(
+      'AlertManagerDown', {
+        expr: 'left{job="alertmanager"} == 0',
+        'for': '5m',
+        labels: {
+          namespace: 'alertmanager',
+        },
+        annotations: {
+        },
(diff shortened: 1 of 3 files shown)


promsonnet/prom.libsonnet
```diff
@@ -52,5 +65,17 @@
...
+
+    patchGroup(group, patch):: {
+      groups_map+:: {
+        [group]+: {
+          patches+: [patch],
+        },
+      },
+    },
+
+    patchAll(patch):: {
```
Duologic, line 77:
```suggestion
    patchAllRules(patch):: {
```

promsonnet/prom.libsonnet
```diff
@@ -22,9 +35,9 @@
         name: name,
         rules_map:: {},
         rules_order:: [],
+        patches:: [],
         local rules_map = self.rules_map,
         local rules_order = self.rules_order,
-        rules: [rules_map[rule] for rule in rules_order],
       },
 
       rule: {
```
Duologic, line 43: `newAlert` should not be nested, just say `newAlertRule` at the same level as `pathRule`, dito for `newRecording`

wdyt?

promsonnet/prom.libsonnet
```diff
@@ -4,9 +4,22 @@
...
         local groups_order = self.groups_order,
-        groups: [groups_map[group] for group in groups_order],
+        local patch_rules_all = self.patch_rules_all,
+        groups: [
+          local group = groups_map[group_name];
+          {
+            name: group_name,
+            rules: [
+              local rule = group.rules_map[rule_name];
+              std.foldl(function(rule, patch) rule + patch, patch_rules_all + group.patches, rule)
```
Duologic, line 17: break into multiple lines increases readability imo

malcolmholmes: done

promsonnet/prom.libsonnet
malcolmholmes: done

promsonnet/prom.libsonnet
malcolmholmes: done

promsonnet/prom.libsonnet
```diff
@@ -22,24 +39,22 @@
...
-        rules: [rules_map[rule] for rule in rules_order],
       },
 
-      rule: {
-        newAlert(name, rule):: {
-          rules_map+:: {
-            [name]: rule { alert: name },
-          },
-          rules_order+:: [name],
+      newAlertRule(name, rule):: {
```
beorn7, line 47: I liked `newAlert` and `newRecording` better than `newAlertRule` and `newRecordingRule`. The latter appears as stutter to me.

beorn7: Perhaps change `newRecording` to `newRecord` to be even more consistent.

beorn7 (changes requested):

promsonnet/example.jsonnet
```diff
@@ -4,20 +4,44 @@ local promRuleGroup = prom.v1.ruleGroup;
...
           namespace: 'prometheus',
-          severity: 'critical',
         },
         annotations: {
         },
       }
+    )
+    + promRuleGroup.newAlertRule(
+      'AlertManagerDown', {
+        expr: 'left{job="alertmanager"} == 0',
```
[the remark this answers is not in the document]
malcolmholmes: Bah, stupid up, down, left, right. Was not supposed to survive into a commit.

promsonnet/prom.libsonnet
Duologic: Ha, I found the opposite true and suggested 'unnesting', just semantics, right?

beorn7: Yes, I saw that. It certainly a matter of style. I just wanted to state that I liked the previous way (`rule.NewAlert` and `rule.NewRecord`) better. I leave it to @malcolmholmes to be the tie breaker.

PR #371, opened 2020-10-23
Introduce Go runtime mixin

sdboyer:
This adds a Go runtime mixin to our little collection.

I originally made this PR against prometheus/client-golang (https://github.com/prometheus/client_golang/pull/808), and eventually, i'd like to see it move there. However, @beorn7 raised the excellent point about the visibility of that repository, and experimenting with the "just use JSON" form of a mixin isn't appropriate to do there as a result.

So, i figure we can put this here for now, until we've time to improve it a bit - even if it's just converting it to jsonnet.

There are no rules or alerts included here, as the metrics exported from the Go runtime are too general for us to meaningfully define an alert.

### DIFF
--- a/go-runtime-mixin/dashboards/go-runtime.json
+++ b/go-runtime-mixin/dashboards/go-runtime.json
@@ -1,0 +1,1153 @@
+{
+  "__inputs": [],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "7.0.4"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    }
+  ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "description": "Go runtime metrics",
+  "editable": true,
+  "gnetId": 13076,
+  "graphTooltip": 0,
+  "id": null,
+  "iteration": 1602794777869,
+  "links": [],
+  "panels": [
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "$datasource",
+      "description": "Total bytes of memory reserved by the process.",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "fill": 1,
+      "fillGradient": 0,
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 0,
+        "y": 0
+      },
+      "hiddenSeries": false,
+      "id": 16,
+      "legend": {
+        "avg": false,
+        "current": false,
+        "max": false,
+        "min": false,
+        "show": true,
+        "total": false,
+        "values": false
+      },
+      "lines": true,
+      "linewidth": 1,
+      "links": [],
+      "nullPointMode": "null",
+      "options": {
+        "dataLinks": []
+      },
+      "percentage": false,
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "go_memstats_sys_bytes{job=\"$job\", instance=~\"$instance\"}",
+          "interval": "",
+          "legendFormat": "{{instance}}: sys",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Total Reserved Memory",
+      "tooltip": {
+        "shared": true,
+        "sort": 0,
+        "value_type": "individual"
+      },
+      "type": "graph",
+      "xaxis": {
+        "buckets": null,
+        "mode": "time",
+        "name": null,
+        "show": true,
+        "values": []
+      },
+      "yaxes": [
+        {
+          "format": "decbytes",
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        },
+        {
+          "format": "short",
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        }
+      ],
+      "yaxis": {
+        "align": false,
+        "alignLevel": null
+      }
+    },
+    {
+      "aliasColors": {},
+      "bars": false,
+      "dashLength": 10,
+      "dashes": false,
+      "datasource": "$datasource",
+      "description": "",
+      "fieldConfig": {
+        "defaults": {
+          "custom": {}
+        },
+        "overrides": []
+      },
+      "fill": 1,
+      "fillGradient": 0,
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 12,
+        "y": 0
+      },
+      "hiddenSeries": false,
+      "id": 24,
+      "legend": {
+        "avg": false,
+        "current": false,
+        "max": false,
+        "min": false,
+        "show": true,
+        "total": false,
+        "values": false
+      },
+      "lines": true,
+      "linewidth": 1,
+      "links": [],
+      "nullPointMode": "null",
+      "options": {
+        "dataLinks": []
+      },
+      "percentage": false,
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "go_memstats_stack_sys_bytes{job=\"$job\", instance=~\"$instance\"}",
+          "interval": "",
+          "legendFormat": "{{instance}} stack inuse",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Stack Memory Use",
+      "tooltip": {
+        "shared": true,
+        "sort": 0,
+        "value_type": "individual"
+      },
+      "type": "graph",
+      "xaxis": {
+        "buckets": null,
+        "mode": "time",
+        "name": null,
+        "show": true,
+        "values": []
+      },
+      "yaxes": [
+        {
+          "format": "decbytes",
+          "label": null,
+          "logBase": 1,
+          "max": null,
+          "min": null,
+          "show": true
+        },
+        {
+          "format": "short",
+          "label": null,
+          "logBase": 1,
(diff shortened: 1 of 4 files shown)


go-runtime-mixin/mixin.libsonnet
```diff
@@ -0,0 +1,11 @@
+{
+  grafanaDashboards: {
+    'go-runtime.json': (import 'dashboards/go-runtime.json'),
+  },
+
+  // Helper function to ensure that we don't override other rules, by forcing
+  // the patching of the groups list, and not the overall rules object.
+  local importRules(rules) = {
```
hoenn, line 8: I don't think we need this if there are no rules.

beorn7:
Note that I left a few comments about the content of the dashboard (i.e. things in `go-runtime.json`) in https://github.com/prometheus/client_golang/pull/808 . Those are still applicable here.

sdboyer:
@beorn7 whoops! Those changes didn't survive the code move between repos. Sorry about that.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+        "skipUrlSync": false,
+        "type": "datasource"
+      },
+      {
+        "allValue": null,
+        "current": {},
+        "datasource": "$datasource",
+        "definition": "label_values(go_goroutines, job)",
+        "hide": 0,
+        "includeAll": false,
```
tomwilkie, line 1080: Can we have an "all" option please?  You'll need to update your PromQL queries to use `job=~"$job"`.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+        "type": "query",
+        "useTags": false
+      },
+      {
+        "allValue": "",
+        "current": {},
+        "datasource": "$datasource",
+        "definition": "label_values(go_goroutines{job=\"$job\"}, instance)",
+        "hide": 0,
+        "includeAll": true,
```
tomwilkie, line 1102: Can we have an "all" option please? You'll need to update your PromQL queries to use instance=~"$instance".

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+        "refresh": 1,
+        "regex": "",
+        "skipUrlSync": false,
+        "type": "datasource"
+      },
+      {
+        "allValue": null,
+        "current": {},
+        "datasource": "$datasource",
+        "definition": "label_values(go_goroutines, job)",
```
tomwilkie, line 1078: Nit: you should use the `go_info` metric here - in general I advise people to use "info" metrics to select labels, instead of "normal" metrics.  It not super relevant here, as the cardinality of `go_routines` and `go_info` will be the same - but I've seen people put super high cardinality metrics in the template variables and that makes loading the dashboard slow.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+          "min": null,
+          "show": true
+        }
+      ],
+      "yaxis": {
+        "align": false,
+        "alignLevel": null
+      }
+    },
+    {
```
tomwilkie, line 469: I get "1:118: parse error: missing unit character in duration" when loading this panel into Grafana v7.0.4.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+          "interval": "",
+          "legendFormat": "{{instance}}: mallocs/sec",
+          "refId": "A"
+        }
+      ],
+      "thresholds": [],
+      "timeFrom": null,
+      "timeRegions": [],
+      "timeShift": null,
+      "title": "Heap Object Allocation Rate",
```
tomwilkie, line 623: I get "1:118: parse error: missing unit character in duration" in Grafana v7.0.4.

I think this is due to the $__rate_interval, in which case we might want to open a feature request with Grafana to have a minimal Grafana version for a dashboard?

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+      "pointradius": 2,
+      "points": false,
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "go_memstats_sys_bytes{job=\"$job\", instance=~\"$instance\"}",
```
tomwilkie, line 92: This will give us a line per selected job / instance - and therefore when there are many instances of a given job, it will give you many lines.  One rule of thumb I have is never have more than ~5 lines per graph.

In this case I would do a `sum(...)` here, and only show the total for all jobs.  Ditto other graphs.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
+{
```
tomwilkie, line 1: I'd consider adding a table at the top of the dashboard, broken down by job & instance, showing things like the go version and allowing drill down to a specific job / instance.

Welcome to punt this to later though, just please open a ticket for it.

go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -0,0 +1,1152 @@
...
+      "renderer": "flot",
+      "seriesOverrides": [],
+      "spaceLength": 10,
+      "stack": false,
+      "steppedLine": false,
+      "targets": [
+        {
+          "expr": "go_gc_duration_seconds{job=\"$job\", instance=~\"$instance\"}",
+          "interval": "",
+          "legendFormat": "{{instance}}: {{quantile}}",
```
tomwilkie, line 904: This is pretty tricky to understand IMO; I'd have multiple queries that explicitly a different quantile each, and then make the legend more readable (eg "50th quantile" or "95th" quantile).

tomwilkie:
LGTM - bunch of really minor nits @sdboyer, feel free to punt on them for later.

Thanks!

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: From the [original PR](https://github.com/prometheus/client_golang/pull/808):

> The panels in the dashboard apply no aggregations. I waffled on this a bit, but opted for that approach on the premise that the dashboard is basic enough, and the panels show few enough individual metrics, that the dashboard itself is most likely to be helpful for folks really just getting started. (That, and, they can use the job/instance template variables to drill into a particular instance if they really need to see something that's stuck.) Very open to persuasion on changing this to an aggregated approach.

But in retrospect, i think my reasoning should have led to the opposite conclusion - apply aggregations, on the premise that the user can drill into the individual instance, at which point the aggregation becomes a no-op.

Will change.

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: That seems like the most likely culprit? Though there's already a field for the required version, see the top of the JSON:

```json
{
  "__inputs": [],
  "__requires": [
    {
      "type": "grafana",
      "id": "grafana",
      "name": "Grafana",
      "version": "7.0.4"
    },
```

i will try to reproduce.

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: :+1: makes sense (i think this actually came straight through from the marketplace dashboard)

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: i omitted that option because it didn't seem especially meaningful to look for information across _all_ Go processes. That is, it was easy to imagine ways in which aggregating across two processes that are unrelated beyond being written in Go could be confusing/misleading, but i couldn't come up with a scenario in which someone would need to look at all Go processes.

Do you have such a scenario in mind?

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: Certainly seems useful, this dash does feel kinda naked without such a thing. Here, my relative inexperience with Grafana bites me - is there an example of what you have in mind out there that i could work from?

(Aside, seems like `client_golang` should also probably emit a new gauge with `runtime.BuildInfo.Path` and `runtime.BuildInfo.Module.Version`)

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: (Oh, right. i was also trying to avoid the headache of deciding which aggregation was appropriate for the different metrics.)

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: Reproduced - yup, it's a result of the Grafana version.

Seems like, if there's an issue to be opened, it's to...actually do something with that field that already exists?

tbqh, i am circumspect about having a 'required version' field. It's ~impossible to automatically determine, which means it can produce both false negatives and false positives.

What can be automatically determined is the version of Grafana that generated the dashboard export. That can be compared against the current version, and offered as a warning if it was generated as a newer version (especially if there are e.g. parse errors).

(Points if anyone recognizes the echoes of old conversations, here :wink:)

go-runtime-mixin/dashboards/go-runtime.json
tomwilkie: If we do a drill down table I suggested then aggregating across different jobs might make sense, but without that I'd tend to agree.

go-runtime-mixin/dashboards/go-runtime.json
tomwilkie: If we do a drill down table I suggested then aggregating across different jobs might make sense, but without that I'd tend to agree.

go-runtime-mixin/dashboards/go-runtime.json
tomwilkie: As this is targets at Cloud, where we know the grafana version and its going to support this, I'm inclined to just ignore this TBH.  I'll upgrade my local one.

go-runtime-mixin/dashboards/go-runtime.json
tomwilkie: Yeah, but thats part of the fun of building these mixins!

go-runtime-mixin/dashboards/go-runtime.json
tomwilkie: Yeah, checkout the tables in the kubernetes mixin: https://admin-ops-tools1.grafana.net/grafana/d/efa86fd1d0c121a26444b636a3f509a8/kubernetes-compute-resources-cluster?orgId=1&refresh=10s

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: That is...literally true.

This is now revised, but i did avg, not sum. As the SUO here is the individual Go process, IMO the goal is to understand the behavior of each process, relative to one another; summing would be more about a fleet-level view of resource consumption.

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: OK, gonna slate that for a follow-up.

go-runtime-mixin/dashboards/go-runtime.json
sdboyer: Actually had to change this to just the 0/min and 1/max quantiles, as they're the only ones one can meaningfully aggregate.

PR #376, opened 2020-11-04
Surface Prometheus Requests + Limits for cpu, mem

halcyondude:
Surface the basics for prometheus requests and limits in _config, vs.
requiring consumers of prometheus-ksonnet to ferret out how/where to
override.  On a modest cluster (25 nodes) the defaults yield a crashing
prometheus due to resource constraints.

This does not change defaults so should not disturb existing deploys.


malcolmholmes (approved):
This is simple, and useful, thanks!

PR #379, opened 2020-11-04
feat(helm-util): setNamespace function

Duologic:
ref: https://github.com/grafana/tanka/issues/418

### DIFF
--- a/helm-util/helm.libsonnet
+++ b/helm-util/helm.libsonnet
@@ -97,4 +97,21 @@ local d = import 'github.com/sh0rez/docsonnet/doc-util/main.libsonnet';
         },
       }
     ),
+
+  '#setNamespace':: d.fn(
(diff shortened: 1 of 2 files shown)


helm-util/helm.libsonnet
```diff
@@ -97,4 +97,21 @@ local d = import 'github.com/sh0rez/docsonnet/doc-util/main.libsonnet';
         },
       }
     ),
+
+  '#setNamespace':: d.fn(
```
dfang, line 101: it doesn't check if a resource is cluster wide (non namespaced). though adding a namespace to non-namespaced resource (eg. ClusterRole) can be created as well.

Duologic:
I think this is a nice example of how to use the `patchKubernetesObjects` but it doesn't need to be included in this library. Closing it for now.

PR #380, opened 2020-11-04
Provide MinIO mixin

mdisibio:
This PR provides a mixin for MinIO.  There are a couple of things to note:

* The starting point was the existing dashboard: https://grafana.com/grafana/dashboards/11568
* In local testing MinIO does not compute latency for write/delete requests, so there is only a latency graph for reads.
* In local testing the Disks Total/Offline stats are always zero, but still including because thinking this is just an artifact of local testing and the metrics do normally function.
* There are a couple alarms that seem valuable to include since they are invariant:  when free storage space is low, or disks offline is non-zero.  However, not sure of the best way to include those alarms here, so looking for some feedback.  Things to account for:
** Is there support in the jsonnet-libs for creating the alarms, or a good example?
** My understanding is that alarms require a graph panel, so the current gauge panel would have to change, or another graph panel created with somewhat duplicate information.  This was the approach I was originally going with.

### DIFF
--- a/minio-mixin/minio_v1.libsonnet
+++ b/minio-mixin/minio_v1.libsonnet
@@ -1,0 +1,98 @@
+local g = import 'grafana-builder/grafana.libsonnet';
+
+local panel_settings_bytes = {
+    yaxes: g.yaxes('bytes')
+};
+
+local panel_settings_qps = {
+    stack: true,
+};
+
+// Add common filters
+local f(s) = s % 'instance=~"$instance", job=~"$job"';
+
+g.dashboard('MinIO distributed cluster metrics', std.md5('minio_v1'))
+.addTemplate('job', 'minio_version_info', 'job')
+.addMultiTemplate('instance', 'minio_version_info{job="$job"}', 'instance')
+.addMultiTemplate('disk', 'disk_storage_available{job="$job"}', 'disk')
+.addRow(
+    g.row('Overview')
+    .addPanel(
+        g.panel('Storage Used') + 
+        g.statPanel('sum(disk_storage_used{disk=~"$disk", job=~"$job"}) by (disk) 
+                   / sum(disk_storage_total{disk=~"$disk", job=~"$job"}) by (disk)') + 
+        {
+            type: 'gauge',
+            // The default 'percentunit' format of statPanel() doesn't seem to have an effect
+            fieldConfig: {
+                defaults: {
+                    unit: 'percentunit'
+                },
+            },
+        }
+    )
+    .addPanel(
+        g.panel('Disks Total') +
+        g.statPanel(f('sum(minio_disks_total{%s})'), 'none')
+    )
+    .addPanel(
+        g.panel('Disks Offline') +
+        g.statPanel(f('sum(minio_disks_offline{%s})'), 'none')
+    )
+    .addPanel(
+        g.panel('Errors') +
+        g.queryPanel(f('s3_errors_total{%s}'), '')
+    )
+)
+.addRow(
+    g.row('Storage')
+    .addPanel(
+        g.panel('Storage Used') +
+        g.queryPanel(f('disk_storage_used{disk=~"$disk",%s}'), '') +
+        panel_settings_bytes,
+    )
+    .addPanel(
+        g.panel('Storage Available') +
+        g.queryPanel(f('disk_storage_available{disk=~"$disk",%s}'), '') +
+        panel_settings_bytes,
+    )
+    .addPanel(
+        g.panel('Storage Total') +
+        g.queryPanel(f('disk_storage_total{disk=~"$disk",%s}'), '') +
+        panel_settings_bytes,
+    )
+)
+.addRow(
+    g.row('Requests')
+    .addPanel(
+        g.panel('Read rps') +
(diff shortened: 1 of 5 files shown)


sdboyer:
Awesome progress, thank you! re: your alerts questions:

> Is there support in the jsonnet-libs for creating the alarms, or a good example?
> ** My understanding is that alarms require a graph panel, so the current gauge panel would have to change, or another graph panel created with somewhat duplicate information. This was the approach I was originally going with.

The standard approach in mixins is actually to define Prometheus alerts, not Grafana alerts. It's the latter that have always been a property of dashboards (though that may be changing in Grafana v8). The [etcd mixin](https://monitoring.mixins.dev/etcd/#alerts) has a good set.

minio-mixin/minio_v1.libsonnet
```diff
@@ -0,0 +1,97 @@
...
+  )
+)
+.addRow(
+  g.row('Performance')
+  .addPanel(
+    g.panel('Read latency') +
+    g.latencyPanel('s3_ttfb_seconds', f('{api=~"get.*|list.*|head.*",%s}'))
+  )
+  .addPanel(
+    g.panel('Internode traffic') +
```
sdboyer, line 88: Let's at least include a note on this indicating that it's expected to be zero if running in a single-node configuration.

minio-mixin/minio_v1.libsonnet
```diff
@@ -0,0 +1,97 @@
...
+    g.queryPanel([
+      f('sum(rate(internode_rx_bytes_total{%s}[$__rate_interval]))'),
```
sdboyer, line 90: Is sum an informative aggregation, here?

sdboyer:
> In local testing the Disks Total/Offline stats are always zero, but still including because thinking this is just an artifact of local testing and the metrics do normally function.

Could you say a bit more about the local testing problem, and why you think it won't transfer? If these turn out not to work for some unanticipated reason, i'd say that would fail to meet our 'not obviously wrong' standard.

mdisibio:
> > In local testing the Disks Total/Offline stats are always zero, but still including because thinking this is just an artifact of local testing and the metrics do normally function.
> 
> Could you say a bit more about the local testing problem, and why you think it won't transfer? If these turn out not to work for some unanticipated reason, i'd say that would fail to meet our 'not obviously wrong' standard.

Yes and good news, after doing some more research I was able to verify that the metrics do work correctly, but there are some cases where it won't.  Hard to say what percent of use cases fall into these scenarios, but it seems like most production setups will be fine.  Some information:
* There was a bug fixed where the disks total/offline stats would be zero in some cases: https://github.com/minio/minio/pull/9610  Users running the affected versions may be missing these metrics.
* Those metrics also only seem to be populated when running in a multi-node setup.  For my previous testing I had a single container only.   After switching to the multi-node docker-compose example here: https://docs.min.io/docs/deploy-minio-on-docker-compose, the metrics are populated as expected. This example runs 4 nodes with 2 disks.

minio-mixin/minio_v1.libsonnet
mdisibio: Sure thing.  Are you thinking a comment in the code here, or something in the dashboard like a panel description?

minio-mixin/minio_v1.libsonnet
mdisibio: Was mostly following the design of the existing dashboard for this, but agree, it seems more informative to show it unaggregated (per-node).

minio-mixin/minio_v1.libsonnet
sdboyer: Sorry, ambiguous - yeah, i was thinking something terse in the panel description.

sdboyer:
> Those metrics also only seem to be populated when running in a multi-node setup. 

Hmm...also a tad concerning. Having to differentiate between operating modes by cluster size isn't great. Anything you can think of to suggest to the user that the absence of the disk space metrics is normal when in single-node mode?

mdisibio:
> > Those metrics also only seem to be populated when running in a multi-node setup.
> 
> Hmm...also a tad concerning. Having to differentiate between operating modes by cluster size isn't great. Anything you can think of to suggest to the user that the absence of the disk space metrics is normal when in single-node mode?

To clarify, it is the disk _count_ metrics that are in question.  Disk space metrics are ok. Have a few more findings: 

It is actually related to the backend type and not the number of nodes.  The backend type can be either local filesystem or Erasure.  "Erasure" is the distributed backend using erasure codes and requires at least 4 disks.  When the backend type is the local filesystem (1-3 disks) then metrics are not populated (always zero).  When 4+ drives are configured as an Erasure backend then the metrics work as expected.
https://docs.min.io/docs/minio-erasure-code-quickstart-guide.html 
https://docs.min.io/docs/distributed-minio-quickstart-guide.html

The disks_total metric will then always be 0 or 4+.    There are a few things we could do:
(1) Change the title of the panels to be more explicit about the disk types, for example "Erasure Disk Count"
(2) Change the query to `sum(minio_disks_total{%s}) > 0` which will cause the panel to display "N/A" unless a non-zero value is present.

Thoughts?  Maybe making both changes sounds best.

sdboyer:
> To clarify, it is the disk count metrics that are in question.

Oh! I did indeed misunderstand. Less concerning.

> Maybe making both changes sounds best.

:+1:

sdboyer (approved):
i'd say we've reached threshold! 🎉 🎉

PR #384, opened 2020-11-09
New nginx directory app

malcolmholmes:
prometheus-ksonnet as a simple nginx install to provide a directory of services. This is neat, but isn't particularly pretty. This makes it more visually appealing, and also adds the capability for sub-pages, allowing one application to have a number of links, making longer directory listings more useful.

### DIFF
--- a/nginx-directory/config.libsonnet
+++ b/nginx-directory/config.libsonnet
@@ -1,0 +1,20 @@
+{
+  apps_map: {},
+  app_order:: [],
+
+  _config+:: {
+    cluster_name: error 'cluster_name required',
+
+    // oauth2-proxy
+    oauth_enabled: false,
+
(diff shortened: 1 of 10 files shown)


nginx-directory/config.libsonnet
```diff
@@ -0,0 +1,20 @@
+{
+  apps_map: {},
+  app_order:: [],
+
+  _config+:: {
+    cluster_name: error 'cluster_name required',
+
+    // oauth2-proxy
+    oauth_enabled: false,
```
sh0rez, line 9: Could this be a mixin instead?

malcolmholmes:
This PR shows what is possible, but, to actually merge it, it needs to be broken down in a sequence of smaller changes. I'll cover this work in separate PRs.

PR #387, opened 2020-11-10
Add config for Prometheus HA-aware alerts

beorn7:
This requires https://github.com/prometheus/prometheus/pull/8170 .

Signed-off-by: beorn7 <beorn@grafana.com>

### DIFF
--- a/prometheus-ksonnet/mixins.libsonnet
+++ b/prometheus-ksonnet/mixins.libsonnet
@@ -26,6 +26,8 @@
 
         _config+:: {
           prometheusSelector: 'job="default/prometheus"',
+          prometheusHAGroupLabels: 'job,cluster,namespace',
+          prometheusHAGroupName: '{{$labels.job}} in {{$labels.cluster}}',
         },
       },
 


malcolmholmes:
This approach should work. However, if statements in jsonnet tend to create ugly code. A better approach would be to have a hidden element in the prometheus library, then you add the hidden element to your visible one if the feature is required. Effectively you are mixing into your mixin. Happy to discuss/help.

beorn7:
I think I understand what you are up to. Working on it…

beorn7:
:tada: @malcolmholmes please have another look.

beorn7:
I mean, the change is over in https://github.com/prometheus/prometheus/pull/8170.

Duologic:
I think @malcolmholmes propose something like this, wdyt? `withHARules` would override the rules 'without HA'.
```
    local prometheus_mixin = import 'prometheus-mixin/mixin.libsonnet',
    prometheus:
      prometheus_mixin
      + prometheus_mixin.withHARules('job,cluster,namespace', '{{$labels.job}} in {{$labels.cluster}}')
      + {
        grafanaDashboardFolder: 'Prometheus',

        _config+:: {
          prometheusSelector: 'job="default/prometheus"',
        },
      },
```

beorn7:
Yes, understood.

I like that approach, but given that we ask the user to configure other stuff with variables, I prefer in this particular case to just stick with that pattern for now, i.e. let the user fill the two config variables, and then the rest will automatically happen.

If we picked the above approach, we should keep it consistent and set the other config variables in the same method-based way. But I would prefer to do that in a separate PR.

Duologic (approved):

PR #391, opened 2020-11-22
Pull directory out of prom-ksonnet

malcolmholmes:
In #384 I separated out the Nginx portion into a separate library, but also added a heap of other (cool) stuff. That other stuff, though cool, made the PR hard to review.

Here's a simpler one, a step on the way there. This does just two things:
* splits out the Nginx code into a separate `jsonnet-libs/nginx-directory` dir
* splits the configmaps in two, one for the nginx config, and another for the html. This means we can have multiple files (css, whatever) in the standard `/var/www/html` directory, rather than having config and HTML in the same place.

Other than splitting the config map in two, this should otherwise be a no-op.


Duologic:
It would be nice if we'd be able to observe an actual no-op on our internal codebase, the configmap split is nice but doesn't guarantee that.

malcolmholmes:
> It would be nice if we'd be able to observe an actual no-op on our internal codebase, the configmap split is nice but doesn't guarantee that.

We can undo that element, and add it in a separate step, easily enough @Duologic .

tomwilkie:
(unsub as Dulogic is on it)

Duologic:
I'd like to see that in 2 PRs, will be easier for other users to update prometheus-ksonnet in 2 steps too.

nginx-directory/deployment.libsonnet
```diff
@@ -0,0 +1,29 @@
...
+
+  nginx_deployment:
+    deployment.new('nginx', 1, [$.nginx_container]) +
+    $.util.configMapVolumeMount($.nginx_config_map, '/etc/nginx') +
+    $.util.podPriority('critical'),
+
+  nginx_service:
+    $.util.serviceFor($.nginx_deployment),
+
+  local oauth2_proxy = import 'oauth2_proxy/oauth2-proxy.libsonnet',
```
Duologic, line 19: I don't see this file or a (dep in a) jsonnetfile.json for nginx-directory.

nginx-directory/deployment.libsonnet
```diff
@@ -0,0 +1,29 @@
+{
+  local container = $.core.v1.container,
```
Duologic, line 2: There is no import of `k.libsonnet` or `kausal.libsonnet`, might need that here.

prometheus-ksonnet/lib/config.libsonnet
```diff
@@ -87,13 +87,5 @@
...
     // Node exporter options.
     node_exporter_mount_root: true,
-
-    // oauth2-proxy
-    oauth_enabled: false,
-
-    // Nginx proxy_read_timeout (in seconds) 60s is the nginx default
-    nginx_proxy_read_timeout: '60',
-    // Nginx proxy_send_timeout (in seconds) 60s is the nginx default
-    nginx_proxy_send_timeout: '60',
```
Duologic, line 97: nit: It is not necessary to remove this block.

nginx-directory/deployment.libsonnet
```diff
@@ -0,0 +1,29 @@
...
+  nginx_container::
+    container.new('nginx', $._images.nginx) +
+    container.withPorts($.core.v1.containerPort.new('http', 80)) +
+    $.util.resourcesRequests('50m', '100Mi'),
+
+  local deployment = $.apps.v1.deployment,
+
+  nginx_deployment:
+    deployment.new('nginx', 1, [$.nginx_container]) +
+    $.util.configMapVolumeMount($.nginx_config_map, '/etc/nginx') +
```
Duologic, line 13: Where is nginx_config_map in nginx-directory?

Duologic (changes requested):

nginx-directory/deployment.libsonnet
malcolmholmes: in `configfile.libsonnet`

nginx-directory/deployment.libsonnet
malcolmholmes: good point, will fix.

prometheus-ksonnet/lib/config.libsonnet
malcolmholmes: fair enough, reverting this chagne.

nginx-directory/deployment.libsonnet
malcolmholmes: I've fixed that.

Duologic (approved):

PR #392, opened 2020-11-24
add ServiceAccount annotations & pull secrets

halcyondude:
Optionally add imagePullSecrets and ServiceAccount Annotations to helper
functions `rbac()` and `namespacedRBAC` - both essential for production
deployments.

Image Pull Secrets are used to allow for pulling images from a private
repository or from a public repository requiring authenticated pulls.

https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry

Service Account (SA) annotations are typically used by public cloud
providers such as AWS to annotate SA's with a role to be assumed.

In the case of Cortex, Loki, or Tempo this is typically a role allowing
for access to S3.

For example:

https://docs.aws.amazon.com/eks/latest/userguide/specify-service-account-role.html

These are (in the EKS use case):

```
eks.amazonaws.com/role-arn: arn:aws:iam::<AWS_ACCOUNT_ID>:role/<IAM_ROLE_NAME>
```

An example of a Service Account generated with this patch, passing both
inputs follows.  Existing consumers of the 2 modified functions are not
impacted.

```
apiVersion: v1
imagePullSecrets:
- name: dockerhub
kind: ServiceAccount
metadata:
  annotations:
    eks.amazonaws.com/role-arn: arn:aws:iam::123456789012:role/loki-ingester=loki1-east1-us-prod
  name: ingester
  namespace: loki
```

Signed-off-by: Matt Young <halcyondude@gmail.com>

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -252,7 +252,7 @@ k {
 
     // rbac creates a service account, role and role binding with the given
     // name and rules.
-    rbac(name, rules)::
+    rbac(name, rules, pullSecrets={}, annotations={})::
       if $._config.enable_rbac
       then {
         local clusterRole = $.rbac.v1beta1.clusterRole,
@@ -261,7 +261,14 @@ k {
         local serviceAccount = $.core.v1.serviceAccount,
 
         service_account:
-          serviceAccount.new(name),
+          serviceAccount.new(name)
+          + (
+            if (pullSecrets == {}) then {}
+            else serviceAccount.withImagePullSecrets(pullSecrets)
+          ) + (
+            if (annotations == {}) then {}
+            else serviceAccount.metadata.withAnnotationsMixin(annotations)
+          ),
 
         cluster_role:
           clusterRole.new() +
@@ -283,7 +290,7 @@ k {
       }
       else {},
 
-    namespacedRBAC(name, rules)::
+    namespacedRBAC(name, rules, pullSecrets={}, annotations={})::
       if $._config.enable_rbac
       then {
         local role = $.rbac.v1beta1.role,
@@ -293,7 +300,14 @@ k {
 
         service_account:
           serviceAccount.new(name) +
-          serviceAccount.mixin.metadata.withNamespace($._config.namespace),
+          serviceAccount.mixin.metadata.withNamespace($._config.namespace)
+          + (
+            if (pullSecrets == {}) then {}
+            else serviceAccount.withImagePullSecrets(pullSecrets)
+          ) + (
+            if (annotations == {}) then {}
+            else serviceAccount.metadata.withAnnotationsMixin(annotations)
+          ),
 
         role:
           role.new() +


halcyondude:
I'm a little confused why the linter is failing on this...

```
--- ./ksonnet-util/kausal.libsonnet
+++ -
@@ -264,8 +264,8 @@
           serviceAccount.new(name) +
           if (pullSecrets == {}) then {}
           else serviceAccount.withImagePullSecrets(pullSecrets) +
-          if (annotations == {}) then {}
-          else serviceAccount.metadata.withAnnotationsMixin(annotations),
+               if (annotations == {}) then {}
+               else serviceAccount.metadata.withAnnotationsMixin(annotations),
 
         cluster_role:
           clusterRole.new() +
@@ -300,8 +300,8 @@
           serviceAccount.mixin.metadata.withNamespace($._config.namespace) +
           if (pullSecrets == {}) then {}
           else serviceAccount.withImagePullSecrets(pullSecrets) +
-          if (annotations == {}) then {}
-          else serviceAccount.metadata.withAnnotationsMixin(annotations),
+               if (annotations == {}) then {}
+               else serviceAccount.metadata.withAnnotationsMixin(annotations),
```

halcyondude:
I rather think the one-liners are most pleasant to look at (see red below) - heartily seeking advice from maintainers as to the right stylistic (and linter happy) formatting for this patch please :)

image

nabadger:
@halcyondude regarding the linter you should be able to run `tk fmt` on this which just applies defaulting formating (makes life easy)

Ends up fixing the indentation to:

```
Regarding the linter, I think it's indentation related.

You should be able to just run tk fmt on it to fix which results in 

diff --git a/ksonnet-util/kausal.libsonnet b/ksonnet-util/kausal.libsonnet
index df5e337..5a6fcd1 100755
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -264,8 +264,8 @@ k {
           serviceAccount.new(name) +
           if (pullSecrets == {}) then {}
           else serviceAccount.withImagePullSecrets(pullSecrets) +
-          if (annotations == {}) then {}
-          else serviceAccount.metadata.withAnnotationsMixin(annotations),
+               if (annotations == {}) then {}
+               else serviceAccount.metadata.withAnnotationsMixin(annotations),
 
         cluster_role:
           clusterRole.new() +
@@ -300,8 +300,8 @@ k {
           serviceAccount.mixin.metadata.withNamespace($._config.namespace) +
           if (pullSecrets == {}) then {}
           else serviceAccount.withImagePullSecrets(pullSecrets) +
-          if (annotations == {}) then {}
-          else serviceAccount.metadata.withAnnotationsMixin(annotations),
+               if (annotations == {}) then {}
+               else serviceAccount.metadata.withAnnotationsMixin(annotations),
 
         role:
           role.new() +

```

halcyondude:
@nabadger thanks :)

ksonnet-util/kausal.libsonnet
```diff
@@ -252,7 +252,7 @@ k {
 
     // rbac creates a service account, role and role binding with the given
     // name and rules.
-    rbac(name, rules)::
+    rbac(name, rules, annotations={}, pullSecrets={})::
```
Duologic, line 255: I'd put the arguments in the same order as how they are referenced below.

ksonnet-util/kausal.libsonnet
```diff
@@ -261,7 +261,11 @@ k {
         local serviceAccount = $.core.v1.serviceAccount,
 
         service_account:
-          serviceAccount.new(name),
+          serviceAccount.new(name) +
+          if (pullSecrets == {}) then {}
+          else serviceAccount.withImagePullSecrets(pullSecrets) +
+               if (annotations == {}) then {}
+               else serviceAccount.metadata.withAnnotationsMixin(annotations),
```
Duologic, line 268: You can wrap the if/else in brackets to have preserve the indentation.
```suggestion
          serviceAccount.new(name)
          + (
            if (pullSecrets == {}) then {}
            else serviceAccount.withImagePullSecrets(pullSecrets)
          ) + (
            if (annotations == {}) then {}
            else serviceAccount.metadata.withAnnotationsMixin(annotations)
          ),
```

Duologic (commented):
I left some purely suggestive comments.

On the proposed features itself, the if/else feels a bit like templating in Helm. I think it might be more jsonnet-y as a mixin functions:

```jsonnet
{
  rbacUtils:: {
    addPullSecrets(pullSecrets):: {
      service_account+: serviceAccount.withImagePullSecrets(pullSecrets),
    },
    addAnnotations(annotations):: {
      service_account+: serviceAccount.metadata.withAnnotationsMixin(annotations),
    },
  },
  account:
    $.rbac('name', [/*rules*/])
    + $.rbacUtils.addPullSecerts([/*pullSecrets*/]),
}

```

WDYT?

ksonnet-util/kausal.libsonnet
halcyondude: Thanks for the feedback!  Updated per your suggestion.

ksonnet-util/kausal.libsonnet
halcyondude: Thanks for the feedback!  Updated per your suggestion.

halcyondude:
> I left some purely suggestive comments.
> 
> On the proposed features itself, the if/else feels a bit like templating in Helm. I think it might be more jsonnet-y as mixin functions:
> 
> ```jsonnet
> {
>   rbacUtils:: {
>     addPullSecrets(pullSecrets):: {
>       service_account+: serviceAccount.withImagePullSecrets(pullSecrets),
>     },
>     addAnnotations(annotations):: {
>       service_account+: serviceAccount.metadata.withAnnotationsMixin(annotations),
>     },
>   },
>   account:
>     $.rbac('name', [/*rules*/])
>     + $.rbacUtils.addPullSecerts([/*pullSecrets*/]),
> }
> ```
> 
> WDYT?

caveat: Still learning...

I started with this approach, and ran into not seeing a clean way to handle the null case.  I was hoping to update these shared functions such that:

* existing users/consumers of the rbac functions don't change (via optional parameters)
* allowing projects that consume these (loki, cortex, tempo) to surface via `_config::`, by having variables that default to `{}` (**_but that exist_**).

The advantage of this approach is that in the (many) places these functions are called, they could pass these `_config::` params directly, without having to have conditional logic at each call site, and the actual conditional if/else being done once in the library function.

I also attempted to layer in via mixins atop the consuming projects (e.g. loki), with a series of patches on each of the service accounts they create - which is presently only a handful (The rest needed to be created directly)

What I found was that this worked for the accounts I've created via jsonnet in my library (for now), because the projects do not yet create service accounts for many of the services needed.  I do intend to propose PR's for these, but this is a much larger impact, and suffers from some of the same "how to optionally create SA's" larger question I didn't want to attempt to solve in this PR.  In scoping out that, it became apparent that:

* Having a mechanism to cleanly (and optionally) provide the capability in a backwards-compatible way necessitates updating the library to use optional params
* I didn't see another way (apart from the if-else construct, which admittedly is borderline "helmy" :) ) to implement this.
* patches to the service accounts (post creation) became quite brittle and difficult to consume...as what to (specifically) patch varies, and ends up introducing hard coupling on both the (hard-coded) service names, as well as the overall structure/topology of the consuming projects.

_Aside: if `{}` is passed thru to  `.withImagePullSecrets()` or `.withAnnotationsMixin()` malformed k8s yaml is produced_

halcyondude:
(rebased to master)

halcyondude:
Here's an example of what this PR enables.  This is how we're running Loki today in production.

I've just finished moved our EKS config for Loki --> open, with Cortex and Tempo to follow 

https://github.com/tanka-tools/observe

I've re-factored to a reusable module --> https://github.com/tanka-tools/observe/tree/main/loki-aws, which uses this PR (via shadowing) so this is possible.  I plan to PR this --> Loki as described above.

Using config like this:

```jsonnet
{
  _config+:: {
    namespace: error 'namespace is a required input',
    cluster: error 'cluster is a required input',

    // TODO: default these to {} to generalize
    // IAM Roles annotated --> k8s Service Accounts, allowing access to S3
    arn_ingester_sa:      error ' arn_ingester_sa is required',
    arn_querier_sa:       error ' arn_querier_sa is required',
    arn_table_manager_sa: error ' arn_table_manager_sa is required',

    // named k8s secret --> 'dockerhub-credentials'  MUST be created in whatever
    // namespace(s) are used. imagePullSecrets are added to each Service Account
    image_pull_secrets: { 'name': 'dockerhub-credentials' },
...
```

https://github.com/tanka-tools/observe/blob/main/loki-aws/tt-loki-aws.libsonnet

```jsonnet
  //
  // StatsfulSet - Ingester, compactor, querier
  //

  ingester_rbac:
    $.util.namespacedRBAC('ingester', emptyPolicy,
      annotations={ 'eks.amazonaws.com/role-arn': $._config.arn_ingester_sa },
      pullSecrets=$._config.image_pull_secrets),

  ingester_statefulset+:
    statefulSet.spec.template.spec.withServiceAccountName('ingester') +
    statefulSet.mixin.spec.withReplicas($._config.ingester_replicas),

  compactor_rbac:
    $.util.namespacedRBAC('compactor', emptyPolicy,
    annotations={ 'eks.amazonaws.com/role-arn': $._config.arn_compactor_sa },
    pullSecrets=$._config.image_pull_secrets),

  compactor_statefulset+:
    statefulSet.spec.template.spec.withServiceAccountName('compactor'),

  querier_rbac:
    $.util.namespacedRBAC('querier', emptyPolicy,
    annotations={ 'eks.amazonaws.com/role-arn': $._config.arn_querier_sa },
    pullSecrets=$._config.image_pull_secrets),

  querier_statefulset+:
    statefulSet.spec.template.spec.withServiceAccountName('querier') +
    statefulSet.mixin.spec.withReplicas($._config.querier_replicas),
```

Duologic:
Hey @halcyondude 

Sorry for the delay here. I had another look at your use case. It looks like your main problem is handling the 'null' case, right? If so, I think I might have found the culprit.

`withImagePullSecrets()` actually renders an array `[]`, so if we pass an empty array, it would still render fine. If we pass in an object, an empty object would render a `[{}]` instead. The trick is to prune the list: `std.pune([$._config.image_pull_secrets])`.

From the example above, this should render a valid kubernetes resource, even if one of them is null:

```jsonnet
  local annotations = { 'eks.amazonaws.com/role-arn': $._config.arn_ingester_sa },
  ingester_rbac:
    $.util.namespacedRBAC('ingester', emptyPolicy)
    {
      service_account+:
        $.core.v1.serviceAccount.mixin.metadata.withAnnotationsMixin(annotations)
        + $.core.v1.serviceAccount.withImagePullSecretsMixin(std.prune([$._config.image_pull_secrets])),
    },
```

I hope this solves your problem and we don't have to update ksonnet-util for this use case.

Duologic:
Considering above poses a solution to the problem at hand, I assume this is solved. Feel free to reach out again.

PR #393, opened 2020-11-24
Allow override of target address by using fqdn

dafydd-t:
this backwards-compatible change allows us to use a fqdn instead of deployment name / namespace for choosing the mysql target. This makes it cleaner to use this library when the target mysql instance is another container in the same pod.

You can see an example of how we intend on using this change [here](https://github.com/grafana/deployment_tools/pull/6136).

### DIFF
--- a/mysql-exporter/mysql-exporter.libsonnet
+++ b/mysql-exporter/mysql-exporter.libsonnet
@@ -10,13 +10,20 @@ local deployment = k.apps.v1.deployment;
     mysql_password: error 'must specify mysql password',
     deployment_name: error 'must specify deployment name',
     namespace: error 'must specify namespace',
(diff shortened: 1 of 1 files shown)


mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -10,13 +10,20 @@ local deployment = k.apps.v1.deployment;
     mysql_password: error 'must specify mysql password',
     deployment_name: error 'must specify deployment name',
     namespace: error 'must specify namespace',
+    mysql_fqdn: '',
```
bergquist, line 13: I suggest you make this a hidden variable like `image` unless it's a required param.

bergquist (approved):
LGTM

PR #394, opened 2020-11-24
Enable AM on new installs

malcolmholmes:
We've done a lot of work on the Alert Manager setup, to support complex multi-cluster environments, and in the process have left behind single cluster setups. At present, deploying prometheus-ksonnet will exclude the alertmanager, because it doesn't have a named cluster listed - this is kinda excessive in terms of unnecessary configuration.

This PR simply enables a single AM instance should the list of clusters be empty (the default case).


Duologic (approved):
Sounds reasonable.

PR #398, opened 2020-12-07
Fix moved dependency for alertmanager mixin

xvzf:
Reference to upstream PR here: https://github.com/prometheus/alertmanager/pull/1629

Signed-off-by: Matthias Riegler <matthias.riegler@taotesting.com>


Duologic (approved):
https://github.com/prometheus/alertmanager/pull/1629

PR #400, opened 2020-12-13
feat: standalone prometheus library

Duologic:
I'd like to have a standalone Prometheus setup for use in a "sharded" prometheus installation. This is PR pulls out the
prometheus install from prometheus-ksonnet to a bare configuration (without mixins/scrape configs/alerting). This should
be a good base to start of new prometheus installs.

I've also modified prometheus-ksonnet to use this 'bare' prometheus. I've tested this on a 'complex' internal setup and
it renders perfectly, only the 'new' `mixins.base` rules move around as expected.


prometheus/prometheus.libsonnet
```diff
@@ -0,0 +1,141 @@
+local kausal = import 'ksonnet-util/kausal.libsonnet';
+
+
```
malcolmholmes, line 3: two blank lines??

malcolmholmes (approved):
This is fantastic. Soooo much clearer than what we have. But, do fix the whitespace! :-)

PR #401, opened 2020-12-14
Grafana builder: use $__rate_interval over $__interval

bergquist:

### DIFF
--- a/grafana-builder/grafana.libsonnet
+++ b/grafana-builder/grafana.libsonnet
@@ -382,7 +382,7 @@
         expr:
           |||
             sum by (status) (
-              label_replace(label_replace(rate(%s[$__interval]),
+              label_replace(label_replace(rate(%s[$__rate_interval]),
               "status", "${1}xx", "%s", "([0-9]).."),
               "status", "${1}", "%s", "([a-z]+)"))
           ||| % [selector, statusLabelName, statusLabelName],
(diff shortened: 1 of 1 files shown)


grafana-builder/grafana.libsonnet
```diff
@@ -399,23 +399,23 @@
     nullPointMode: 'null as zero',
     targets: [
       {
-        expr: 'histogram_quantile(0.99, sum(rate(%s_bucket%s[$__interval])) by (le)) * %s' % [metricName, selector, multiplier],
+        expr: 'histogram_quantile(0.99, sum(rate(%s_bucket%s[$__rate_interval])) by (le)) * %s' % [metricName, selector, multiplier],
```
bergquist, line 402: I've never used `histogram_quantile` so I dont know if this makes sense.. But I dont see why It wouldn't :man_shrugging:

tomwilkie:
When was rate_interval introduced?  This will break people on old Grafana IIUC.  Just want to make sure old in this context doesn't mean "last week".

bergquist:
It was introduced in Grafana 7.2.0 released (2020-09-23)

ref https://github.com/grafana/grafana/pull/26937

tomwilkie:
Awesome, thanks Carl.  I'll leave it to you to judge if 3 months is enough - this library is used by the k8s mixin, which is included in the kube-prometheus, which embeds grafana/grafana:7.3.5 in head (see https://github.com/prometheus-operator/kube-prometheus/blob/master/manifests/grafana-deployment.yaml).  So I'd say LGTM

bergquist:
3 months in mixin lands is how many human years?

PR #406, opened 2020-12-16
upgrade cert-manager to 1.1.0

captncraig:
All I did was:

- delete charts/cert-manager
- update chartfile
- `tk tool charts vendor`


Duologic:
As discussed directly, we need to consult the cert-manager changelog before upgrading, it might not be as smooth as this looks.

captncraig:
That leaves an interesting problem for this library. If we merge these changes as in, then do manual steps to upgrade, we could cause problems for any (I doubt it tbh) other users of this lib. 

I am testing with this commit directly in one dev environment for now.

Duologic:
The lack of a frictionless upgrade path is not caused by this Jsonnet library, rather the upstream Helm Chart.

Duologic (approved):

captncraig:
Upgrade experience from dev-us-central1:

All deployments come up just fine. No hanging pods or anything like that. Some components failed because of missing CRDs, so I turned off custom_crds, and the chart made a thousand updates, then everything worked.

captncraig:
But there isn't extensive usage there, so we will need to watch carefully when we roll to production clusters.

PR #408, opened 2020-12-23
feat(prometheus): move HA mixin

Duologic:
This should move too.

I'm sneaking in a tiny compat layer for ksonnet-lib, it should just provide compat inside this library, not on all
libraries. And also a fix for using upstream images.


prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -1,104 +1,8 @@
-local k = import 'ksonnet-util/kausal.libsonnet';
-local container = k.core.v1.container;
-local statefulset = k.apps.v1.statefulSet;
-local configMap = k.core.v1.configMap;
-
+(import 'prometheus/ha-mixin.libsonnet') +
```
dannykopping, line 1: I wonder if it might not make sense to `git mv prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet prometheus/ha-mixin.libsonnet`, thereby maintaining all the history of quite a fundamental file - and then creating a new file in `prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet` with the current contents.

dannykopping (approved):
One minor suggestion, otherwise LGTM.
You should probably get a review from one of the others with more context than me before merging

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
Duologic: I think this gets lost in the squash merge anyway. :neutral_face:

dannykopping: Don't think I follow.
If you `git mv`, the path of the file may change but its history will be maintained. Squashing will combine all your changes in this PR only.

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
```diff
@@ -1,104 +1 @@
-local k = import 'ksonnet-util/kausal.libsonnet';
```
malcolmholmes, line 1: suggestion:

```
(import 'prometheus/ha-mixin.libsonnet')
```

There are a few people using this mixin, by name, this will prevent that from failing.

malcolmholmes (commented):
How about adding `withHA():: (import 'ha-mixin.libsonnet'),` to the end of prometheus.libsonnet, and making that the recommended way of loading the feature, rather than invoking it by filename?

prometheus-ksonnet/lib/prometheus-ha-mixin.libsonnet
Duologic: Already there, please check again.

Duologic:
> How about adding `withHA():: (import 'ha-mixin.libsonnet'),` to the end of prometheus.libsonnet, and making that the recommended way of loading the feature, rather than invoking it by filename?

Already there, please check again.

malcolmholmes (approved):
Given that everything I proposed is actually already there, I think I can approve.

PR #409, opened 2020-12-23
Alertmanager: Only page for PD failures

beorn7:
Other integrations, in particular Slack, both fail more often and are
at the same time not used to deliver critical alerts. Therefore, we
don't want a failure there to trigger a page.

This requires an update of the alertmanager-mixin to the current
version.


Duologic (approved):

Duologic:
Maybe this is too specific for jsonnet-libs?

beorn7:
That's a good point. But I guess most organizations will override this parameter anyway, and then it doesn't matter if we set `.*` or `pagerduty`. Arguably, the latter will be a better default for many other organizations than the former (but I didn't want to call out a specific vendor in the upstream prometheus/alertmanager repo).

PR #410, opened 2020-12-27
feat(prometheus-metamonitoring): add new lib

Duologic:
Initial release of this library. A blogpost is on the way and I plan to document this with docsonnet very soon.


beorn7:
I'm still on vacation until 2021-01-04. I'll have a look ASAP.

beorn7 (approved):
Broadly looks right to me. I'd feel better if @malcolmholmes would do a sanity check on the Jsonnet parts.

malcolmholmes (approved):
I would rather have at least a minimal README, but it all looks good to me.

PR #414, opened 2021-01-02
K8s style observability resources

malcolmholmes:
Prometheus ksonnet consumes observability resources as hidden elements to prevent them being accessible to Kubernetes. This works well when the thing that is consuming the resources is also jsonnet - i.e. it has access to those hidden elements.

However, the library can happily accept resources such as this:
```
{
  mixins:: {
    foo: {
      grafanaDashboards: {...},
    },
  },
}
```

In this example, we could invoke `(import 'main.jsonnet').mixins` to gain access to the hidden mixins element, but still, the `grafanaDashboards` element is also set to hidden, and thus won't be exported.

A cleaner way is to allow the library (and associated sub-libraries) to communicate with the outside world using kubernetes style objects (e.g. having `apiVersion`/`kind` attributes).

This PR adds functions to each relevant library that can be invoked with `grr(import 'main.jsonnet)` and will return k8s style objects for everything that would otherwise be rendered from mixins/etc.

Grizzly will shortly have the ability to consume resources in this format. In fact, it is quite possible that this could become its default method.

Note, these are designed to look like CRDs. A preliminary search for existing CRDs didn't reveal anything hugely obvious to model after, but it would be reasonable to adapt this format to match any pre-existing CRDs for Grafana or Prometheus.

### DIFF
--- a/grizzly/README.md
+++ b/grizzly/README.md
@@ -12,7 +12,7 @@ or associated libraries, where other resources are using Tanka, place a file
 named `grr.jsonnet` next to your `main.jsonnet`, and give it this content:
 
 ```
-local main = (import 'main.jsonnet').o11y.o11y;
+local main = (import 'main.jsonnet');
 local grizzly = (import 'grizzly/grizzly.libsonnet');
 
 grizzly.fromPrometheusKsonnet(main)
--- a/grizzly/README.md
+++ b/grizzly/README.md
@@ -1,0 +1,23 @@
+# Grizzly Library
+
+This library provides utilities for [Grizzly](https://github.com/grafana/grizzly),
+a utility for managing observability resources on Grafana and hosted Prometheus
+installations.
+
+## Kubernetes Style Observability Objects
+It provides functions to render Kubernetes style objects from Monitoring Mixins.
+
+If deploying dashboards structured to be consumed by [prometheus-ksonnet](https://github.com/grafana/jsonnet-libs/prometheus-ksonnet)
+or associated libraries, where other resources are using Tanka, place a file
+named `grr.jsonnet` next to your `main.jsonnet`, and give it this content:
+
+```
+local main = (import 'main.jsonnet').o11y.o11y;
+local grizzly = (import 'grizzly/grizzly.libsonnet');
+
+grizzly.fromPrometheusKsonnet(main)
+```
+
+Then, you can invoke this from Grizzly like so:
+
+`grr show -r environments/<my-environment>/grr.libsonnet`
--- a/grizzly/grafana.libsonnet
+++ b/grizzly/grafana.libsonnet
@@ -1,0 +1,21 @@
+local util = import 'util.libsonnet';
+
+{
+  getFolder(main):: util.get(main, 'grafanaDashboardFolder', 'General'),
+
+  fromMap(dashboards, folder):: {
+    [k]: util.makeResource('Dashboard', k, dashboards[k], { folder: folder })
+    for k in std.objectFields(dashboards)
+  },
+
+  fromMixins(mixins):: {
+    [key]:
+      local mixin = mixins[key];
+      {
+        local folder = util.get(mixin, 'grafanaDashboardFolder', 'General'),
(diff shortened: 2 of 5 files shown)


prometheus-ksonnet/grafana/grr.libsonnet
```diff
@@ -0,0 +1,48 @@
+/*
+ * The `grr` function below will render Grafana resources generated by this
+ * Grafana library (e.g. as `grafanaDashboards`) into Kubernetes style objects,
+ * with kind/apiVersion. (Grizzly)[https://github.com/grafana/grizzly)
+ * has the ability to consume resources in this format.
```
Duologic, line 5: I like this, can you make it a separate library? The concept of $.mixins is bigger than prometheus-ksonnet.

prometheus/grr.libsonnet
```diff
@@ -0,0 +1,56 @@
+/*
+ * The `grr` function below will render prometheus resources generated by this
+ * prometheus library (e.g. in either `prometheusRules` or `prometheusAlerts`)
+ * into Kubernetes style objects, with kind/apiVersion. (Grizzly)[https://github.com/grafana/grizzly)
+ * has the ability to consume resources in this format.
```
Duologic, line 5: With above, this should be in one library, too much code duplication and grr.libsonnet is not relevant for deploying prometheus.

prometheus-ksonnet/grafana/grr.libsonnet
```diff
@@ -0,0 +1,48 @@
...
+ *
+ * It is invoked with a different entrypoint and handed your jsonnet as input. E.g
+ * if your main entrypoint is `main.jsonnet`, this can be invoked like so:
+ *
+ * grr.jsonnet:
+ * ```
+ * local grafana = import 'prometheus-ksonnet/grafana/grr.libsonnet';
+ * grafana.grr(import 'main.jsonnet')
+ */
+local grafanaAPI = 'grafana.com/v1';
```
Duologic, line 15: Does this version mean something? I'd refer to who is intended to consume it, for example `grizzly.dev/v1alpha1` or something, in line with `tanka.dev/v1alpha1`?

prometheus/grr.libsonnet
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * It is invoked with a different entrypoint and handed your jsonnet as input. E.g
+ * if your main entrypoint is `main.jsonnet`, this can be invoked like so:
+ *
+ * grr.jsonnet:
+ * ```
+ * local prom = import 'prometheus/grr.libsonnet';
+ * prom.grr(import 'main.jsonnet')
```
Duologic, line 13: The API could be something like his, not grabbing things it likes in out of root, rather allowing someone to feed in mixins. It could be named differently or be nested somewhere:

```jsonnet
local grr = import 'grizzly/grr.libsonnet';
local main = import 'main.jsonnet';

grr.prometheus.fromMixins(main.mixins)
+ grr.prometheus.fromRules(main.prometheusRules)
```

Duologic (commented):
I like the outcome, I don't like that it is scattered around other libraries.

prometheus-ksonnet/grafana/grr.libsonnet
malcolmholmes: This is fair enough, and your suggestion below about a library api works around my objections.

prometheus/grr.libsonnet
malcolmholmes: the prometheus library isn't just about deploying prometheus, it also covers configuring it with rules/alerts/etc. This is relevant to that portion.

prometheus-ksonnet/grafana/grr.libsonnet
malcolmholmes: given we don't own grizzly.dev, I suggest grafana.com/grizzly/v1alpha1

prometheus/grr.libsonnet
malcolmholmes: looks good

Duologic (approved):

PR #415, opened 2021-01-06
fix(ksonnet-util): prune instead of hide

Duologic:
Hiding is also affected volumeClaimTemplates that are added in a later mixin.

### DIFF
--- a/ksonnet-util/kausal.libsonnet
+++ b/ksonnet-util/kausal.libsonnet
@@ -190,7 +190,7 @@ k {
 
         // remove volumeClaimTemplates if empty (otherwise it will create a diff all the time)
         (if std.length(volumeClaims) == 0 then {
-           spec+: { volumeClaimTemplates:: {} },
+           spec+: std.prune({ volumeClaimTemplates: {} }),
          } else {}),
     },
   },


malcolmholmes:
How does this pruning work? Don't we still end up with {}? Or if not, doesn't the empty element get pruned away so removing nothing?

Duologic:
If therer are no claims, then we can safely assume `volumeClaimTemplates` equals `{}`. So, then what `std.prune()` does is remove 'keys' that are in fact a null value like [], {} and null, or that at least that is how I understand it.

From the docs:
> Recursively remove all "empty" members of a. "Empty" is defined as zero length `arrays`, zero length `objects`, or `null` values. The argument a may have any type.

The hardest part for me to understand is to know when `std.prune()` is applied, but it looks like this happens after the jsonnet tree has formed, how else could it do it recursively?

Side node: the 'recursively' is also treacherous, I learned it the hard way a while back while using it on a whole tree, it messed up the cert-manager CRDs that expected an empty array somewhere.

malcolmholmes:
I don't think this will work.
```
{
  spec: {
    volumeClaimTemplates: {},
  },
} + {
  spec+: std.prune({ volumeClaimTemplates: {} }),
}
```
In this scenario, we are adding `{}` to the spec, the prune will always reduce to just `{}` which is a no-op.

Can we come up with some test cases that prove the code here works as expected?

Duologic:
I'm not doing that, am I? This only happens inside the new() function, it you doing something else outside that function, then that is not going to solve that.

Duologic:
Covered for in #426

PR #421, opened 2021-01-14
feat: add pentagon library

Duologic:
Publicize our Pentagon library.

### DIFF
--- a/pentagon/README.md
+++ b/pentagon/README.md
@@ -1,0 +1,24 @@
+# Pentagon
+
+This is a library to deploy [Grafana's fork of Pentagon](https://github.com/grafana/pentagon).
+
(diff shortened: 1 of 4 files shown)


pentagon/README.md
```diff
@@ -0,0 +1,24 @@
+# Pentagon
+
+This is a library to deploy [Grafana's fork of Pentagon](https://github.com/grafana/pentagon).
+
+The library is intended to deploy 1 pentagon instance per namespace.
```
captncraig, line 5: Probably need to make a note about the service account / namespace combinations it will authenticate to vault with, and note that vault policy management is out of scope for this.

captncraig (approved):

Issue #422, opened 2021-01-14
Can't apply prometheus-ksonnet/ instructions 


colega:
I was following the instructions in [`prometheus-ksonnet/README.md`](https://github.com/grafana/jsonnet-libs/blob/master/prometheus-ksonnet/README.md) and I encountred with the following issue:

```
~/tmp/k8s-config 17:04:30
$ tk init
GET https://github.com/ksonnet/ksonnet-lib/archive/0d2f82676817bbf9e4acf6495b2090205f323b9f.tar.gz 200
GET https://github.com/grafana/jsonnet-libs/archive/fd1d2f7e962560e3a20abfed161c70f91c991964.tar.gz 200
Directory structure set up! Remember to configure the API endpoint:
`tk env set environments/default --server=https://127.0.0.1:6443`

~/tmp/k8s-config 17:04:50
$ export CONTEXT=$(kubectl config current-context)

~/tmp/k8s-config 17:04:53
$ tk env set environments/default  --server-from-context=$CONTEXT
updated spec.apiServer (`https:// -> `https://kubernetes.docker.internal:6443`)

~/tmp/k8s-config 17:04:58
$ jb install github.com/grafana/jsonnet-libs/prometheus-ksonnet
GET https://github.com/grafana/jsonnet-libs/archive/fd1d2f7e962560e3a20abfed161c70f91c991964.tar.gz 200
GET https://github.com/grafana/grafana/archive/21ae4802ceba6a8f1de908fba17d2e3c5a220e42.tar.gz 200
GET https://github.com/grafana/jsonnet-libs/archive/fd1d2f7e962560e3a20abfed161c70f91c991964.tar.gz 200
GET https://github.com/kubernetes-monitoring/kubernetes-mixin/archive/4a8e078147dbca51067521e6ac59c7b54d44d3bd.tar.gz 200
GET https://github.com/prometheus/alertmanager/archive/ce108378d4c8d580804452dcc3a31222067793b2.tar.gz 200
GET https://github.com/prometheus/node_exporter/archive/0ce1f39c64a0a37d5aa81da8006661920b097e82.tar.gz 200
GET https://github.com/grafana/jsonnet-libs/archive/fd1d2f7e962560e3a20abfed161c70f91c991964.tar.gz 200
GET https://github.com/grafana/jsonnet-libs/archive/fd1d2f7e962560e3a20abfed161c70f91c991964.tar.gz 200
[...]

PR #423, opened 2021-01-15
Fix steps in prometheus-ksonnet/README.md

colega:
Solves #422

We need a custom k8s jsonnet artifacts to make it work.

Also removed `$` from bash code blocks to make the code easier to copy and paste directly into your terminal.

We should consider checking that this steps work in the CI.


jesusvazquez (approved):
I just tried this locally on my computer and it worked :+1:

pracucci (approved):
LGTM, thanks!

PR #424, opened 2021-01-15
Add Caddy mixin

besartberisha:
This PR adds a simple Caddy mixin, will probably spend some time in the future to see if I can improve it a bit and add more information as well.

Screen Shot 2021-01-15 at 13 18 08


hoenn (approved):

malcolmholmes:
Can this README be fixed to note what Caddy is? As we add more mixins, we can't assume that it is clear what these things are. Even just a link to the Caddy website would help.

PR #426, opened 2021-01-17
refactor: kausal.libsonnet

Duologic:
Another attempt at refactoring kausal.libsonnet. This time with a bigger effort to maintain compatibility with ksonnet-lib.

This is intended to be a no-op (beyond some APIVersion/Kind changes).

As well as reordering some files, it adds compatibility functions for users of both ksonnet-lib and k8s-alpha, increasing interoperability for libraries that target one or the other.

If `kausal-shim.libsonnet` is imported in `lib/k.libsonnet` for k8s-alpha, it should be removed as this compatibility layer is moved into this library..

refs: #215 #293


ksonnet-util/k-mixin.libsonnet
```diff
@@ -0,0 +1,114 @@
+// Override defaults paramters for objects in the ksonnet libs here.
```
malcolmholmes, line 1:
```suggestion
// Override defaults parameters for objects in the ksonnet libs here.
```

malcolmholmes (approved):
This looks great. One separation of adaptations to k.libsonnet alongside replacing references to $ with references to k. This will fix some issues I've had with this library.

malcolmholmes (approved):
repeating my approval. This is good stuff. Gets us that bit closer to a full migration to k8s-alpha

PR #428, opened 2021-01-18
fix(ksonnet-util): allow overriding root level k import

Duologic:
If someone wants to extend kausal.libsonnet on the root level, the imported util.libsonnet wouldn't take this change
into account into its subsequent functions.

This change recovers that feature while retaining the ability to use util.libsonnet as a separate library.

Minimal reproducible usecase:

```
(import 'ksonnet-util/kausal.libsonnet')
{
  core+: {
    v1+: {
      // Extend service
      service+: {
        new(name, selector={}, ports=[])::
          super.new(name, selector, ports)
          + {
            dns: std.trace('a', '%s.%s.svc.cluster.local' % [self.metadata.name, $._config.namespace]),
          },
      },
    },
  },
}
+
{
  _config+:: {
    namespace: 'test',
  },

  sts:
    $.apps.v1.statefulSet.new('name', 1, [], volumeClaims=[], podLabels={})
    + $.util.antiAffinity,

  svc: $.util.serviceFor(
    self.sts,
  ),
}
```

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -1,6 +1,6 @@
 // util.libsonnet provides a number of useful (opinionated) shortcuts to replace boilerplate code
 
-function(k=(import 'grafana.libsonnet')) {
+local util(k) = {
   // mapToFlags converts a map to a set of golang-style command line flags.
   mapToFlags(map, prefix='-'): [
     '%s%s=%s' % [prefix, key, map[key]]
@@ -242,4 +242,8 @@ function(k=(import 'grafana.libsonnet')) {
   local deployment = k.apps.v1.deployment,
   podPriority(p):
     deployment.mixin.spec.template.spec.withPriorityClassName(p),
+};
+
+util((import 'grafana.libsonnet')) + {
+  withK(k):: util(k),
 }
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -1,7 +1,6 @@
 // util.libsonnet provides a number of useful (opinionated) shortcuts to replace boilerplate code
-local k = import 'grafana.libsonnet';
 
-{
+function(k=(import 'grafana.libsonnet')) {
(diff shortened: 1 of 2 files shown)


Duologic:
@xvzf Please have a go at this branch.

xvzf (approved):
It looks a bit hacky but it works 👍🏿

ksonnet-util/util.libsonnet
```diff
@@ -1,7 +1,6 @@
 // util.libsonnet provides a number of useful (opinionated) shortcuts to replace boilerplate code
-local k = import 'grafana.libsonnet';
 
-{
+function(k=(import 'grafana.libsonnet')) {
```
malcolmholmes, line 3: this means we need to consume `util.libsonnet` with `local util = import 'ksonnet-util/util.libsonnet'()`, which is unusual, correct?

malcolmholmes:
```
local k = import 'grafana.libsonnet';
local _util = import '_util.libsonnet'; // <-- current contents of util.libsonnet go here
_util(k) + {
  withMyK(k): _util(k),
}
```
Now, we can do `import 'ksonnet-util/util.libsonnet'` as normal, but also: `(import 'util.libsonnet').withMyK(this)`

xvzf:
Maybe it make more sense to extend the standard library within `lib/k.libsonnet`?

Duologic:
`k-compat.libsonnet` imports `k.libsonnet`, so yes, you can extend it there, but also after `kausal.libsonnet`. Either are valid use cases that were possible before the refactoring and should be possible after this PR.

ksonnet-util/util.libsonnet
Duologic: Excellent feedback, the top level function would indeed require `(import 'ksonnet-util/util.libsonnet')()`, have a look at 79d4a3c, that should prevent it.

malcolmholmes (approved):
Looks great. So simple!

xvzf:
You rock. Thanks guys!

PR #429, opened 2021-01-19
PG Bouncer Mixin added, based on nodejs mixin

mattdurham:
Creation of initial mixin for PGBouncer


sdboyer (changes requested):
In general we should be including $job and $instance variables in the dashboard as template variables.+ on the queries, so that the dashboard can at least translate into a world where there are multiple pgbouncer instances being scraped and feeding back into a central place

pgbouncer-mixin/README.md
```diff
@@ -0,0 +1,14 @@
+PgBouncer Mixin
+
+The PgBouncer Mixin is a set of configurable, reusable, and extensible alerts and dashboards based on the metrics exported by the PgBouncer Exporter. The mixin creates recording and alerting rules for Prometheus and suitable dashboard descriptions for Grafana.
```
malcolmholmes, line 3: Can we have a sentence describing what PgBouncer is, for those that don't know? Perhaps a link to it, and to the PgBouncer Exporter?

PR #436, opened 2021-01-22
mysql-exporter: allow credentials from env vars

dafydd-t:
Currently the mysql-exporter requires `config.mysql_user` and `config.mysql_password` as jsonnet variables to build the `DATA_SOURCE_NAME` env var. 

To enable the mysql credentials to be sourced from some key store, this change creates env vars for the `mysql_user` and `mysql_password`, and uses those env vars to build the `DATA_SOURCE_NAME` variable.

This change also introduces the ability to optionally define the full host name instead just the deployment name and namespace.

There are assertions for the following:
- `mysql_password` is no longer required, but exactly one of `mysql_password` or `mysql_password_secret` must be defined
- exacrly one of `mysql_fqdn` or (`deployment_name` and `namespace`) must be defined

This change should be backwards compatible.

### DIFF
--- a/mysql-exporter/mysql-exporter.libsonnet
+++ b/mysql-exporter/mysql-exporter.libsonnet
@@ -8,22 +8,22 @@ local secrets = import 'secrets.libsonnet';
 
 secrets {
   image:: 'prom/mysqld-exporter:v0.12.1',
+  mysql_fqdn:: '%s.%s.svc.cluster.local' % [
+    $._config.deployment_name,
+    $._config.namespace,
+  ],
 
   _config:: {
     mysql_user: error 'must specify mysql user',
     mysql_password: '',
-    mysql_fqdn:: '%s.%s.svc.cluster.local' % [
-      $._config.deployment_name,
-      $._config.namespace,
-    ],
-    deployment_name: '',
-    namespace: '',
+    deployment_name: error 'must specify deployment name',
+    namespace: error 'must specify namespace',
   },
 
   mysqld_exporter_env:: {
     MYSQL_USER: $._config.mysql_user,
     MYSQL_PASSWORD: $._config.mysql_password,
-    MYSQL_HOST: $._config.mysql_fqdn,
+    MYSQL_HOST: $.mysql_fqdn,
   },
 
   mysqld_exporter_container::
@@ -32,11 +32,12 @@ secrets {
     container.withArgsMixin([
       '--collect.info_schema.innodb_metrics',
     ]) +
-    container.withEnvMap($.mysqld_exporter_env) +
+    container.withEnvMap($.mysqld_exporter_env) + 
     // Force DATA_SOURCE_NAME to be declared after the variables it references
     container.withEnvMap({
       DATA_SOURCE_NAME: '$(MYSQL_USER):$(MYSQL_PASSWORD)@tcp($(MYSQL_HOST):3306)/',
     }),
+
 
   mysql_exporter_deployment:
     deployment.new('%s-mysql-exporter' % $._config.deployment_name, 1, [$.mysqld_exporter_container]),
--- a/mysql-exporter/mysql-exporter.libsonnet
+++ b/mysql-exporter/mysql-exporter.libsonnet
@@ -1,4 +1,4 @@
-local k = import 'github.com/grafana/jsonnet-libs/ksonnet-util/kausal.libsonnet';
+local k = import 'ksonnet-util/kausal.libsonnet';
 local container = k.core.v1.container;
 local service = k.core.v1.service;
 local containerPort = k.core.v1.containerPort;
--- a/mysql-exporter/mysql-exporter.libsonnet
+++ b/mysql-exporter/mysql-exporter.libsonnet
@@ -13,7 +13,7 @@ local mysql_credential(config) =
     error 'must define one of _config.mysql_password or _config.mysql_password_secret.'
   else if std.length(config.mysql_password) > 0 then
     [{ name: 'MYSQL_PASSWORD', value: config.mysql_password }]
-  else [envVar.fromSecretRef('MYSQL_PASSWORD', config.mysql_password_secret, 'password')];
+  else [envVar.fromSecretRef('MYSQL_PASSWORD', config.mysql_password_secret, config.mysql_password_secret_key)];
 
 
 local mysql_host(config, fqdn) =
@@ -30,6 +30,7 @@ local mysql_host(config, fqdn) =
     mysql_user: error 'must specify mysql user',
     mysql_password: '',
     mysql_password_secret: '',
+    mysql_password_secret_key: 'password',
     deployment_name: error 'must specify deployment name',
     namespace: error 'must specify namespace',
   },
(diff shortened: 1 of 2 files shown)


mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -3,38 +3,61 @@ local container = k.core.v1.container;
...
-  mysql_fqdn:: '',
+  assert (
+    !(std.length($._config.mysql_password) == 0 && std.length($._config.mysql_password_secret) == 0) &&
+    !(std.length($._config.mysql_password) > 0 && std.length($._config.mysql_password_secret) > 0)
+  ) : 'mysql-exporter: exactly one of _config.mysql_password and _config.mysql_password_secret must be defined.',
+
+  assert (
+    (!(std.length($.mysql_fqdn) == 0 && (std.length($._config.deployment_name) == 0 || std.length($._config.namespace) == 0))) &&
+    (!(std.length($.mysql_fqdn) > 0 && !(std.length($._config.deployment_name) == 0 && std.length($._config.namespace) == 0)))
+  ) : 'mysql-exporter: exactly one of (_config.deployment_name and _config.namespace) or mysql_fqdn must be specified.',
```
Duologic, line 45: Due to the lazy loading nature of jsonnet, this many conditionals is usually not needed.

mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -3,38 +3,61 @@ local container = k.core.v1.container;
...
+    '%s' % fqdn
+  else
+    '%s.%s.svc.cluster.local' % [
+      config.deployment_name,
+      config.namespace,
+    ];
 
 {
+  image:: 'prom/mysqld-exporter:v0.12.1',
+  mysql_fqdn:: '',
```
Duologic, line 26:
```suggestion
  mysql_fqdn:: '%s.%s.svc.cluster.local' %
               [
                 $._config.deployment_name,
                 $._config.namespace,
               ],
```
This gets rid of the `mysql_host()` function, if `$.mysql_fqdn is set by the use, this will be overridden and above errors will simply be ignored.

mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -3,38 +3,61 @@ local container = k.core.v1.container;
...
       '--collect.info_schema.innodb_metrics',
-    ]),
+    ]) +
+    container.withEnvMixin(
+      mysql_credential($._config) +
+      [
+        { name: 'MYSQL_USER', value: $._config.mysql_user },
+        { name: 'DATA_SOURCE_NAME', value: '$(MYSQL_USER):$(MYSQL_PASSWORD)@tcp(%s:3306)/' % mysql_host($._config, $.mysql_fqdn) },
+      ]
+    ),
```
Duologic, line 59: Then you can put everything in a map and overwrite the items in this map if you want to do more elaborate configurations.
```
  mysqld_exporter_env:: {
    MYSQL_USER: $._config.mysql_user,
    MYSQL_PASSWORD: $._config.mysql_password,
    MYSQL_HOST: $.mysql_fqdn,
    DATA_SOURCE_NAME: '$(MYSQL_USER):$(MYSQL_PASSWORD)@tcp($(MYSQL_HOST):3306)/',
  },

  mysqld_exporter_container::
    container.new('mysqld-exporter', $.image) +
    container.withEnvMap($.mysqld_exporter_env) +
    ...
```

mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -3,38 +3,61 @@ local container = k.core.v1.container;
 local service = k.core.v1.service;
 local containerPort = k.core.v1.containerPort;
 local deployment = k.apps.v1.deployment;
+local envVar = k.core.v1.envVar;
+
+
+local mysql_credential(config) =
+  if std.length(config.mysql_password) > 0 then
+    [{ name: 'MYSQL_PASSWORD', value: config.mysql_password }]
+  else [envVar.fromSecretRef('MYSQL_PASSWORD', config.mysql_password_secret, config.mysql_password_secret_key)];
```
Duologic, line 12: With the env dict, you can now create a mixin function like this that would replace this `mysql_credentials()` function.

```
  withSecretPassword(name, key='password'):: {
    env+:: {
      MYSQL_PASSWORD:: '',
    },
    mysqld_exporter_container+::
      container.withEnvMixin([
        envVar.fromSecretRef('MYSQL_PASSWORD', name, key),
      ]),
  },
```

Duologic (commented):
With these suggestions, I think we can get rid of most, if not all conditionals.

dafydd-t:
> With these suggestions, I think we can get rid of most, if not all conditionals.

@Duologic thank you, these suggestions were really informative.

mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -1,40 +1,40 @@
...
+secrets {
+  image:: 'prom/mysqld-exporter:v0.12.1',
+
   _config:: {
     mysql_user: error 'must specify mysql user',
-    mysql_password: error 'must specify mysql password',
-    deployment_name: error 'must specify deployment name',
-    namespace: error 'must specify namespace',
+    mysql_password: '',
+    mysql_fqdn:: '%s.%s.svc.cluster.local' % [
```
Duologic, line 15: Are you aware that you moved `mysql_fqdn` into `_config`.

In the unlikely case that someone is using this lib, that would be an api change.

mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -1,40 +1,40 @@
...
+
+secrets {
+  image:: 'prom/mysqld-exporter:v0.12.1',
+
   _config:: {
     mysql_user: error 'must specify mysql user',
-    mysql_password: error 'must specify mysql password',
-    deployment_name: error 'must specify deployment name',
-    namespace: error 'must specify namespace',
+    mysql_password: '',
```
Duologic, line 14: You can put the error back here and below in case you want to require it.

Duologic (approved):
This looks great, very clean.

mysql-exporter/mysql-exporter.libsonnet
dafydd-t: Ah yes. I thought it was tidier, but there's a risk of breaking something. I'll move it back. Thanks!

dafydd-t:
Ah I'm finding one issue with this at the moment, which is that the order in which the environment variables is defined is not guaranteed.

This causes an issue because when `DATA_SOURCE_NAME` gets declared first, the other variables do not exist. Therefore we end up with a host in the DSN of value `$(MYSQL_HOST)`.

I'll take a look at how we can force the ordering of these environment variables.

dafydd-t:
I forced the `DATA_SOURCE_NAME` env to be declared after the variables it references by moving it to a separate function called after the other env vars.

mysql-exporter/mysql-exporter.libsonnet
dafydd-t: I don't think we want to require `mysql_password`. If the user uses the `withSecretPassword` mixin, they won't need a password.

Duologic: If one is using the mixin, it would not touch the error as it does lazy loading.

Duologic (approved):

PR #437, opened 2021-01-24
Add simple datasource functions

malcolmholmes:
There are various functions in this file to add new datasources, but they all include the word 'grafana' in the name.

This naming scheme is based upon the assumption that the library will be merged into the root of the jsonnet tree, which is a bad practice anyway. If, however, this grafana library is loaded into a local variable called 'grafana', then, with this new change, we can do `grafana.datasource.new()` which makes plenty sense.

### DIFF
--- a/prometheus-ksonnet/grafana/datasources.libsonnet
+++ b/prometheus-ksonnet/grafana/datasources.libsonnet
@@ -1,4 +1,26 @@
 {
+  datasource+:: {
+    new(name, url, type, default=false):: {
+      name: name,
+      type: type,
+      access: 'proxy',
+      url: url,
+      isDefault: default,
+      version: 1,
+      editable: false,
+    },
+    withBasicAuth(username, password):: {
+      basicAuth: true,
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/grafana/datasources.libsonnet
```diff
@@ -25,6 +25,29 @@
...
+      jsonData: {
+        httpMethod: method,
+      },
+    },
+    withBasicAuth(username, password):: {
+      basicAuth: true,
+      basicAuthUser: username,
+      basicAuthPassword: password,
+    },
+    withJsonData(data):: {
```
Duologic, line 46:
```suggestion
    withJsonData(data, method='GET'):: {
```
?

malcolmholmes: I've achieved this by removing it from the `new()` method and adding a `withHttpMethod()` function.

Duologic (approved):
clean

PR #440, opened 2021-01-27
mysql exporter: force password env to be defined before `DATA_SOURCE_NAME`

dafydd-t:
I made a PR recently which allows the mysql password to be provided as a environment variable from a secret, using the `withPasswordSecret` function. However, if you use this function, the new `MYSQL_PASSWORD` envVar is defined after `DATA_SOURCE_NAME` in the container env list. This means `DATA_SOURCE_NAME` will not reference `MYSQL_PASSWORD` correctly in this case.

This PR proposes adding a kind of placeholder where additional envs can get added before `DATA_SOURCE_NAME` to avoid this problem.

I'm happy to take suggestions for alternative ways of solving this problem!


mysql-exporter/mysql-exporter.libsonnet
```diff
@@ -26,22 +24,35 @@ secrets {
...
     }),
 
-
   mysql_exporter_deployment:
     deployment.new('%s-mysql-exporter' % $._config.deployment_name, 1, [$.mysqld_exporter_container]),
 
   mysql_exporter_deployment_service:
     k.util.serviceFor($.mysql_exporter_deployment),
+
+  withSecretPassword(name, key='password'):: {
```
dafydd-t, line 49: I moved this function into this file because its not being used anywhere else.

The behaviour of the function has also changed. It writes the secret ref to the new list, rather than appending to the container definition.

Duologic (approved):
This is definitely a weird quirk, I didn't think order mattered, but apparently for interpolation it is a big deal.

PR #443, opened 2021-02-01
fix: remove unnecessary certs and tokens from scrape config

### LINKED ISSUE #442: prometheus-ksonnet: `bearer_token_file` configured in kubernetes_sd_configs that don't require authorization
For example in [`kubernetes-pods`](https://github.com/grafana/jsonnet-libs/blob/6db386fca27344b0642b945f34fd911ead28d8af/prometheus-ksonnet/lib/prometheus-config.libsonnet#L61). The service account file and Kubernetes server certificate is configured for the scrape config. Presumably this was actually meant for authentication to the Kubernetes API server for service discovery (or it exists as an artifact of copy-paste).

The service discovery still works as we don't specify and API server in the `kubernetes_sd_config` block and Prometheus automatically loads the certificate and service account token.

https://prometheus.io/docs/prometheus/latest/configuration/configuration/#kubernetes_sd_config
> \# The API server addresses. If left empty, Prometheus is assumed to run inside
> \# of the cluster and will discover API servers automatically and use the pod's
> \# CA certificate and bearer token file at /var/run/secrets/kubernetes.io/serviceaccount/.

jdbaldry:
Fixes #442

### DIFF
--- a/prometheus-ksonnet/lib/prometheus-config.libsonnet
+++ b/prometheus-ksonnet/lib/prometheus-config.libsonnet
@@ -235,12 +235,6 @@
           },
         }],
 
-        tls_config: {
-          ca_file: '/var/run/secrets/kubernetes.io/serviceaccount/ca.crt',
-          insecure_skip_verify: $._config.prometheus_insecure_skip_verify,
-        },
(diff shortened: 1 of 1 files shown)


jdbaldry:
Closing because of spelling :D

PR #448, opened 2021-02-05
Add jsonnetfile for the envoy mixin.

cristiangreco:
Add jsonnetfile for the envoy mixin and fix other linting errors.

This fixes the `mixin lint` step with more recent versions of the
`mixtool` tool. The problem is likely not appearing in CI as the
`mixtool` version used is a few months old.

I've updated the Dockerfile as well to use the latest available revision
of `mixtool`, so the a new docker image should be published once this is
merged.

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -18,7 +18,7 @@ RUN chmod +x /usr/bin/jb
 
 # Build mixtool
 FROM golang:1.15-alpine AS mixtool-builder
-RUN GO111MODULE=on go get github.com/monitoring-mixins/mixtool/cmd/mixtool@59d44357240d
+RUN GO111MODULE=on go get github.com/monitoring-mixins/mixtool/cmd/mixtool@ae18e31161ea
 
 FROM alpine:3.7
 RUN apk --no-cache add make python git openssh-client bash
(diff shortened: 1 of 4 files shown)


consul-mixin/dashboards.libsonnet
```diff
@@ -14,7 +14,7 @@ local panel_settings = {
 {
   grafanaDashboards+:: {
     'consul-overview.json':
-      g.dashboard('Consul Overview')
+      g.dashboard('Consul Overview', std.md5('20210205-consul'))
```
cristiangreco, line 17: mixtool was complaining about missing UID.

build/Dockerfile
```diff
@@ -18,7 +18,7 @@ RUN chmod +x /usr/bin/jb
 
 # Build mixtool
 FROM golang:1.15-alpine AS mixtool-builder
-RUN GO111MODULE=on go get github.com/monitoring-mixins/mixtool/cmd/mixtool@59d44357240d
+RUN GO111MODULE=on go get github.com/monitoring-mixins/mixtool/cmd/mixtool@ae18e31161ea
```
cristiangreco, line 21: Once this change is merged, image needs to be published and circleci config needs to be updated.

sdboyer (approved):
i'm good to try running ahead with these

PR #450, opened 2021-02-08
Grizzly lib updates

malcolmholmes:
Firstly, let's use `v1alpha1` as we're in now way sure about this API.

Secondly, replace a reference to the incorrect kind `RuleGroup` with  the correct `PrometheusRuleGroup`.

Finally, tidy up a little with some local use.


malcolmholmes:
Note, the [kubernetes-style-resources](https://github.com/grafana/grizzly/pull/72) PR on Grizzly matches this API version now.

Duologic (approved):

PR #458, opened 2021-02-11
Set shared tooltip to false by default in grafana-builder but don't enforce it. 

bergquist:


malcolmholmes (approved):

bboreham:
The change in `prometheus-ksonnet/grafana/dashboards.libsonnet` does not seem to be covered by the commit message.

PR #459, opened 2021-02-12
fix(prometheus): add antiAffinity for HA sts

Duologic:
I think this makes sense here, a HA set should not run on the same node.

for ref: https://github.com/grafana/jsonnet-libs/blob/2cef89c/ksonnet-util/util.libsonnet#L227-L238


beorn7 (approved):
Yes, I think this makes sense.

jvrplmlmn (approved):
👍

PR #460, opened 2021-02-12
feat(grafana): Add convenience functions for enterprise license configuration

jdbaldry:
https://grafana.com/docs/grafana/latest/enterprise/enterprise-configuration/

### DIFF
--- a/grafana/config.libsonnet
+++ b/grafana/config.libsonnet
@@ -55,4 +55,20 @@
       },
     },
   }),
+
+  withEnterpriseLicenseText(text):: self.withGrafanaIniConfig({
+    sections+: {
+      enterprise+: {
+        license_text: text,
+      },
+    },
+  }),
+
+  withEnterpriseLicensePath(path):: self.withGrafanaIniConfig({
+    sections+: {
+      enterprise+: {
+        license_path: path,
+      },
+    },
+  }),
 }


malcolmholmes:
Questions:
* If someone passes in a license text, then great, that's simple. If they want to provide a path, surely we should do the handling of that? e.g. mount it as a ConfigMap or as a secret into the Grafana container? Perhaps we could optionally create the secret too.
* Presumably we want to change the image if it is going to be enterprise?

jdbaldry:
Do you want a separate PR for the fix commit?

>If someone passes in a license text, then great, that's simple. If they want to provide a path, surely we should do the handling of that? e.g. mount it as a ConfigMap or as a secret into the Grafana container? Perhaps we could optionally create the secret too.
Honestly, I feel like this is a little brittle. With Jsonnet's hermetic nature, the license file would need to be local for that to work and I would say it's probably just easier and less prone to breaking to do `withEnterpriseLicenseText(import 'path/to/license')`

> Presumably we want to change the image if it is going to be enterprise?
Is there a separate docker image for this?

malcolmholmes:
> Do you want a separate PR for the fix commit?

No, here is fine.
 
> > If someone passes in a license text, then great, that's simple. If they want to provide a path, surely we should do the handling of that? e.g. mount it as a ConfigMap or as a secret into the Grafana container? Perhaps we could optionally create the secret too.
> 
> Honestly, I feel like this is a little brittle. With Jsonnet's hermetic nature, the license file would need to be local for that to work and I would say it's probably just easier and less prone to breaking to do `withEnterpriseLicenseText(import 'path/to/license')`.

That can work for now, although I'm sure we'll see demand for secrets eventually.
 
> > Presumably we want to change the image if it is going to be enterprise?
> Is there a separate docker image for this?

I'm not sure. If the current image is enterprise capable, then we have no issue!

jdbaldry:
Yeah I think the Grafana image is the same for Enterprise, you only need to supply a valid license to activate it.

jdbaldry:
@malcolmholmes Are you happy with this for now?

malcolmholmes (approved):
Looks great to me!

PR #462, opened 2021-02-12
Must also add non-mixin dashboards

malcolmholmes:
We need to add dashboards that are *not* in a mixin into the General folder.


malcolmholmes:
Seems this also fixes a bug where the General folder is missing an id, causing rendering failure.

jdbaldry (approved):
LGTM

PR #465, opened 2021-02-15
Don't create CMs for empty folders/mixins

malcolmholmes:


grafana/configmaps.libsonnet
```diff
@@ -77,7 +78,7 @@ local configMap = k.core.v1.configMap;
   ),
 
   // Helper to mount a variable number of sharded config maps.
-  local shardedMounts(folder) =
+  local shardedMounts(folder) = if std.length(folder.dashboards) == 0 then {} else
```
Duologic, line 81:
```suggestion
  local shardedMounts(folder) = 
    if std.length(folder.dashboards) == 0 
    then {} 
    else
```

Duologic (approved):

PR #471, opened 2021-02-17
Set BasicAuth for Grafana Datasource in secureJson

colega:
Having basicAuthPassword in the plugin configuration body is deprecated according to the provisioning docs (https://grafana.com/docs/grafana/latest/administration/provisioning/)

Now the password is located in the secureJsonData

Since there are inevitably outdated plugins out there, I propse to have an optional `legacy` flag that would default to the newer (non-deprecated) behavior.

### DIFF
--- a/grafana/datasources.libsonnet
+++ b/grafana/datasources.libsonnet
@@ -13,7 +13,7 @@
     basicAuthUser: username,
     basicAuthPassword: if legacy then password,
     secureJsonData+: if !legacy then {
-        basicAuthPassword: password,
+      basicAuthPassword: password,
     },
   },
   withJsonData(data):: {
--- a/grafana/datasources.libsonnet
+++ b/grafana/datasources.libsonnet
@@ -12,7 +12,7 @@
     basicAuth: true,
     basicAuthUser: username,
     [if legacy then 'basicAuthPassword']: password,
-    secureJsonData+: if !legacy then {
+    [if !legacy] then 'secureJsonData']+: {
       basicAuthPassword: password,
     },
   },
--- a/grafana/datasources.libsonnet
+++ b/grafana/datasources.libsonnet
@@ -11,7 +11,7 @@
   withBasicAuth(username, password, legacy=false):: {
     basicAuth: true,
     basicAuthUser: username,
-    basicAuthPassword: if legacy then password,
+    [if legacy then 'basicAuthPassword']: password,
     secureJsonData+: if !legacy then {
       basicAuthPassword: password,
     },
--- a/grafana/datasources.libsonnet
+++ b/grafana/datasources.libsonnet
@@ -12,7 +12,7 @@
     basicAuth: true,
     basicAuthUser: username,
     [if legacy then 'basicAuthPassword']: password,
-    [if !legacy] then 'secureJsonData']+: {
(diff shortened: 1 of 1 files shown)


malcolmholmes:
Would be great to get this change in, we just need to validate it on some real infrastructure before rolling it out.

colega:
This isn't backwards compatible, as we'd need to add `legacy=true` to the existing usages to be compatible.

malcolmholmes:
what do you mean by isn't backwards compatible? It will certainly create a diff across existing data sources, but presumably using a `secure` password is better? Or is it that not all datasources use `secure` yet?

colega:
I tried passing it through the secure json to Prometheus datasource, and it didn't work.

colega:
Just saw this in our Grafana's logs:
```
t=2021-03-23T14:39:23+0000 lvl=warn msg="[Deprecated] the use of basicAuthPassword field is deprecated. Please use secureJsonData.basicAuthPassword" logger=provisioning.datasources datasource name="Prometheus (41413 @ cortex-dev-01)"
```

colega:
Checked that Prometheus datasource works properly with this change.

tonychoe:
Looks like Grafana 6.2 started supporting the password in secureJsonData per https://github.com/grafana/grafana/pull/16175

i think it’s quite safe to not ask if`legacy`.

colega:
As far as I understand there still could be datasources that don't read the `secureJsonData` and I understand that Grafana doesn't magically leak the secureJsonData password into plain json data, so it would be safer to leave the option.

grafana/datasources.libsonnet
```diff
@@ -8,10 +8,13 @@
     version: 1,
     editable: false,
   },
-  withBasicAuth(username, password):: {
+  withBasicAuth(username, password, legacy=false):: {
     basicAuth: true,
     basicAuthUser: username,
-    basicAuthPassword: password,
+    basicAuthPassword: if legacy then password,
```
malcolmholmes, line 14:
```suggestion
    [if legacy then 'basicAuthPassword']: password,
```

grafana/datasources.libsonnet
```diff
@@ -8,10 +8,13 @@
...
+    secureJsonData+: if !legacy then {
```
malcolmholmes, line 15:
```suggestion
    [if !legacy] then 'secureJsonData']+: {
```

malcolmholmes (commented):
Legacy is a useful thing to have. Would be good to have a comment about what it is (e.g. what Grafana version introduced the secure option).

My suggestion will hide the elements if necessary, rather than having them be present, but null.

grafana/datasources.libsonnet
```diff
@@ -8,10 +8,13 @@
     version: 1,
     editable: false,
   },
-  withBasicAuth(username, password):: {
+  withBasicAuth(username, password, legacy=false):: {
     basicAuth: true,
     basicAuthUser: username,
-    basicAuthPassword: password,
+    [if legacy then 'basicAuthPassword']: password,
+    [if !legacy] then 'secureJsonData']+: {
```
colega, line 15:
```suggestion
    [if !legacy then 'secureJsonData']+: {
```

malcolmholmes (approved):

PR #472, opened 2021-02-17
feat(alertmanager/prometheus): abstract building blocks for global alertmanager setup

Duologic:
This PR reconstructs a powerful idea from prometheus-ksonnet into building blocks to configure a global alertmanager
cluster. I've included inline documentation that should explain these building blocks, the intention is to document the
respective libraries with docsonnet at some point.

I expect this to be fully backwards compatible on the prometheus-ksonnet end, however a small no-op diff should be
expected.

### DIFF
--- a/alertmanager/alertmanager.libsonnet
+++ b/alertmanager/alertmanager.libsonnet
@@ -49,12 +49,12 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
       memory: '40Mi',
     }),
 
-  // `buildPeers` constructs an array of alertmanager gossiping peer
-  // URLs. Together with `withAlertmanagers` in the prometheus
-  // jsonnetlib, this is a building block for configuring one global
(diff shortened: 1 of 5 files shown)


alertmanager/alertmanager.libsonnet
```diff
@@ -49,6 +49,40 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
       memory: '40Mi',
     }),
 
+  // `buildPeers` constructs an array of alertmanager gossiping peer
+  // URLs. Together with `withAlertmanagers` in the prometheus
```
malcolmholmes, line 53:
```suggestion
  // URLs. Note, 'gossiping' is the method that a cluster of alertmanagers use to communicate.
  Together with `withAlertmanagers` in the prometheus
```

malcolmholmes (approved):

PR #473, opened 2021-02-18
fix(prom-ksonnet): fix alertmanager_cluster_self

Duologic:
ref: https://github.com/grafana/jsonnet-libs/pull/472/

### DIFF
--- a/prometheus-ksonnet/lib/config.libsonnet
+++ b/prometheus-ksonnet/lib/config.libsonnet
@@ -99,14 +99,24 @@
       for c in std.objectFields($._config.alertmanager_clusters)
     },
 
-    // Shortcut to alertmanager_clusters entry for this cluster.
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/lib/config.libsonnet
```diff
@@ -99,14 +99,24 @@
       for c in std.objectFields($._config.alertmanager_clusters)
     },
 
-    // Shortcut to alertmanager_clusters entry for this cluster.
+    // Backwards compatible base entry $.alertmanager_cluster_self
```
malcolmholmes, line 102: this comment no longer makes sense here?

malcolmholmes (approved):

prometheus-ksonnet/lib/config.libsonnet
Duologic: how does it not make sense?

PR #476, opened 2021-02-25
Add pvcVolumeMount function

malcolmholmes:
We have `configVolumeMount`, `secretVolumeMount` etc, but lack the ability to mount a PVC!

Whilst mostly we will do this with a statefulset, and that is why we haven't seen this need so readily, there are usecases where we might want to mount a PVC manually, e.g. a read/write one to multiple pods.

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -155,6 +155,21 @@ local util(k) = {
       volume.fromHostPath(name, hostPath),
     ]),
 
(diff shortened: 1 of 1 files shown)


Duologic:
Maybe we can include that in k8s-alpha under custom/apps.v1....?

Duologic (approved):

malcolmholmes:
you mean adding all of the mount functions, not just this one, right?

Duologic:
Sure, but adding it here first to test/harden it is fine too.

jdbaldry (approved):
LGTM

PR #477, opened 2021-02-25
Promsonnet extra patching

malcolmholmes:
I'm finally using Promsonnet in anger.

What I've been thinking is that while we're not gonna succeed at getting the majority of mixin authors to refactor their mixins to use promsonnet (with its hidden maps), we still want to be able to patch rules.

Therefore, in this PR I've left the original `patchRule` code in place, but added code that detects which mode, and allows us to patch old-style rules.

I also added a `v1.mixin.patchRule()` function that will patch a specific mixin.

### DIFF
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -6,6 +6,10 @@ rules internally as a map, which allows users to easily patch alerts and
 recording rules after the fact. In contrast, lists require complex
 iteration logic.
 
+It also provides useful functions for patching existing rules and rule
+groups, along with patching these embedded in [monitoring mixins](https://github.com/monitoring-mixins/docs).
+
+## Creating Rule Groups with PromSonnet
 Take this example:
 
 ```
@@ -55,8 +59,9 @@ simply with code such as:
 
 We no longer need to iterate over all alerts to do so.
 
-Better than this though, `PromSonnet` provides a helper function, making
-this trivial:
+## Patching Rules
+The power of Jsonnet is its ability to patch existing resources. PromSonnet provides support for
+patching existing rule groups, e.g.:
 ```
 prometheusAlerts+: prom.v1.patchRule('prometheus_metamon', 'PrometheusDown', { 'for': '10m' }),
 ```
@@ -68,10 +73,39 @@ $ jsonnet example.jsonnet
 $ # or:
 $ jsonnet patch.jsonnet
 ```
+> NOTE: This `v1.patchRule` function is intelligent: it can work with PromSonnet created rule groups
+as well as ones that were created manually (without the `groups_map` for example).
 
+## Working with Mixins
 Given that we often want to patch existing rules, and those often exist within mixins,
 we can also do:
 ```
 mixins+: prom.v1.mixin.patchRule('base', 'prometheus_metamon', 'PrometheusDown', { 'for': '10m' }),
 ```
 This will apply the same patch to the `base` mixin.
+
+## 'Upgrading' existing rule groups and mixins
+PromSonnet provides a `v1.legacy` API which includes functions for 'upgrading' rule groups without the
+`groups_map` element into ones that do. This can make manual patching easier. It doesn't make any
+significant difference if using the PromSonnet patch functions. When working with mixins, there are
+two available approaches:
+```
+local mixin = 'my-mixin.libsonnet';
+{
+  mixins+:: {
+    mymixin: mixin + prom.v1.legacy.mixin.update,
+  }
+}
+
+```
+Or, if the mixin has already been added to your `$.mixins`, perhaps by code you don't control:
+```
+local mixin = 'my-mixin.libsonnet';
+{
+  mixins+:: {
+    mymixin: mixin,
+  }
+} + {
+  mixins+:: prom.v1.legacy.mixin.updateExisting('mymixin'),
+}
+```
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -1,4 +1,4 @@
-#PromSonnet
+# PromSonnet
 
 `PromSonnet` is intended as a very simple library for creating Prometheus
 alerts and rules. It is 'patching friendly', as in, it maintains the
--- a/promsonnet/README.md
+++ b/promsonnet/README.md
@@ -68,3 +68,10 @@ $ jsonnet example.jsonnet
 $ # or:
 $ jsonnet patch.jsonnet
 ```
+
+Given that we often want to patch existing rules, and those often exist within mixins,
+we can also do:
+```
+mixins+: prom.v1.mixin.patchRule('base', 'prometheus_metamon', 'PrometheusDown', { 'for': '10m' }),
+```
+This will apply the same patch to the `base` mixin.
--- a/promsonnet/legacy.libsonnet
+++ b/promsonnet/legacy.libsonnet
@@ -21,7 +21,8 @@
                         if g.name == group
                         then g {
                           rules: std.map(function(r)
-                                           if ('alert' in r && r.alert == rule) || ('record' in r && r.record == rule)
+                                           if ('alert' in r && r.alert == rule) ||
+                                              ('record' in r && r.record == rule)
                                            then r + patch
                                            else r,
                                          g.rules),
--- a/promsonnet/legacy.libsonnet
+++ b/promsonnet/legacy.libsonnet
@@ -1,0 +1,56 @@
+{
+  // In this file, the word 'update' refers to retrofitting maps and _order
+  // lists to data structures that lack them.
+
+  v1: {
+    local v1 = self,
+    ruleGroupSet: {
+      update():: {
+        local groups = super.groups,
+        local this = self,
+        local groupsAsMap(group, acc) = acc { [group.name]+: group + v1.ruleGroup.update() },
+        local groupsAsList(group, acc) = acc + [group.name],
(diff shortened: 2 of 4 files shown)


beorn7 (approved):

beorn7:
WRT adopting promsonnet: Have we already tried? I'm the de-facto maintainer of a bunch of mixins in the Prometheus GH org, so I'd definitely help you to test the waters.

malcolmholmes:
thanks, that's useful. The question is: do we stick with what we've got (lists) and use "ugly" `std.map` calls that are hard to read but work everywhere, or do we go with the map method where the library is easier to read but it requires mixin authors to switch to the new lib?

Hmm, if we had a `parseOldStyleRules()` method, we'd be sorted: then we could do `parseOldStyleRules($.prometheusAlerts) + prom.v1.patchRule(...)` and hence the new lib works with old stuff. **That** is the way to go. I'll fix this PR to do that.

Then we can discuss whether it is useful to migrate some mixins to fully rely upon promsonnet.

beorn7:
Yes, being able to work with the old stuff while (potentially) migrating the old stuff gradually to the new world order would be ideal.

malcolmholmes:
This updates `promsonnet` such that, before a mixin can use `patchRule`, it must be "updated". This adds the necessary internal data structures to make patching easier. Here's an example where first the `base` mixin is updated, then two of its rules are patched:

```    mixins+:: promsonnet.v1.mixin.updateMixin('base')
              + promsonnet.v1.mixin.patchRule('base', 'prometheus-extra', 'PromScrapeFailed', {
                expr: 'up{instance!~"myhost-.*"}!=1',
              })
              + promsonnet.v1.mixin.patchRule('base', 'prometheus-extra', 'PromScrapeFlapping', {
                expr: 'avg_over_time(up{instance!~"myhost-.*"}[5m]) < 1',
              }),
```

I'm noticing that similar functions for `scrape_config` could be really useful too.

beorn7:
OK, let's try it out in production? (o:

beorn7:
I mean this PR, not necessarily the scrape_config thing. We can do that once we have gathered experience with this PR.

malcolmholmes:
@Duologic and I discussed this further. Have another proposal to work on. Basically, something like this:
```
local promsonnet = import 'promsonnet/prom.libsonnet';
local oldStyleMixin = import 'old-style-mixin.libsonnet';
local promsonnetBasedMixin = import 'promsonnet-based-mixin.libsonnet';
{
    mixins+:: 
              oldStyleMixin
            + promsonnet.v1.mixin.legacy.patchRule('base', 'prometheus-extra', 'PromScrapeFailed', {
                expr: 'up{instance!~"myhost-.*"}!=1',
              })
            + promsonnetBasedMixin
            + promsonnet.v1.mixin.patchRule('base', 'prometheus-extra', 'PromScrapeFailed', {
                expr: 'up{instance!~"myhost-.*"}!=1',
              }),
```
So rather than forcing users of mixins to "upgrade", or having one `patchRule` function that has to decide which type it is, we would have two APIs, one for promsonnet based mixins, and one for 'legacy' list based ones.

I guess we need to choose one of these three approaches. Thoughts @beorn7 ?

Duologic:
I think we should not put the distinction burden on the consumer, this should be handled within promsonnet. We introduce the Map-style in promsonnet so we can start using it in our libraries, the outcome of promsonnet is and should still be an array so other mixin implementations keep working. The patchRule/patchAlert functions are useful outside the Map-style idea so they should operate on the Array. I propose the following interface: (as discussed in DM)

```
local someMixin = import 'some-mixin.libsonnet';
local promsonnet = import 'promsonnet/prom.libsonnet';
{
  mixins+:: {
    base+:
      someMixin
      + promsonnet.util.patchAlert(
        ruleGroup='prometheus-extra',
        alertName='PromScrapeFailed',
        patch={
          expr: 'up{instance!~"myhost-.*"}!=1',
        }
      ),
  },
}
```

Duologic (commented):
I think we should not put the distinction burden on the consumer, this should be handled within promsonnet. We introduce the Map-style in promsonnet so we can start using it in our libraries, the outcome of promsonnet is and should still be an array so other mixin implementations keep working. The patchRule/patchAlert functions are useful outside the Map-style idea so they should operate on the Array. I propose the following interface: (as discussed in DM)

```
local someMixin = import 'some-mixin.libsonnet';
local promsonnet = import 'promsonnet/prom.libsonnet';
{
  mixins+:: {
    base+:
      someMixin
      + promsonnet.util.patchAlert(
        ruleGroup='prometheus-extra',
        alertName='PromScrapeFailed',
        patch={
          expr: 'up{instance!~"myhost-.*"}!=1',
        }
      ),
  },
}
```

malcolmholmes:
I did some testing. I wanted to show that Jsonnet is much faster at patching maps than it is at patching lists, as that is a clear justification for using the map approach.

Unfortunately, the evidence did not back me up.

I loaded in the `prometheus-mixin`, and 'updgraded' its alerts. Then I patched the list, and then I patched the alerts. I did this 30 or 100 times. Rendering without patching took 14s. Patching with maps or lists took approx 27s, regardless of method, which is totally disappointing. Leaves me temporarily baffled about how to proceed.

beorn7:
I agree with @Duologic that this should ideally be transparent to the user.

I don't think that configuration rendering has to be super efficient in terms of runtime. As long as it finishes in acceptable time in real-world scenarios, it's fine.

malcolmholmes:
I've just pushed what I hope is a last iteration. This one has a unified way into the `mixin.patchRules` function, puts the legacy and new functionality into separate files for clarity, and also keeps the `update()` functionality should someone want to use it, but it is nicely kept in the `legacy.libsonnet` file.

malcolmholmes:
And the justification for using PromSonnet to create your rules is clearly not performance, but rather clarity of structure and simplicity and readability of code.

beorn7 (approved):
From what I can see, this looks good. @Duologic might have stronger opinions. (I have to admit that I'm currently not as deeply into jsonnet as I would like to be.)

promsonnet/README.md
```diff
@@ -68,3 +73,39 @@ $ jsonnet example.jsonnet
...
+
+## Working with Mixins
+Given that we often want to patch existing rules, and those often exist within mixins,
+we can also do:
+```
+mixins+: prom.v1.mixin.patchRule('base', 'prometheus_metamon', 'PrometheusDown', { 'for': '10m' }),
+```
+This will apply the same patch to the `base` mixin.
+
+## 'Upgrading' existing rule groups and mixins
```
Duologic, line 87: The maps are useful, totally agree that it is the way forward, but I would see it as a layer on top of the current array model instead of trying to upgrade the array model to match the layer on top. This upgrade process follows a weird process of  `array -> map -> array` because eventually Prometheus consumes an array.

Maybe the promsonnet should have a `asLegacyMixin()` function that converts the internal maps to arrays for use with the current mixin model? That way the process works in one direction `maps -> array`.

promsonnet/legacy.libsonnet
```diff
@@ -0,0 +1,56 @@
...
+                        then g {
+                          rules: std.map(function(r)
+                                           if ('alert' in r && r.alert == rule) || ('record' in r && r.record == rule)
+                                           then r + patch
+                                           else r,
+                                         g.rules),
+                        }
+                        else g,
+                      super.groups),
+    },
```
Duologic, line 31: Add a few more newlines 
```suggestion
    patchRule(group, rule, patch):: {
      groups: std.map(
        function(g)
          if g.name == group
          then g {
            rules: std.map(
              function(r)
                if ('alert' in r && r.alert == rule) ||
                   ('record' in r && r.record == rule)
                then r + patch
                else r,
              g.rules
            ),
          }
          else g,
        super.groups
      ),
    },
```

promsonnet/README.md
Duologic: Or maybe the `update()` just needs to be called inside patchRule and other functions where relevant? Taking the burden and complexity away from the consumer.

If so, there is even more reason not to 'update' as the `array -> map -> array` approach is now part of the promsonnet code, making it even harder to reason about.

Duologic (commented):
It looks to me the readme is not 100% in line with the intention of the code, is that right?

promsonnet/README.md
```diff
@@ -68,3 +73,39 @@ $ jsonnet example.jsonnet
...
 $ jsonnet patch.jsonnet
 ```
+> NOTE: This `v1.patchRule` function is intelligent: it can work with PromSonnet created rule groups
+as well as ones that were created manually (without the `groups_map` for example).
+
+## Working with Mixins
+Given that we often want to patch existing rules, and those often exist within mixins,
+we can also do:
+```
+mixins+: prom.v1.mixin.patchRule('base', 'prometheus_metamon', 'PrometheusDown', { 'for': '10m' }),
```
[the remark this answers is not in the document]
malcolmholmes: yes, that's fair. I made that change with the `update` functionality - primary is mixin into the mixin, secondary is 'patch the mixin in $.mixins'

promsonnet/README.md
malcolmholmes: there's no benefit to upgrading for the sake of `patchRule`, turns out it does not have a performance benefit. So the benefit to upgrade is just that a user might want to be able to patch rules manually, and this will allow them to do it. I don't expect it to be a highly used feature, but who knows.

Duologic (approved):

PR #478, opened 2021-02-25
Configure default AM

malcolmholmes:
By default, `prometheus-ksonnet` now removes any alert managers that were previously configured. This change prevents that.


Duologic (changes requested):
There is conditional logic in `alertmanager_cluster_self` that will get overridden by this change.

beorn7:
@Duologic @malcolmholmes have you figured out how to reconcile those two things?

Duologic (commented):
I think we have to fix line 117:

```diff
- else if std.length(self.alertmanagers) == 0 then
+ else if std.length(std.objectFields(self.alertmanagers)) == 0 then
```

PR #479, opened 2021-02-26
Add missing datasource template

hoenn:
Other template vars failing to load since they rely on `DS_PROMETHEUS`
image

Working change example:
image


jeschkies (approved):
Thanks a ton!

PR #481, opened 2021-02-27
Nginx directory separate html

malcolmholmes:
In the current nginx-directory, the HTML and config are intimately entwined. This PR separates them out into their own jsonnet files and their own configmaps. Functionally this is a no-op, in terms of actual diffs, we should just see the HTML move into its own configmap.

### DIFF
--- a/nginx-directory/html.libsonnet
+++ b/nginx-directory/html.libsonnet
@@ -1,16 +1,11 @@
-local kausal = import 'ksonnet-util/kausal.libsonnet';
+local k = import 'k.libsonnet';
+local configMap = k.core.v1.configMap;
 {
-  local _config = self._config,
-  local k = kausal { _config+:: _config },
-
-  local configMap = k.core.v1.configMap,
 
   nginx_html_config_map:
     local vars = {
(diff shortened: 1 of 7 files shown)


Duologic (approved):

nginx-directory/html.libsonnet
```diff
@@ -0,0 +1,18 @@
+local k = import 'k.libsonnet';
```
jvrplmlmn, line 1: Late to this one, a question: Can we use `ksonnet-util/kausal.libsonnet` and `k.libsonnet` interchangeably?

I noticed that `nginx-directory/configfile.libsonnet` uses `ksonnet-util/kausal.libsonnet`.

malcolmholmes: think of kausal.libsonnet as k.libsonnet + utils. If you are just using raw k8s, just use k.libsonnet.

malcolmholmes: Although, there are some opinions that we might loose there - e.g. default labels, etc, which is worth thinking about.

PR #483, opened 2021-03-03
feat: Add enterprise-metrics Jsonnet library

jdbaldry:

### DIFF
--- a/enterprise-metrics/main.libsonnet
+++ b/enterprise-metrics/main.libsonnet
@@ -1,0 +1,379 @@
+local cortex =
+  (import 'github.com/grafana/cortex-jsonnet/cortex/cortex.libsonnet')
+  + (import 'github.com/grafana/cortex-jsonnet/cortex/gossip.libsonnet')
+  + (import 'github.com/grafana/cortex-jsonnet/cortex/tsdb.libsonnet')
+  + {
+    _config+:: {
+      // Remove the blocksStorageConfig from Cortex as this library provides its own config interface.
+      blocksStorageConfig:: {},
+      // No resources should have a pre-configured namespace. These should be removed with removeNamespaceReferences.
+      // TODO: Remove upstream.
+      namespace:: 'namespace',
+      ringConfig+: {
+        // Hide the consul hostname because it isn't used with memberlist.
+        'consul.hostname':: '',
+      },
+      memcached_chunks_enabled: true,
+      memcached_index_queries_enabled: true,
+      memcached_metadata_enabled: true,
+    },
+    distributor_args+:: {
+      // Disable the ha-tracker by default.
+      'distributor.ha-tracker.enable':: null,
+      'distributor.ha-tracker.enable-for-all-users':: null,
+      'distributor.ha-tracker.store':: null,
+      'distributor.ha-tracker.etcd.endpoints':: null,
+      'distributor.ha-tracker.prefix':: null,
+      // Memberlist gossip is used instead of consul for the ring.
+      'distributor.ring.consul.hostname':: null,
+    },
+  };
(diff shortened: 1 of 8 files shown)


enterprise-metrics/main.libsonnet
```diff
@@ -0,0 +1,379 @@
...
+      + roleBinding.roleRef.withApiGroup('rbac.authorization.k8s.io')
+      + roleBinding.roleRef.withKind('Role')
+      + roleBinding.roleRef.withName(target)
+      + roleBinding.withSubjects([
+        subject.new()
+        + subject.withName(target)
+        + subject.withKind('ServiceAccount'),
+      ]),
+  },
+}
```
malcolmholmes, line 379: this file is large, would it make sense to split it for each tool?

enterprise-metrics/main.libsonnet
```diff
@@ -0,0 +1,379 @@
+local cortex =
```
malcolmholmes, line 1: would anyone ever want to override this `cortex`? If so, would it be better as a hidden?

enterprise-metrics/main_test.jsonnet
```diff
@@ -0,0 +1,53 @@
...
+    std.foldr(function(module, acc) acc && assertNoNamespaceInArgs(module), modules, true),
+
+  testNoVisibleDocsonnetFields:
+    local isDocsonnetField(field) = std.startsWith(field, '#');
+    local docsonnetFields(obj) = std.filter(isDocsonnetField, std.objectFields(obj))
+                                 + std.flatMap(function(field) docsonnetFields(obj[field]),
+                                               std.filter(function(field) std.isObject(obj[field]), std.objectFieldsAll(obj)));
+    local fields = docsonnetFields(l);
+    assert std.length(fields) == 0 : 'docsonnet fields should not be visible, fields=%s' % std.toString(fields); true,
+}
```
malcolmholmes, line 53: wow, tests!

malcolmholmes (commented):
Looking great!

enterprise-metrics/main.libsonnet
jdbaldry: That's a really good question and one I have been struggling with a lot through the course of writing this library.

While `cortex` is a local variable, I can completely control the user interface. For instance, all modules are represented by a consistent `{ args, container, [job || deployment || statefulSet], service }` where by their nature, hidden fields function as a more flexible function input and the visible fields become output. Sure, a user can do something nonsensical like `{ args:: 'this is now a string' }` and it would break the functioning of the rest of that module but I would imagine that it would be fairly easy for a user to understand the resulting error.

If I make`cortex` a hidden field, I then expose the entire `cortex` interface in my library which gives users great flexibility but at a huge overhead in complexity to debug. Admittedly the Jsonnet errors are getting better so that this isn't quite so difficult to decipher any more.

Right now I'm leaning towards keeping this private and reserving the option to expose `cortex` in the future.

enterprise-metrics/main_test.jsonnet
jdbaldry: Not the neatest but they have helped me out a bit already. I want to explore using https://github.com/yugui/jsonnetunit/tree/master/jsonnetunit at some point.

enterprise-metrics/main.libsonnet
jdbaldry: It certainly is. I'd like to work on splitting this out into smaller files but I am of the opinion that each Jsonnet file should work as a unit to help users understand the library while tooling is still relatively weak. I'm not sure it's good for files to be reaching for configuration outside of its own file which makes sharing common configuration awkward across multiple files.

PR #485, opened 2021-03-05
Allow for custom nginx directives as array of strings

dannykopping:
I've run into a specific scenario, and none of the available options solve the problem for me.

I need to add the following lines:
```nginx
sub_filter "./" "/app/";
sub_filter_once off;
```

I would use it like this:

```jsonnet
{ title: 'App', path: 'app', url: 'http://app.default.svc.cluster.local:80/', custom: [
      'sub_filter "./" "/app/";',
      'sub_filter_once off;',
    ] }
```


malcolmholmes (approved):
Looks neat.

PR #488, opened 2021-03-08
refactor(prometheus): generic HA setup

Duologic:
Reduce code duplication by iterating over `replicas`. As a side effect this also allows for more than 2 replicas.

This is generally a no-op, only the `__replica__` external label is now using `_config.name` to prevent conflict when
running multiple prometheis in the same cluster with remote write enabled.

### DIFF
--- a/prometheus/ha-mixin.libsonnet
+++ b/prometheus/ha-mixin.libsonnet
@@ -40,17 +40,15 @@ function(replicas=2) {
     function(i, acc)
       local name = _config.name + '-' + i;
       local config =
-        this._config.prometheus_config {
-          prometheus+: {
-            global+: {
-              external_labels+: {
-                __replica__: name,
-              },
+        this.prometheus_config {
+          global+: {
+            external_labels+: {
+              __replica__: name,
             },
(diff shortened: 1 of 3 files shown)


prometheus/ha-mixin.libsonnet
```diff
@@ -35,67 +35,48 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
...
+                'prometheus.yml': k.util.manifestYaml(config),
+              }),
+        alerts: configMap.new('%s-alerts' % name) +
+                configMap.withData({
+                  'alerts.rules': k.util.manifestYaml(this.prometheusAlerts),
+                }),
+        rules: configMap.new('%s-recording' % name) +
+               configMap.withData({
+                 'recording.rules': k.util.manifestYaml(this.prometheusRules),
+               }),
```
Duologic, line 65: We could take these out of the iteration and just provide the same rules/alerts to all replicas, but I didn't want to trigger a major diff.

malcolmholmes (approved):
Looks good

PR #489, opened 2021-03-08
Introducing "weight" field to allow for arbitrary location block ordering

dannykopping:
These routes do not need to be sorted, unless i'm missing something @malcolmholmes?

This sorting currently interferes with a definition which relies on precise ordering of `location` blocks.

### DIFF
--- a/nginx-directory/configfile.libsonnet
+++ b/nginx-directory/configfile.libsonnet
@@ -52,7 +52,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
     local vars = {
       location_stanzas: [
         buildLocation(service)
-        for service in $._config.admin_services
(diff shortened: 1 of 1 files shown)


malcolmholmes (commented):
I would say this aims to uniquify rather than sort. What problem does this cause you?

dannykopping:
The documentation says that `std.set` is an alias of `std.uniq(std.sort(var))`
The order of `location` blocks in nginx is significant. I need to override behaviour in one part of an application.

dannykopping:
@malcolmholmes I've added a `weight` field to sort the blocks (the higher the weight, the lower down the block is pushed), and restored the functionality to `uniq` all services by their `url` field.

malcolmholmes (approved):

PR #490, opened 2021-03-09
Not all Prometheus rules are alerts.

tomwilkie:
This allows us to put recording rules in the same rule groups as alerts, which is useful when the alerts depend on the recording rules.


jtlisi (approved):
LGTM

PR #495, opened 2021-03-15
feat(consul): Add consul-sidekick only when consul is run in distributed fashion.

kavirajk:
feat(consul): Add consul-sidekick only when consul is run in distributed fashion.

Rationale:
1. We don't need consul-sidekick if there is only one replica of consul running.
2. consul-sidekick is depricated and not being maintained anymore.

Currently Loki runs single consul. And we face some issue with consul-sidekick. So we need a way to vendor consul without sidekick.

### DIFF
--- a/consul/consul.libsonnet
+++ b/consul/consul.libsonnet
@@ -296,15 +296,27 @@ k {
   local deployment = $.apps.v1.deployment,
 
   consul_deployment:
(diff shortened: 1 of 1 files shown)


kavirajk:
Fixes  https://github.com/grafana/loki/issues/2444

PR #497, opened 2021-03-15
refactor(prometheus): shard mixins over multiple configmaps

Duologic:
We are hitting bug https://github.com/kubernetes/kubernetes/issues/85840

Smaller configmaps should workaround that. This PR will break a few backwards compat assumptions.

### DIFF
--- a/mixin-utils/utils.libsonnet
+++ b/mixin-utils/utils.libsonnet
@@ -161,7 +161,7 @@ local g = import 'grafana-builder/grafana.libsonnet';
         group {
           rules: std.map(function(rule)
             rule +
-            if std.objectHas(overrides, rule.alert)
+            if 'alert' in rule && std.objectHas(overrides, rule.alert)
             then overrides[rule.alert]
             else {}, super.rules),
         }, super.groups),
--- a/mixin-utils/utils.libsonnet
+++ b/mixin-utils/utils.libsonnet
@@ -147,14 +147,24 @@ local g = import 'grafana-builder/grafana.libsonnet';
     groups: std.prune(std.map(removeRuleGroup, currentRuleGroups)),
   },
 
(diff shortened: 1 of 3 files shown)


mixin-utils/utils.libsonnet
```diff
@@ -140,4 +140,31 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
+    prometheusAlerts+:: {
+      groups: std.map(function(group)
+        group {
+          rules: std.map(function(rule)
+            rule +
+            if 'alert' in rule && std.objectHas(overrides, rule.alert)
+            then overrides[rule.alert]
+            else {}, super.rules),
+        }, super.groups),
+    },
```
tomwilkie, line 168: Nit: might be worth breaking each of the functions out into a local.

prometheus/mixins.libsonnet
```diff
@@ -52,27 +57,76 @@
...
-  local emptyMixin = {
-    prometheusAlerts+:: {},
-    prometheusRules+:: {},
+  withMixinsConfigmaps(mixins):: {
+    local configMap = k.core.v1.configMap,
+    prometheus_config_maps_mixins+: [
+      local mixin = mixins[mixinName] {
+        prometheusAlerts+:: {},
+        prometheusRules+:: {},
+      };
```
tomwilkie, line 66: Nit: some repetition here, might consider breaking this out into a local "emptyMixin".

tomwilkie (approved):

beorn7:
Thanks for nominating me as a reviewer. But this is clearly above my current level of Jsonnet mastery. I see that @tomwilkie has already reviewed, and this is certainly very relevant for @malcolmholmes. I don't think it makes sense right now to invest the time I would need to fully comprehend this.

malcolmholmes (approved):

PR #498, opened 2021-03-15
fix: Change composition of cortex and common arguments

### LINKED ISSUE #496: enterprise-metrics: `$._config.commonArgs` confusing configuration
All module arguments inherit and override `$._config.commonArgs`. In order to benefit from the cortex-jsonnet best practices, Cortex modules (modules that aren't specific to GEM) inherit their arguments from the vendored library. Since the inheritance is `$._config.commonArgs + cortex.<module>_args`, configuration made in `$_config.commonArgs` may not appear in the manifestation of Cortex modules.

As a user I would expect the inheritance to behave more like:
`cortex.module_args + $._config.commonArgs + $.<module>.args`

jdbaldry:
Resolves #496 


jdbaldry:
Closing in favor of #499 because `jsonnet-bundler` doesn't like forward slashes in branch names.

PR #502, opened 2021-03-16
fix: Ensure gossip-ring Service is created

jdbaldry:


jdbaldry:
I'm not sure why the docsonnet args have changed...

PR #506, opened 2021-03-17
Enable netstat metrics again.

simonswine:
Netstat metrics are fairly useful when it comes to debugging ingress issues and cascading failures. I couldn't find a good reason why they were disable (cf. https://github.com/grafana/jsonnet-libs/commit/755b20239c0ca235744421744d4c3c5c99225746) and think those metrics are not adding cardinality too excessively.

### DIFF
--- a/prometheus-ksonnet/lib/node-exporter.libsonnet
+++ b/prometheus-ksonnet/lib/node-exporter.libsonnet
@@ -8,7 +8,6 @@
       '--path.procfs=/host/proc',
       '--path.sysfs=/host/sys',
 
-      '--no-collector.netstat',
       '--collector.netdev.ignored-devices=^veth.+$',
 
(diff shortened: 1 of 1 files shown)


Duologic:
@tomwilkie Care to shed your light on why that was the case? Maybe too verbose? High cardinality?

tomwilkie:
Yeah I don't recall.  Can we test this in eg ops-tools and see what the impact is?

Duologic (approved):
LGTM considering the internal tests run fine.

simonswine:
I have deployed the change to "ops-tools" earlier. And it think the cardinality impact is relatively minimal (34 metrics per node)
scrn-2021-03-23-17-57-56

tomwilkie (approved):

PR #507, opened 2021-03-17
Update node exporter.

simonswine:
This updates node exporter to v1.1.2. There are breaking changes in this update: 

https://github.com/prometheus/node_exporter/blob/v1.0.0/CHANGELOG.md#breaking-changes

Would appreciate any feedback, if this is feasible


Duologic:
Can we have a look at updating the node-mixin too? The rules/alerts/dashboards there probably reflect those changes too.

simonswine:
> Can we have a look at updating the node-mixin too? The rules/alerts/dashboards there probably reflect those changes too.

@Duologic I am acutally not too sure how to do that, it seems like there is "master" pinned in the jsonnetfile and no lock is in the repo?

Duologic:
Ah, good point, we can do that internally.

Duologic (approved):

PR #512, opened 2021-03-19
fix(prometheus): only split on config, not rules

Duologic:
There is not real reason to create seperate configmaps for the rules/alerts, so ha-mixin does not need to split those
into a configmap per replica. The only difference would be on the config level, which we can do on a file level instead
of directory.

Essentially this solves the problem that the watch container didn't see the configmap changes from the mixins. In
general it also reduces the complexity of the code.

### DIFF
--- a/prometheus/ha-mixin.libsonnet
+++ b/prometheus/ha-mixin.libsonnet
@@ -18,7 +18,7 @@ function(replicas=2) {
   ),
 
   _config+:: {
-    prometheus_config_dir: '/etc/$(POD_NAME)',
+    prometheus_config_file: _config.prometheus_config_dir + '/$(POD_NAME).yml',
   },
 
   // The '__replica__' label is used by Cortex for deduplication.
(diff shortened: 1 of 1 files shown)


prometheus/ha-mixin.libsonnet
```diff
@@ -36,46 +36,37 @@ function(replicas=2) {
...
-    std.foldl(
-      function(mounts, obj)
-        mounts
-        + k.util.configMapVolumeMount(obj.base, '/etc/%s' % obj.name)
-        + k.util.configMapVolumeMount(obj.alerts, '/etc/%s/alerts' % obj.name)
-        + k.util.configMapVolumeMount(obj.rules, '/etc/%s/recording' % obj.name),
-      self.prometheus_config_maps,
-      {},
-    )
-  ,
```
malcolmholmes, line 78: How come you are removing this? I don't see changes to how you mount these configmaps?

Duologic: `prometheus_config_mount` does not change anymore, it is the same as in `prometheus.libsonnet`.

malcolmholmes: So this isn't *adding* a `prometheus_config_mount` element, it is overriding an existing one. All you're doing here is removing an override, which makes total sense.

malcolmholmes (approved):

PR #530, opened 2021-03-25
fix(prometheus): don't generate empty files

Duologic:
Even though it is not a problem for prometheus to receive empty rules files, it is much cleaner not to generate them.


malcolmholmes (approved):
Looks good, useful feature, and makes the code easier to read (with the `mixin_data` element with questions already answered).

PR #532, opened 2021-03-25
fix(prometheus): don't mount if not exists

Duologic:
Missed this in the main prometheus.libsonnet, we should not create the configmaps there either, nor should we mount if
they don't exist.

### DIFF
--- a/prometheus/prometheus.libsonnet
+++ b/prometheus/prometheus.libsonnet
@@ -5,6 +5,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
 + (import 'images.libsonnet')
 + (import 'mixins.libsonnet')
 + {
+  local this = self,
   local _config = self._config,
   local k = kausal {
     _config+:: _config,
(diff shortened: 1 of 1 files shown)


prometheus/prometheus.libsonnet
```diff
@@ -38,17 +39,26 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
...
-      }),
-
-      configMap.new('%s-recording' % _config.name)
-      + configMap.withData({
-        'recording.rules': k.util.manifestYaml(prometheusRules),
-      }),
-    ],
+    ]
+    + (
+      if std.objectHas(this, 'prometheusAlerts') && std.prune(this.prometheusAlerts) != {}
```
mplzik, line 44: nit: this code repeats few times, it _might_ live better alone in a separate function (or not, no strong opinion here).

mplzik (approved):
LGTM

prometheus/prometheus.libsonnet
Duologic: Agreed, I want to consolidate this further in the near future.

PR #533, opened 2021-03-25
fix(prometheus): create legacy endpoints instead of checking if they exist

Duologic:
std.objectHas didn't find the hidden elements, causing the configmaps to be empty and removed. this should restore that.

### DIFF
--- a/prometheus/prometheus.libsonnet
+++ b/prometheus/prometheus.libsonnet
@@ -24,6 +24,9 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
     else {}
   ),
 
+  prometheusAlerts+:: {},
+  prometheusRules+:: {},
+
   withHighAvailability(replicas=2):: ha_mixin(replicas),
 
   local configMap = k.core.v1.configMap,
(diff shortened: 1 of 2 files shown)


prometheus/prometheus.libsonnet
```diff
@@ -41,7 +44,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
       }),
     ]
     + (
-      if std.objectHas(this, 'prometheusAlerts') && std.prune(this.prometheusAlerts) != {}
+      if std.prune(this.prometheusAlerts) != {}
```
julienduchesne, line 47: Shouldn't you prune while defining your local above (l38)? You prune here while checking for an empty definition but you don't while creating your yaml. Isn't that inconsistent? How often do these maps have empty members?

julienduchesne (approved):

mplzik (approved):
LGTM

prometheus/prometheus.libsonnet
Duologic: No, empty members sometimes have meaning, I have fallen in this trap before. :grimacing:

julienduchesne: Why prune at all then?

PR #539, opened 2021-04-07
Add Istio-mixin

chaudyg:
Add a Mixin to monitor an istio service mesh.

In this initial PR, we import the [official dashboard](https://istio.io/latest/docs/tasks/observability/metrics/using-istio-dashboard/) maintained by the istio community.

Next steps involve writing prometheus recording-rules to aggregate istio_metrics and alerting rules.


istio-mixin/jsonnetfile.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version": 1,
+  "dependencies": [
+    {
+      "source": {
+        "git": {
+          "remote": "https://github.com/grafana/jsonnet-libs.git",
+          "subdir": "grafana-builder"
```
trotttrotttrott, line 8: Where is this being used?

chaudyg: Oups. Good catch. Removed.

PR #544, opened 2021-04-17
Rename legacy vendor name for rfrail3

malcolmholmes:
Prometheus Ksonnet imports a dashboard from an rfrail3 github repo. Except, jsonnet-bundler
puts a symink into `/vendor/grafana-dashboards`, which is misleading. This fix fixes the
link to be `rfrail3-grafana-dashboards` which is much more informative.

### DIFF
--- a/prometheus-ksonnet/jsonnetfile.json
+++ b/prometheus-ksonnet/jsonnetfile.json
@@ -98,7 +98,7 @@
           "subdir": ""
         }
       },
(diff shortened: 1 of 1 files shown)


julienduchesne (commented):
hmm, not sure I understand. This isn't valid JSON. Missing a key here?

julienduchesne (approved):

jvrplmlmn (approved):

PR #545, opened 2021-04-21
fix: Only enable enterprise authentication on the gateway

jdbaldry:
Some problems are evident when this is set for all
components. Notably, the query-frontend strips authentication from
query_range requests which affects range queries sent to the queriers.


MichelHollands (approved):
LGTM

PR #551, opened 2021-04-28
Add missing kube-state-metrics cluster role policies

nikosmeds:
Add missing `kube-state-metrics` policies from the example [cluster-role.yaml](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/cluster-role.yaml) to our ksonnet library.

These resolve permission errors that we currently see in `kube-state-metrics` pods.

### DIFF
--- a/prometheus-ksonnet/lib/kube-state-metrics.libsonnet
+++ b/prometheus-ksonnet/lib/kube-state-metrics.libsonnet
@@ -57,6 +57,11 @@
       policyRule.withVerbs(['list', 'watch']),
 
       policyRule.new() +
+      policyRule.withApiGroups(['authorization.k8s.io']) +
+      policyRule.withResources(['subjectaccessreviews']) +
+      policyRule.withVerbs(['create']),
+
+      policyRule.new() +
       policyRule.withApiGroups(['ingresses']) +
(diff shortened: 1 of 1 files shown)


trotttrotttrott (commented):
If I understand correctly, there are access errors relating to `mutatingwebhookconfigurations`, `validatingwebhookconfigurations`, `storageclasses`, `volumeattachments`, and `networkpolicies`.

Why do we also need additional policy around `subjectaccessreviews`, `ingresses`, and `leases`?

nikosmeds:
> Why do we also need additional policy around `subjectaccessreviews`, `ingresses`, and `leases`?

We don't currently see errors for these resources, however they are included in the example [cluster-role.yaml](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/cluster-role.yaml). I guess whether or not we include them depends on if we...

1. Want to stay 1:1 with the upstream example, or
2. Simply include the resources for which we currently see errors

I can remove these additional policies for now if that's preferred.

trotttrotttrott (approved):
> ...they are included in the example cluster-role.yaml...

Ah, sorry, I didn't look closely enough. I don't mind the additions, just didn't understand the motive. LGTM.

PR #552, opened 2021-04-30
feat: Update to use enterprise metrics v1.3.0

jdbaldry:


simonswine (approved):
LGTM

PR #558, opened 2021-05-18
refactor(nginx-directory)

Duologic:
- Consolidate the Kubernetes objects to the main file `directory.libsonnet`, reducing the places where to look for k8s
    objects and only needing to instantiate ksonnet-util once.
- Removing the squashing of multiple files together stops relying on objects to exist on the root level (with the
    exception of `config.libsonnet`).
- Add function to set a general name
- Add function to include the OAuth2 proxy
- Ensure default values are set on a service, reducing the need for if/else loops in a more jsonnet-y way.
- Move the read/send timeouts to the service level, allows for a more granular approach.

This library is missing some docs still, I see this out-of-scope for this PR.

This should be backwards compatible and generate a no-op. I've tested this on our internal clusters.


nginx-directory/config.libsonnet
```diff
@@ -1,14 +1,38 @@
...
+      //    path: 'prometheus',
+      //    url: 'http://prometheus.default.svc.cluster.local/prometheus/',
+      //
+      //    # Optional
+      //    params: '',
+      //    read_timeout: '60',
+      //    send_timeout: '60',
+      //    redirect: false,
+      //    allowWebsockets: false,
+      //    subfilter: false,
```
malcolmholmes, line 20: could usefully explain what `subfilter` does!

nginx-directory/configfile.libsonnet
```diff
@@ -1,67 +1,65 @@
...
-  servicesByWeight(services)::
+  local servicesByWeight(services) =
     std.sort(services, function(s) if std.objectHas(s, 'weight') then s.weight else 0),
+
+  location_stanzas: [
+    buildLocation(service)
+    for service in std.uniq(servicesByWeight(services), function(s) s.url)
+  ],
+  locations: std.join('\n', self.location_stanzas),
 }
```
malcolmholmes, line 65: now `configfile.libsonnet` generates a random `locations` thing, whereas before it generated a clear configMap. Not sure this bit is an improvement.

malcolmholmes (commented):
I'm not clear how these changes improve things. Can you elucidate the benefits?

nginx-directory/config.libsonnet
Duologic: Yes, someone should write docs. :innocent:

nginx-directory/configfile.libsonnet
Duologic: I think the Kubernetes object and Nginx config should be 2 separate things. If anything `configfile.libsonnet` should yield the `nginx.conf` file, what do you think? Maybe also do this for the html.

Duologic:
> Can you elucidate the benefits?

Updated the description accordingly.

malcolmholmes (approved):

PR #570, opened 2021-05-25
refactor(mysql_exporter): modular approach to passwords

Duologic:
This change is expected to be 100% backwards compatible. Most logic moved into functions in `main.libsonnet`. Notice the
use of the `withPassword` function in `mysql-exporter.libsonnet`.

I've included a few options we use downstream that will be available in mysqld_exporter v0.13.0, ~the rc already seems to support it~ v0.13.0 is now out, updated accordingly.

I've moved the new code into mysqld-exporter and added a deprecation notice to mysql-exporter, this should allow for a transition period.


bergquist:
> I wonder whether we should take the opportunity to rename this lib to mysqld_exporter for more consistency across the board feedback appreciated.

Would it make sense to have two libs during a transition period?

Duologic:
What would each encompass?

dafydd-t (approved):
Happy with the changes. Next time I make any jsonnet libs I'll attempt to take inspiration from this function-based approach.

Happy with the rename too. 

> What would each encompass?

The exact same thing, I imagine.

Duologic:
Ah, yis, misunderstood the question, will do.

Duologic:
> I've moved the new code into mysqld-exporter and added a deprecation notice to mysql-exporter, this should allow for a transition period.

PR #575, opened 2021-06-08
Docker Mixin

rgeyer:
This adds a docker mixin with a single dashboard overview, based on cadvisor metrics.


rgeyer:
It looks like this

image

PR #578, opened 2021-06-17
JIRA Mixin

stephaniehingtgen:
This PR adds an initial JIRA mixin, specifically for those running a hosted JIRA instance or their own JIRA servers. It uses the JIRA prometheus exporter that can be found on [github](https://github.com/AndreyVMarkelov/jira-prometheus-exporter) or in the [Atlassian marketplace](https://marketplace.atlassian.com/apps/1222502/prometheus-exporter-for-jira?hosting=server&tab=overview).

This Mixin includes:
A dashboard:
Dashboard1
Dashboard2

5 preconfigured alerts:
Alerts


jira-mixin/README.md
```diff
@@ -0,0 +1,38 @@
...
+On the Atlassian Marketplace, there is a [free version of the exporter](https://marketplace.atlassian.com/apps/1222502/prometheus-exporter-for-jira?hosting=server&tab=overview) if you are running your own JIRA server and a [pro version](https://marketplace.atlassian.com/apps/1217960/prometheus-exporter-pro-for-jira?tab=overview&hosting=datacenter) if you are running on a hosted JIRA server.
+
+## Alerts
+
+The Mixin includes 5 pre-configured alerts:
+- InstanceDown: Alerts when the JIRA server is down for more than 1 minute
+- LicenseExpired: Alerts when the JIRA license has expired
+- LicenseWarning: Warns when there are 7 days left until the JIRA license expires
+- NoUserCapacity: Alerts if there are no more user accounts available on the server
+- EmailErrorsHigh: Alerts if more than 1% of emails fail to send
```
hairyhenderson, line 20: markdown nitpick:

```suggestion
- `InstanceDown`: Alerts when the JIRA server is down for more than 1 minute
- `LicenseExpired`: Alerts when the JIRA license has expired
- `LicenseWarning`: Warns when there are 7 days left until the JIRA license expires
- `NoUserCapacity`: Alerts if there are no more user accounts available on the server
- `EmailErrorsHigh`: Alerts if more than 1% of emails fail to send
```

stephaniehingtgen: Good idea - I updated the README.

jira-mixin/alerts/general.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - alert: InstanceDown
+    expr: up == 0
+    for: 1m
+    labels:
+      severity: "critical"
+    annotations:
+      summary: "JIRA server down"
+      description: "The JIRA server has been down for more than 1 minute."
+  - alert: LicenseExpired
+    expr: jira_maintenance_expiry_days_gauge == 0
```
hairyhenderson, line 13: I wonder if this gauge might be negative if the license has expired in the past...

How about:

```suggestion
    expr: jira_maintenance_expiry_days_gauge <= 0
```

jira-mixin/alerts/general.yaml
```diff
@@ -0,0 +1,43 @@
+groups:
+- name: alert.rules
+  rules:
+  - alert: InstanceDown
+    expr: up == 0
```
hairyhenderson, line 5: Hmm... `up` is the synthetic metric that'll show whether the exporter is reachable, but probably more useful than that is to know whether the exporter itself can reach JIRA.

I've looking through the exporter code to see if there are any metrics that track that, but I can't find any obvious ones. I _think_ `jira_application_link_status_gauge` might be useful, but I can't figure out if that's relevant...

jira-mixin/alerts/general.yaml
```diff
@@ -0,0 +1,43 @@
...
+  - alert: LicenseExpired
+    expr: jira_maintenance_expiry_days_gauge == 0
+    for: 1m
+    labels:
+      severity: "critical"
+    annotations:
+      summary: "JIRA license expired"
+      description: "The JIRA license has expired."
+  - alert: LicenseWarning
+    expr: jira_maintenance_expiry_days_gauge == 7
```
hairyhenderson, line 21: I think this'll stop firing once it goes down to 6 days, so probably should use a `<=` here 😉

```suggestion
    expr: jira_maintenance_expiry_days_gauge <= 7
```

stephaniehingtgen: Good point. 

I updated it to be:
`jira_maintenance_expiry_days_gauge <= 7 and jira_maintenance_expiry_days_gauge > 0`

so that it doesn't overlap with the expired alert.

jira-mixin/alerts/general.yaml
stephaniehingtgen: I'm not sure of the functionality for that, so that's a good point. I updated it to be <= 0

jira-mixin/alerts/general.yaml
stephaniehingtgen: So at first I had tried using the metric `jvm_uptime_gauge`. But when I tested it locally (by making it go into a OOM killed crashloop), it wasn't alerting that it was down. But changing it to `up` then did set off that alert. 

The original metric was
`jvm_uptime_gauge == 0`

Is there a better way to do that that will capture when it's crash looping as well?

hairyhenderson: > Is there a better way to do that that will capture when it's crash looping as well?

Usually alerting on crashlooping pods is done in a general way for all pods using the metrics that come from kube-state-metrics. I was thinking more of the use-case where the exporter is running but is misconfigured and isn't properly reaching its target...

I just realized though that this exporter is a plugin that runs _in_ JIRA as a plugin! That explains why `up` works 😂

In that case it'd probably be best to omit this entirely, and leave up/down detection to something more general. In the past I've configured a generic Prometheus alert something like `sum (up) by (job) / count (up) by (job) < 0.75` - that'll return the ratio of up to discovered and trigger when it's <75%... Either way that part should probably be left to a higher-level more general alert 😉

stephaniehingtgen: Thanks for the explanation! I removed the InstanceDown alert from the alert yaml and from the README.

hairyhenderson (approved):
LGTM (though I think it also needs to be reviewed by someone else!)

hairyhenderson:
cc @rgeyer maybe?

rgeyer (approved):
This looks great!

One minor thing. Can you add templating for the datasource, job, and instance please?

This provides a bit more flexibility for using the mixin in different environments and in cases where there may be more than one Jira instance for which metrics are collected.

[Here ](https://github.com/grafana/jsonnet-libs/blob/master/go-runtime-mixin/dashboards/go-runtime.json#L1062-L1128)is a good example from the golang mixin. You'd need to adjust the metric you use to filter the label_values from.

Once that's good, we can add this to an integration for the onboarding plugin too 😃

stephaniehingtgen:
@rgeyer Good idea! I just changed it to templating and tested it out locally. Let me know if you'd like to take a second look at that - otherwise I'll merge this in at the end of the day today

rgeyer:
> @rgeyer Good idea! I just changed it to templating and tested it out locally. Let me know if you'd like to take a second look at that - otherwise I'll merge this in at the end of the day today

Looks great! Thanks for this addition :)

PR #581, opened 2021-06-18
feat(alertmanager): add slack and pd receiver libs

Duologic:
- Moves the Slack one into a composable lib.
- Adds Pagerduty one with battle tested templates.

Ideally these would be generated from the Prometheus code or docs instead of manually curated.


julienduchesne (approved):
Nice!

PR #582, opened 2021-06-22
Balance Grafana dashboard shards more smartly

julienduchesne:
Instead of going randomly, this algorithm tries to balance them
We used to have shards with no dashboards at all and other huge shards

Also, this calculates dashboard shards dynamically. It limits at 100kB per shard so the `grafanaDashboardShards` attribute is no longer needed

Example from Cortex:

**Before with 10 shards:**
Number of dashboards - Number of lines
```
ConfigMap-dashboards-cortex-0.yaml: 1 - 167
ConfigMap-dashboards-cortex-1.yaml: 0 - 7
ConfigMap-dashboards-cortex-2.yaml: 4 - 1629
ConfigMap-dashboards-cortex-3.yaml: 2 - 434
ConfigMap-dashboards-cortex-4.yaml: 2 - 289
ConfigMap-dashboards-cortex-5.yaml: 0 - 7
ConfigMap-dashboards-cortex-6.yaml: 1 - 270
ConfigMap-dashboards-cortex-7.yaml: 1 - 1027
ConfigMap-dashboards-cortex-8.yaml: 3 - 371
ConfigMap-dashboards-cortex-9.yaml: 5 - 1652
```

**After with 6 shards:**
Number of dashboards - Number of lines
```
ConfigMap-dashboards-cortex-0.yaml: 3 - 1131
ConfigMap-dashboards-cortex-1.yaml: 3 - 840
ConfigMap-dashboards-cortex-2.yaml: 3 - 834
ConfigMap-dashboards-cortex-3.yaml: 3 - 940
ConfigMap-dashboards-cortex-4.yaml: 3 - 962
ConfigMap-dashboards-cortex-5.yaml: 4 - 1115
```

PR #585, opened 2021-06-23
Fix: Mount calculated shards correctly

julienduchesne:
The shardedMounts function still used the `shards` attribute
Follow-up to https://github.com/grafana/jsonnet-libs/pull/582


julienduchesne:
Note that this changes mount paths in this way: 
image

Duologic (approved):

PR #587, opened 2021-06-23
Update from all active users to all users and fix license alert

stephaniehingtgen:
Two bug fixes for alerts:
- Changes jira_all_active_users to jira_all_users. These were the same when running locally, but after looking through the metrics, the max number of users will go off of all users, not all active users.
- Changes jira_maintenance_expiry_days_gauge to jira_license_expiry_days_gauge. These are the same when running locally as well, but the license is what we want to alert on.

Note: this is already fixed in the [integration PR](https://github.com/grafana/cloud-onboarding/pull/179)

### DIFF
--- a/jira-mixin/alerts/general.yaml
+++ b/jira-mixin/alerts/general.yaml
@@ -2,7 +2,7 @@ groups:
 - name: alert.rules
   rules:
   - alert: LicenseExpired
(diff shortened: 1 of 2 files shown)


rgeyer (approved):
LGTM

PR #592, opened 2021-06-25
Prometheus has renamed its default branch from master to main.

tsekine354:
Prometheus git repo doesn't have `master` branch any more, as per https://github.com/prometheus/prometheus/issues/8527.

This change is to follow that change, to track `main` branch.


malcolmholmes (approved):

malcolmholmes:
Thanks!

PR #598, opened 2021-06-29
feat: Upgrade to enterprise-metrics v1.4.0 release

jdbaldry:


enterprise-metrics/scripts/k3d
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env bash
```
chaudum, line 1: I would not name this script `k3d` ;) since it may interfere with the real [`k3d`](https://k3d.io/)

enterprise-metrics/Makefile
```diff
@@ -22,6 +25,25 @@ test: ## Run library tests.
...
+environments/minio/jsonnetfile.json: environments/minio/jsonnetfile.jsonnet jsonnetfile.json test/jsonnetfile.json
+	jsonnet $< > $@
+
+environments/minio/vendor: ## Vendor environment dependencies.
+environments/minio/vendor: environments/minio/jsonnetfile.json
+	cd $(@D)
+	jb install
+
+e2e-test: ## Deploy GEM to a local k3d cluster.
+e2e-test: environments/minio/vendor
```
chaudum, line 43: I would not call this an e2e test, rather a local test setup.
From an e2e test I would expect setup, test case and automatic teardown.

chaudum (approved):
Two small nitpicks, but LGTM!

enterprise-metrics/scripts/k3d
jdbaldry: fair point, updated in https://github.com/grafana/jsonnet-libs/pull/598/commits/f4e41500cd5736eb21b5887cf4c57af5b97c464e

enterprise-metrics/Makefile
jdbaldry: Also a good point, updated in https://github.com/grafana/jsonnet-libs/pull/598/commits/f4e41500cd5736eb21b5887cf4c57af5b97c464e

chaudum:
lgtm!

PR #600, opened 2021-06-29
Make Secret name for GEM admin token configurable

chaudum:


enterprise-metrics/main.libsonnet
```diff
@@ -156,6 +156,17 @@ local removeNamespaceReferences(args) = std.map(function(arg) std.strReplace(arg
       type=d.T.string,
     ),
     licenseSecretName: 'gem-license',
+    '#adminTokenSecretName':: d.val(
+      default=self.adminTokenSecretName,
+      help=|||
+        When generating an admin token using the `tokengen` target, the result is written to the Kubernetes
+        Secret `adminTokenSecretName`.
+        To retrive the token from Kubernetes using `kubectl`:
```
jdbaldry, line 164:
```suggestion
        To retrieve the token from Kubernetes using `kubectl`:
```

enterprise-metrics/main.libsonnet
```diff
@@ -156,6 +156,17 @@ local removeNamespaceReferences(args) = std.map(function(arg) std.strReplace(arg
...
+        $ kubectl get secret gem-admin-token -o jsonpath="{.data.grafana-token}" | base64 --decode ; echo
```
jdbaldry, line 165: IIRC, the grafana-token is actually double base64 encoded, once thanks it being a value in a Kubernetes Secret and another time because Grafana provisioning expects the token to be base64 encoded. 

With that in mind, I think we should clarify the docs here to use the `.data.token` field instead which will give users a usable token. Or perhaps we just need to include some variation of the previous paragraph as additional docs?

enterprise-metrics/main.libsonnet
```diff
@@ -538,9 +549,13 @@ local removeNamespaceReferences(args) = std.map(function(arg) std.strReplace(arg
     createSecretContainer::
       container.new('create-secret', this._images.kubectl)
       + container.withCommand([
-        '/bin/bash',
-        '-euc',
-        'kubectl create secret generic gem-admin-token --from-file=token=/shared/admin-token --from-literal=grafana-token="$(base64 <(echo :$(cat /shared/admin-token)))"',
+        'kubectl',
```
jdbaldry, line 552: This is much easier to read, would you be able to double check this still works as expected, perhaps using the test tooling in #598?

jdbaldry (approved):
LGTM, could you add a CHANGELOG entry too?

enterprise-metrics/main.libsonnet
chaudum: Sure, once you merged I can rebase :)

enterprise-metrics/main.libsonnet
chaudum: Updated the help text a bit. Wdyt?

PR #601, opened 2021-06-29
fixes(promk): override node_exporter container

Duologic:
Properly override the container.

### DIFF
--- a/prometheus-ksonnet/lib/node-exporter.libsonnet
+++ b/prometheus-ksonnet/lib/node-exporter.libsonnet
@@ -10,8 +10,7 @@ local node_exporter = import 'github.com/grafana/jsonnet-libs/node-exporter/main
   node_exporter_container:: exporter.container,
 
   node_exporter_daemonset:
-    exporter.daemonset
-    + $.apps.v1.daemonSet.spec.template.spec.withContainers($.node_exporter_container)
+    (exporter { container:: $.node_exporter_container }).daemonset
     + $.util.podPriority('critical'),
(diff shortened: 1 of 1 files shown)


prometheus-ksonnet/lib/node-exporter.libsonnet
```diff
@@ -10,8 +10,7 @@ local node_exporter = import 'github.com/grafana/jsonnet-libs/node-exporter/main
   node_exporter_container:: exporter.container,
 
   node_exporter_daemonset:
-    exporter.daemonset
-    + $.apps.v1.daemonSet.spec.template.spec.withContainers($.node_exporter_container)
+    (exporter { container:: $.node_exporter_container }).daemonset
```
julienduchesne, line 13: Should this be a `withContainer` function in node-exporter code?

julienduchesne (approved):

prometheus-ksonnet/lib/node-exporter.libsonnet
Duologic: I'm jumping hoops here not to break backwards compatibility.

The node-exporter lib defines the container in the `container::` object, then `daemonSet.new(..., [self.container])` references that. When we call `exporter.daemonset` then `k.util.hostVolumeMount` will be called, it uses `mapContainers` to map the volumes to the containers, overriding the `spec.template.spec` with `withContainers` will undo that mapping. Hence we override `container::` before calling `exporter.daemonset`.

PR #602, opened 2021-07-01
Traefik mixin

DuvalKingJacob:
Screen Shot 2021-07-01 at 1 40 53 PM
Screen Shot 2021-07-01 at 1 41 04 PM

Traefik is all about that Load Balancer Life.  This will all tie together (Nomad, Vault, Traefik) into an eventual Content Release of All three Integrations https://github.com/fhemberger/nomad-demo


nomad-mixin/mixin.libsonnet
```diff
@@ -0,0 +1,6 @@
+{
```
rgeyer, line 1: Looks like you based your traefik branch off of the nomad branch, so these commits are included.

That's fine, but make sure you merge the nomad PR before this one, just for tidyness sake.

rgeyer (approved):
LGTM, Hold off on merging until nomad is merged plz.

PR #603, opened 2021-07-02
chore: s/k8s-alpha/k8s-libsonnet

Duologic:
ref: https://github.com/jsonnet-libs/k8s/pull/39

This library was renamed upstream.


mplzik (approved):
LGTM

PR #609, opened 2021-07-09
fix: update kube-state-metrics image (linked image not available)

xvzf:
Signed-off-by: Matthias Riegler <matthias.riegler@taotesting.com>


jvrplmlmn (approved):
👍

PR #617, opened 2021-07-19
Manage Prometheus pods in parallel

bboreham:
The default behaviour of Kubernetes Statefulset, `OrderedReady` waits for pod-0 to be ready before starting pod-1.

When running two Prometheus' in "HA" configuration, we very much want pod-1 to be running when pod-0 is not ready. The alternative management policy, `Parallel`, is better for that.

Docs: https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#pod-management-policies

### DIFF
--- a/prometheus/prometheus.libsonnet
+++ b/prometheus/prometheus.libsonnet
@@ -149,6 +149,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
       self.prometheus_watch_container,
     ], self.prometheus_pvc)
     + self.prometheus_config_mount
+    + statefulset.mixin.spec.withPodManagementPolicy('Parallel')
     + statefulset.mixin.spec.withServiceName('prometheus')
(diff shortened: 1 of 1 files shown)


Duologic:
One of the reasons for this HA config to exist is that we would always have at least 1 running Prometheus on roll out of a new version (or restarting a statefulSet). Wouldn't this option mean that on a rolling restart, both Prometheis pods can be down at the same time?

Duologic:
pinging @beorn7 @malcolmholmes

bboreham:
No, see https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#rolling-updates

Duologic:
Gotcha, last part of the `Parallel Pod Management` paragraph was "Updates are not affected."

Duologic (approved):

beorn7 (approved):
Nice catch.

jvrplmlmn (approved):

PR #619, opened 2021-07-22
Trigger grafana deployment on every config change

the-it:
General ini config map had already a hash to trigger a new deployment on every config change.
This commit adds this behaviour for the other config maps:

* dashboards
* datasources
* notification channels


jvrplmlmn (approved):
Nice 👍 

This covers the scenario that we witness internally in which after updating a datasource Grafana was not re-deployed to pick up the change. Since we paired on this, LG2M, but I'll defer a second approval to either @Duologic or @malcolmholmes 😃

Duologic (approved):

PR #623, opened 2021-07-24
alertmanager: restart pods on config change

asymmetric:
This is analogous to #619.

### DIFF
--- a/alertmanager/alertmanager.libsonnet
+++ b/alertmanager/alertmanager.libsonnet
@@ -151,7 +151,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
     + statefulset.mixin.spec.template.spec.securityContext.withFsGroup(2000)
     + statefulset.mixin.spec.template.spec.securityContext.withRunAsUser(1000)
     + statefulset.mixin.spec.template.spec.securityContext.withRunAsNonRoot(true)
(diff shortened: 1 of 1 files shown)


asymmetric:
Hm, I now notice that we run a separate container that [triggers a config reload](https://github.com/grafana/jsonnet-libs/blob/fe0ecb1adfcdab5550a5a1c5f3275ae0598e818b/alertmanager/alertmanager.libsonnet#L118) on change.

What's the benefit there, compared to redeploying the statefulset?

Duologic:
I think @tomwilkie introduced the `watch` container, care to share your opinion?

jvrplmlmn:
My understanding is:

The `watch` container will try to reload the config, but if its invalid that won't disrupt the working Alertmanager.

If you would redeploy the StatefulSet with a new config that is invalid, it might prevent it from starting.

tomwilkie:
> The watch container will try to reload the config, but if its invalid that won't disrupt the working Alertmanager.

We have an alert to detect that.

> If you would redeploy the StatefulSet with a new config that is invalid, it might prevent it from starting.

Indeed; its deemed more important that the alerting work than it uses the latest config.  So we'd rather a config reload fails than we restart the service and it fails to start (causing an outage).

PR #624, opened 2021-07-26
nginx-directory: preserve url encoded chars

darrenjaneczek:
It was discovered that encoded characters such as `%20` were converted through the proxy engine, resulting in 400 "bad request." This would affect urls such as:
  e.g., `baseurl.com/server_name/resource%20with%20space`

The previous solution uses a regex location selector, and with that approach the `$2` group variable would decode the url, causing the 400 if there were any special %-encoded characters. The example above, resulted in:
`baseurl.com/server_name/resource with space`

After replacing to a non-regex location selector, some additional efforts were required to ensure:
- relative urls work within the proxy subdir
- redirects from the target server are translated correctly
- %-encoded characters proxy without 400 "bad request"

### DIFF
--- a/nginx-directory/configfile.libsonnet
+++ b/nginx-directory/configfile.libsonnet
@@ -11,9 +11,9 @@ function(services) {
         return 302 %(url)s;
       ||| % service
       else |||
-        #automagically replaces non-regex location identifier with given proxy_pass url.
+        # automagically replaces non-regex location identifier with given proxy_pass url.
         proxy_pass          %(url)s;
-        #specifying redirect with scheme and http_host because the "default" option doesn't persist the same port.
+        # specifying redirect with scheme and http_host because the "default" option doesn't persist the same port.
         proxy_redirect      %(url)s $scheme://$http_host/%(path)s/;
         proxy_set_header    X-Real-IP $remote_addr;
         proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
--- a/nginx-directory/configfile.libsonnet
+++ b/nginx-directory/configfile.libsonnet
@@ -51,7 +51,7 @@ function(services) {
     |||
       location ^~ /%(path)s {
         # Append trailing / as needed
-        return 302 $scheme://$http_host$request_uri/;
+        rewrite ^/%(path)s$ $scheme://$http_host/%(path)s/ redirect;
       }
       location ^~ /%(path)s/ {
     ||| % service +
--- a/nginx-directory/configfile.libsonnet
+++ b/nginx-directory/configfile.libsonnet
@@ -11,8 +11,10 @@ function(services) {
         return 302 %(url)s;
       ||| % service
       else |||
-        proxy_pass      %(url)s$2$is_args$args;
-        proxy_set_header    Host $host;
+        #automagically replaces non-regex location identifier with given proxy_pass url.
+        proxy_pass          %(url)s;
+        #specifying redirect with scheme and http_host because the "default" option doesn't persist the same port.
+        proxy_redirect      %(url)s $scheme://$http_host/%(path)s/;
         proxy_set_header    X-Real-IP $remote_addr;
         proxy_set_header    X-Forwarded-For $proxy_add_x_forwarded_for;
         proxy_set_header    X-Forwarded-Proto $scheme;
@@ -47,7 +49,11 @@ function(services) {
 
   local buildLocation(service) =
     |||
-      location ~ ^/%(path)s(/?)(.*)$ {
+      location ^~ /%(path)s {
+        # Append trailing / as needed
+        return 302 $scheme://$http_host$request_uri/;
+      }
+      location ^~ /%(path)s/ {
     ||| % service +
     buildHeaders(service) +
     |||


nginx-directory/configfile.libsonnet
```diff
@@ -11,8 +11,10 @@ function(services) {
         return 302 %(url)s;
       ||| % service
       else |||
-        proxy_pass      %(url)s$2$is_args$args;
-        proxy_set_header    Host $host;
+        #automagically replaces non-regex location identifier with given proxy_pass url.
+        proxy_pass          %(url)s;
+        #specifying redirect with scheme and http_host because the "default" option doesn't persist the same port.
```
jdbaldry, line 16: If I understand the `proxy_redirect` documentation correctly, "default" should persist the port if our `proxy_pass` URL does not consist of NGINX variables. I guess it's safer to be explicit here because there is nothing stopping a consumer of this library using NGINX variables in the interpolated `url` variable.

nginx-directory/configfile.libsonnet
```diff
@@ -11,8 +11,10 @@ function(services) {
         return 302 %(url)s;
       ||| % service
       else |||
-        proxy_pass      %(url)s$2$is_args$args;
-        proxy_set_header    Host $host;
+        #automagically replaces non-regex location identifier with given proxy_pass url.
```
jdbaldry, line 14: nit: My eyes dislike comments that don't start with a space but it may just be me.

nginx-directory/configfile.libsonnet
```diff
@@ -11,8 +11,10 @@ function(services) {
         return 302 %(url)s;
       ||| % service
       else |||
-        proxy_pass      %(url)s$2$is_args$args;
```
jdbaldry, line 14: Have we lost the handling of query parameters here? If I understand HTTP redirects correctly the client will request the exact Location URL and that needs to include the query parameters if we wish to pass them to redirect.

jdbaldry (commented):
Thanks for the great investigation into these errors. I have a couple questions to help me understand the PR better.

nginx-directory/configfile.libsonnet
darrenjaneczek: I wonder if the disconnect was due to my port was being forwarded from localhost:8080:
`kubectl port-forward --namespace darren-test deployment/nginx 8080:80`.

With the old setup, and various iterations working on this, redirects from my local address (e.g., `localhost:8080/bacon/some_directory`) would redirect to the following:
- `localhost/some_directory/`,
- `localhost/bacon/some_directory/`, and once I think
- `localhost:8080/some_directory/`. 
 
Being this explicit seemed to be "just right".

nginx-directory/configfile.libsonnet
darrenjaneczek: The query parameters do remain on the URL visible in the browser, and the logs from my webserver on the other end of the proxy look like this:
```34.122.9.230 - - [26/Jul/2021:10:07:11 -0400] "GET /?parameter=hello HTTP/1.0" 200 1758 "-" "Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.164 Safari/537.36"
```

darrenjaneczek: It seems that with `location` being in non-regex mode, it is able to automatically keep everything.
Unfortunately when in regex mode, the `$2` in this line caused the `Space%20Separated` to turn into `Space Separated`.

nginx-directory/configfile.libsonnet
darrenjaneczek: fixed.

nginx-directory/configfile.libsonnet
jdbaldry: Thanks for confirming. I don't know NGINX very well but that observational evidence along with the explanation of the regexp behavior makes sense and gives me confidence that we aren't introducing any regression.

nginx-directory/configfile.libsonnet
jdbaldry: Thanks!

nginx-directory/configfile.libsonnet
jdbaldry: I agree, I don't see any problems with being explicit in this case.

nginx-directory/configfile.libsonnet
jdbaldry: One thing I've noticed while experimenting with this change is that this comment will be repeated for every entry in the NGINX directory. I wonder if we should consider moving the comment outside of the actual configuration string to avoid this. The downside would be that the comment would need more context and would be more likely to be neglected in the future.

darrenjaneczek:
Not ready for merge.
Will work in this further in a few weeks.

Required outstanding fixes:
-[ ] Handle paths that end in `/`, e.g., `path: 'datadog-grafana/',`
-[ ] Handle requests to legitimate "path" entries, which end in `//`, e.g., `https://admin-dev-us-central-0.grafana.net/ge-metrics-grafana//`
-[ ] Note potential environment setup errors if there are invalid entries in the configuration; see:
  - https://github.com/grafana/deployment_tools/pull/13996
  - https://github.com/grafana/deployment_tools/pull/13998 
    ```
    $ kubectl -n default logs nginx-6bc87ddb99-74qd2 
    2021/07/30 13:59:12 [emerg] 1#1: host not found in upstream "minio.ge-metrics.svc.cluster.local" in /etc/nginx/nginx.conf:443
    nginx: [emerg] host not found in upstream "minio.ge-metrics.svc.cluster.local" in /etc/nginx/nginx.conf:443
    ```

darrenjaneczek:
Plans to look into this again this quarter.
December at the latest.

jdbaldry:
@darrenjaneczek Is there still desire to get this merged?

darrenjaneczek:
Yes. I still need to correct some of the identified issues above.

gaantunes:
Hello folks, is this still WIP?

jdbaldry:
I believe so. @darrenjaneczek is this something you want to pick up again?

darrenjaneczek:
Sorry for this prolonged story. The effort is over a year old now!

The solution as it appears in this PR isn't complete because it creates some errors, itemized here:
- https://github.com/grafana/jsonnet-libs/pull/624#issuecomment-889938857

My difficulty in picking this up again is that I would need to re-learn how to set up my environment to work with these files. I would probably need some help to get it up and running again, and to refresh my mental model of how to work with nginx.

jdbaldry:
I'm not sure that the cost of revisiting this issue is worth it at this point in time unless we have seen continued issues with the current behavior so I'm in favor of closing this PR and using it as a starting point if we need to revisit the issue at a later point.

darrenjaneczek:
That sounds pragmatic. It is rare that any of our APIs require spaces in the URLs, and that particular one was worked-around long ago.

PR #634, opened 2021-08-15
alertmanager: fix pagerduty routing key

asymmetric:
The correct name for apiv2 is `routing_key`.

Cf. https://prometheus.io/docs/alerting/latest/configuration/#pagerduty_config

### DIFF
--- a/alertmanager/receivers/pagerduty.libsonnet
+++ b/alertmanager/receivers/pagerduty.libsonnet
@@ -45,8 +45,8 @@
       service_key: service_key,
     } + this.config.withConfigTemplate(),
 
-    newRouting(service_key): {
-      service_key: service_key,
+    newRouting(routing_key): {
+      routing_key: routing_key,
     } + this.config.withConfigTemplate(),
 
     withConfigTemplate(): configTemplate,


Duologic:
> Alertmanager now supports Events API v2. However, if you set the routing_key property and use v2, the integration type of the integration corresponding to the routing_key value must also be Events API v2. If you select Prometheus as the integration type in PagerDuty, you will need to use the Events API v1 type and set a value for the service_key property instead.

The documentation says this, so it depends on how you use it. Most of our internal rotation are currently using type 'prometheus'. Maybe an extra jsonnet function that sets `routing_key` can be added to this lib?

asymmetric:
@Duologic what's the point of the `newRouting` function if not to set `routing_key`?

>      # The following two options are mutually exclusive.
>      # The PagerDuty integration key (when using PagerDuty integration type `Events API v2`).
>      routing_key: <tmpl_secret>
>      # The PagerDuty integration key (when using PagerDuty integration type `Prometheus`).
>      service_key: <tmpl_secret>

IIUC, if you set a `routing_key`, then you must use v2 and use the "events v2" integration type, but that is not something that concerns the Prometheus config, it's a setting in PD.

Duologic (approved):
I see, missed the first bit on mobile. :grimacing:

PR #635, opened 2021-08-16
Simpler API for Grizzly resources

malcolmholmes:
The current `util.makeResource()` method for creating resources doesn't feel right.

This PR adds a better way of describing resources, e.g. embedding the apiVersion/kind
into a Jsonnet function. Grizzly examples will be updated to use this clearer method
once merged.

### DIFF
--- a/grizzly/prometheus.libsonnet
+++ b/grizzly/prometheus.libsonnet
@@ -2,6 +2,8 @@ local util = import 'util.libsonnet';
 local kind = 'PrometheusRuleGroup';
 local recordingRules = 'prometheusRules';
 local alertRules = 'prometheusAlerts';
+local resource = import 'resource.libsonnet';
+
 {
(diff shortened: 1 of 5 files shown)


grizzly/prometheus.libsonnet
```diff
@@ -27,4 +29,11 @@ local alertRules = 'prometheusAlerts';
       )
     for key in std.objectFields(mixins)
   },
+
+  rule_group: {
+    new(namespace, name, group)::
+      resource.new('PrometheusRuleGroup', name)
+      + resource.addMetadata('namespace', namespace)
+      + resource.withSpec(group),
```
Duologic, line 37: I'd suggest layering `group` one step deeper so you can add siblings to the group in future revisions:

```
spec:
  group: ...
```

The same for all other resources.

Duologic (approved):

PR #642, opened 2021-09-08
enterprise-metrics: Enable admin-api leader election by default 

jdbaldry:
This change does not affect single replica deployments of the
admin-api but does fix the potential for an inconsistent state when
running with multiple replicas of the admin-api and experiencing
parallel writes for the same objects.


56quarters (approved):
LGTM

PR #644, opened 2021-09-16
use std.manifestYamlDoc function instead of custom native function

Andor:
In case of using jsonnet-libs outside of Tanka, there is no function `std.native('manifestYamlFromJson')` available.
But there are semantically similar function exists in jsonnet itself: `std.manifestYamlDoc`.

So, for portability reasons, I prefer to use std functions instead of custom native function in util.manifestYaml helper function.

Without it, I cannot use this simple code to manage grafana with [qbec](https://qbec.io/)
```
  prometheusDatasource:
    grafana.datasource.new(
      'prometheus',
      p.components.prometheus.ingress.host,
      type = 'prometheus',
      default = true,
    ),

  grafana:
    grafana
    + grafana.withAnonymous()
    + grafana.addFolder('Example')
    + grafana.addDatasource('prometheus', $.prometheusDatasource),
```
because it fails:
```
✘ evaluate 'grafana': RUNTIME ERROR: Unexpected type null, expected function
        vendor/ksonnet-util/util.libsonnet:206:5-27     function <anonymous>
        vendor/github.com/grafana/jsonnet-libs/grafana/configmaps.libsonnet:(40:25)-(56:9)      object <anonymous>
        Field "dashboards.yml"
        Field "data"
        Field "dashboard_provisioning_config_map"
        Array element 0
        During manifestation
```

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -202,8 +202,7 @@ local util(k) = {
     ]),
 
   manifestYaml(value):: (
-    local f = std.native('manifestYamlFromJson');
-    f(std.toString(value))
+    std.manifestYamlDoc(value)
   ),
 
   resourcesRequests(cpu, memory)::


Andor:
This is quite possible I have to replace it with a bit more sophisticated code like this:
```
std.manifestYamlDoc(std.parseJson(std.toString(value)))
```
Which will allow to accept either json-formatted string or object as a value of the helper funciton.

Andor:
@jdbaldry Do you think there is any point to maintain backward compatibility with 3+ year old jsonnet versions?

jdbaldry:
I was thinking less of backwards compatibility with old Jsonnet and more with users that have been using Tanka and expect a certain YAML output. Though the standard library function should produce something semantically equivalent, it will cause big diffs for those used to the Tanka native function.

PR #646, opened 2021-09-20
doc: Provide example runtime override

jdbaldry:


enterprise-metrics/docs/README.md
```diff
@@ -1110,6 +1110,23 @@ $ kubectl create secret generic gem-license -from-file=license.jwt
 ## obj runtime.configuration.overrides
 
 `overrides` are per tenant runtime limits overrides.
+Each field should be keyed by tenant ID and have an object value containing the specific overrides.
```
MichelHollands, line 1113: (nit) Doesn't each object value contain only 1 override?

```suggestion
Each field should be keyed by tenant ID and have an object value containing the specific override.
```

MichelHollands (approved):
LGTM

PR #651, opened 2021-09-24
Fix aggregation in memcached dashboard

pracucci:
The memcached dashboard uses many queries aggregating by `without (...)`, but the label displayed is just `{{ instance }}`. This leads to panels like the following, where multiple series have the same exact label:
Screenshot 2021-09-24 at 16 49 03

To fix it, I suggest to group by the same labels we configure in the panel output.

### DIFF
--- a/memcached-mixin/dashboards.libsonnet
+++ b/memcached-mixin/dashboards.libsonnet
@@ -32,29 +32,29 @@ local g = (import 'grafana-builder/grafana.libsonnet');
         g.row('Ops')
         .addPanel(
           g.panel('Commands') +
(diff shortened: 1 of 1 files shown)


pracucci:
@dannykopping Asking a review to you cause I've seen you worked on it recently.

dannykopping (approved):
LGTM, thanks @pracucci!

jdbaldry:
I understand that _Prometheus: Up & Running_ prefers the `without` clause over the `on` clause but given that we are overriding the legend, this is absolutely an improvement over what we have now.

I'm personally a fan of having the labels rather than specifying a legend because it helps me better relate the graph to the query but I think the legend can be more user friendly for the purposes of visualization.

jdbaldry (approved):

PR #653, opened 2021-09-29
fix: Remove auth.type=default override on querier

jdbaldry:
The bug that this worked around is fixed in recent versions of GEM.


56quarters (approved):
LGTM

PR #654, opened 2021-09-29
Improve the connections panel in memcached dashboard

pracucci:
The "Connections" panel in the memcached dashboard currently suffers two issues:
1. It mixes "connections / sec" and "current connections", and this makes it difficult to have a clear understanding
2. It shows the "max connections" limit setting for every instance, and this introduces a lot of noise if you have many instances

For example, do you get any real value from this panel?
Screenshot 2021-09-29 at 19 01 14

In this PR I propose two changes:
1. Split "connections / sec" and "current connections" into 2 different panels
2. Show "max connections" only once, picking the "min value across all instances"

I've also replaced `[1m]` with `[$__rate_interval]` to play more nicely with the selected time range.

### DIFF
--- a/memcached-mixin/dashboards.libsonnet
+++ b/memcached-mixin/dashboards.libsonnet
@@ -14,7 +14,7 @@ local g = (import 'grafana-builder/grafana.libsonnet');
         g.row('Hits')
         .addPanel(
           g.panel('Hit Rate') +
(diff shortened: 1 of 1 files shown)


dannykopping (approved):
LGTM

rgeyer (approved):
@pracucci I didn't notice this PR when making the same changes for `$__rate_interval` so I just took the liberty of resolving the conflict I created.

This looks good to me as well. I'm in the process of updating some of our Grafana Cloud integrations, and would like to use this latest version of the mixin.

jdbaldry (approved):
LGTM

pracucci:
> @pracucci I didn't notice this PR when making the same changes for `$__rate_interval` so I just took the liberty of resolving the conflict I created.
> 
> This looks good to me as well. I'm in the process of updating some of our Grafana Cloud integrations, and would like to use this latest version of the mixin.

Too kind @rgeyer, thanks!

PR #656, opened 2021-09-30
traefik-mixin rate and timerange standardization

rgeyer:

Update all dashboards to default to 30m timerange.

Some drive-by fixes to queries to better match (I think?) their design intent.


DuvalKingJacob (approved):
LGTM

PR #658, opened 2021-09-30
ruby-mixin rate and timerange standardization

rgeyer:

Update all dashboards to default to 30m timerange.


cristiangreco (approved):
🚀

PR #660, opened 2021-10-01
Caddy-mixin rate standardization

gaantunes:
Changing range to $__rate_interval on promql queries using rate(), irate(), historical_quatile() and increase() functions.

### DIFF
--- a/caddy-mixin/dashboards/caddy-overview.json
+++ b/caddy-mixin/dashboards/caddy-overview.json
@@ -622,6 +622,6 @@
   },
   "timezone": "",
   "title": "Caddy Overview",
-  "uid": "Al2uDkD7z",
-  "version": 2
+  "uid": "9B0qPnfMz",
+  "version": 9
 }
--- a/caddy-mixin/dashboards/caddy-overview.json
+++ b/caddy-mixin/dashboards/caddy-overview.json
@@ -10,6 +10,12 @@
         "limit": 100,
         "name": "Annotations & Alerts",
         "showIn": 0,
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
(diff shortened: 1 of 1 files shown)


caddy-mixin/dashboards/caddy-overview.json
```diff
@@ -69,7 +69,7 @@
       "steppedLine": false,
       "targets": [
         {
-          "expr": "sum(rate(caddy_http_requests_total{handler=\"reverse_proxy\"}[$interval]))",
```
rgeyer, line 72: Now that you have replaced `$interval` with `$__rate_interval` the templating variable named `interval` is no longer used.

It should be removed.

rgeyer (changes requested):
One comment below. Also since @besartberisha contributed this one, we should add him for review.

caddy-mixin/dashboards/caddy-overview.json
```diff
@@ -685,6 +622,6 @@
   },
   "timezone": "",
   "title": "Caddy Overview",
-  "uid": "9B0qPnfMz",
-  "version": 8
+  "uid": "Al2uDkD7z",
```
rgeyer, line 625: Hmmn.. I'm not sure what the implication of changing the UID might be. I know in some cases it's used to link from other dashboards, so if a customer is expecting to link to Caddy with it's current UID this will break it.

Could we restore the old UID, and set the version to 9?

rgeyer (changes requested):
One more small change.

besartberisha:
>One comment below. Also since @besartberisha contributed this one, we should add him for review.

Hey @rgeyer thanks for pinging me but since I haven't checked this since I initially worked on it, I think I'm not the right person to properly review it so feel free to make any adjustments you see fit.

rgeyer (approved):

PR #663, opened 2021-10-01
ceph-mixin rate standardization

gaantunes:

### DIFF
--- a/ceph-mixin/dashboards/ceph-cluster.json
+++ b/ceph-mixin/dashboards/ceph-cluster.json
@@ -117,7 +117,7 @@
           "expr": "sum without (instance) (ceph_health_status{ceph_cluster=\"$ceph_cluster\"})",
           "format": "time_series",
           "instant": true,
-          "interval": "$interval",
+          "interval": "",
           "intervalFactor": 1,
           "legendFormat": " ",
           "refId": "A",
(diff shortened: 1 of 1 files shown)


rgeyer (approved):
LGTM.

The `$interval` is still used on the targets themselves, but not in the promql function calls. I *think* that's okay.

gaantunes:
Do you think it would be better to set the interval to $__rate_interval as well? I didn`t change it on any of the other dashboards, wasn`t paying attention to this to be honest...

rgeyer:
> Do you think it would be better to set the interval to $__rate_interval as well? I didn`t change it on any of the other dashboards, wasn`t paying attention to this to be honest...

Yeah, if it isn't being used explicitly, we can remove it.. I'll make another quick commit to handle that.

rgeyer:
> > Do you think it would be better to set the interval to $__rate_interval as well? I didn`t change it on any of the other dashboards, wasn`t paying attention to this to be honest...
> 
> Yeah, if it isn't being used explicitly, we can remove it.. I'll make another quick commit to handle that.

Sorry, I just realized this was one you're working on. You're welcome to remove interval from the panels, and as a template var 😂

gaantunes:
> > > Do you think it would be better to set the interval to $__rate_interval as well? I didn`t change it on any of the other dashboards, wasn`t paying attention to this to be honest...
> > 
> > 
> > Yeah, if it isn't being used explicitly, we can remove it.. I'll make another quick commit to handle that.
> 
> Sorry, I just realized this was one you're working on. You're welcome to remove interval from the panels, and as a template var 😂

Will do it :D

PR #664, opened 2021-10-01
memcached-mixin rate standardization

rgeyer:


gaantunes (approved):
LGTM

PR #674, opened 2021-10-18
Migrate HAProxy mixin from jdbaldry/haproxy-mixin

jdbaldry:
Created with:

``` console
$ # From jdbaldry/haproxy-mixin repository
$ git log --binary --pretty=email --patch-with-stat --reverse > haproxy-mixin.patch
$ # From grafana/jsonnet-libs repository
$ git am --directory haproxy-mixin --committer-date-is-author-date haproxy-mixin.patch
```

Some empty commits were skipped but most of the history is retained.

The old repository used Drone CI which has been removed as part of this migration.


jdbaldry:
Sorry for the late reply, I forgot about this PR. 

I think this could possibly go upstream but it would need some pruning as this contains a lot of dev tooling and testing code that I'm not sure would make sense in the HAProxy repository. They previously had a `contrib` directory but that appears to have been shuffled away.

I intend to archive my repository and use this repository in the meantime.

PR #681, opened 2021-11-02
feat(docker-registry): add lib to deploy docker registry as pull through cache

Duologic:
Quick lib to get started with the docker registry.


inkel (approved):
Clean and simple, I like it.

PR #712, opened 2021-12-15
enable frontend logging for grafana

bergquist:


Duologic (approved):
defaulting to true? sure :)

bergquist:
We could override this in our internal infra but I think our jsonnet libs should be opinionated :)

PR #713, opened 2021-12-16
ha-mixin: unroll per-replica ConfigMap generation.

falfaro:
This allows overriden a replica-specific parameter of any Prometheus.

### DIFF
--- a/prometheus/ha-mixin.libsonnet
+++ b/prometheus/ha-mixin.libsonnet
@@ -35,29 +35,27 @@ function(replicas=2) {
     },
   },
 
+  // Allows overriding fields on each of the generated ConfigMaps
+  _prometheus_config_maps:: {
+    [_config.name + '-' + i + '.yml']: this.prometheus_config {
+      global+: {
+        external_labels+: {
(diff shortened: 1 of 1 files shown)


Duologic (commented):
I don't think this is needed, you can already access the individual keys in a configMap.

grafanalf:
> I don't think this is needed, you can already access the individual keys in a configMap.

I honestly thing my approach is much cleaner than this and more readable. Given how complex our Jsonnet codebase is, the easier making overrides more readable and easy to use are, the better.

falfaro:
> I don't think this is needed, you can already access the individual keys in a configMap.

I honestly thing my approach is much cleaner than this and more readable. Given how complex our Jsonnet codebase is, the easier making overrides more readable and easy to use are, the better.

falfaro:
So, you don't agree this new code is more readable and easy to follow, and easier to debug (evaluate) than the old one?

Duologic (approved):

Duologic:
The description didn't read like it was a readability improvement.

PR #714, opened 2021-12-17
fix: handle empty enabled features properly.

falfaro:
Without this fix, if no extra enabled featues are configures, an ugly `""` item will be produced among the list of command-line arguments to Prometheus.


the-it (approved):
lgtm

PR #715, opened 2021-12-29
update go-runtime mixin

v-zhuravlev:
Here is PR that fixes some golang mixin issues spotted:
- fix units for gc metric (seconds not ms)
- removed static tns metrics
- small style improvements
image

### DIFF
--- a/go-runtime-mixin/dashboards/go-runtime.json
+++ b/go-runtime-mixin/dashboards/go-runtime.json
@@ -283,19 +283,19 @@
         {
           "expr": "avg by (job)(go_memstats_mspan_sys_bytes{job=\"$job\", instance=~\"$instance\"})",
           "interval": "",
-          "legendFormat": "{{instance}}: mspan (avg)",
+          "legendFormat": "{{job}}: mspan (avg)",
           "refId": "B"
         },
         {
           "expr": "avg by (job)(go_memstats_mcache_sys_bytes{job=\"$job\", instance=~\"$instance\"})",
           "interval": "",
-          "legendFormat": "{{instance}}: mcache (avg)",
+          "legendFormat": "{{job}}: mcache (avg)",
           "refId": "D"
         },
         {
           "expr": "avg by (job)(go_memstats_buck_hash_sys_bytes{job=\"$job\", instance=~\"$instance\"})",
           "interval": "",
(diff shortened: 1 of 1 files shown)


go-runtime-mixin/dashboards/go-runtime.json
```diff
@@ -283,19 +283,19 @@
         {
           "expr": "avg by (job)(go_memstats_mspan_sys_bytes{job=\"$job\", instance=~\"$instance\"})",
           "interval": "",
-          "legendFormat": "{{instance}}: mspan (avg)",
```
rgeyer, line 286: It looks like you changed these to job, rather than instance, since these were the only panels which used instance, and all others used job?

Seems to me that instance is "right" in that it's more specific. You may have a single scrape_config that uses service discovery, and brings in several instances, and you'd want to be able to distinguish between them.

More confusing still, is that all of the panels have a description which indicates "across all instances of a job", but still has `instance="$instance"` in the query.

Perhaps @sdboyer can shed some light on this?

I believe we need to either;
1. Change the description for all panels to indicate it will show the selected instance(s), change the query and legend format to `instance`, and set the panel to stack the graphs.
2. Remove the instance template var entirely, and keep these legend changes.

#1 seems like the "right" solution. #2 would require an exception rule for the linter, since it expects that every dashboard has both job and instance template vars, and that they are used in all queries.

v-zhuravlev: All queries on this dashboard aggregates values across job (`avg by (job)`) effectively **dropping** instance label. 

If users choose instance=All - they would get an average values across all instances. 
If they choose specific instance - they would get values for this instance only, but {{instance}} still wouldn't be available for use in legendFormat field.

Since {{instance}} is always empty, I just replaced legendFormat to {{job}} so legend of this panel is aligned with other panels' legends.

For example, before this fix legend was:
`: mspan(avg)`
and now becomes:
`integrations/minio: mspan(avg)`

rgeyer: Okay, this makes sense. I had not realized that the "instance" var is mulit-select. So if the user selects a subset of instances it would be the average of those selected.

This LGTM. Will approve.

rgeyer (approved):
LGTM. Thank you!

PR #717, opened 2022-01-05
node-exporter: upgrade to v1.3.1

jvrplmlmn:
Most relevant bits are available in the [v1.3.0 CHANGELOG](https://github.com/prometheus/node_exporter/blob/master/CHANGELOG.md#130--2021-10-20)


the-it (approved):
lgtm

PR #719, opened 2022-01-12
NSQ mixin

v-zhuravlev:
This PR introduces [NSQ](https://nsq.io/overview/design.html) mixin.
NSQ [exposes](https://nsq.io/components/nsqd.html#statsd--graphite-integration) metrics via statsd, so Grafana agent/statsd_exporter required so get metrics into Prometheus.

This NSQ mixin interesting features:
1) Set of two dashboards:
NSQ Topics - view metrics grouped by topics and channels
screenshot-0
NSQ Instances - view metrics  grouped by nsqd instances
screenshot-1

2) Dashboard links to quickly jump from one dashboard(view) to another.
3) [Data links ](https://grafana.com/docs/grafana/latest/linking/data-links/#data-links) to contextually jump from instance view to topics view and vice versa
image
4) Alerts for topics and channel depth.
image

In order to test this mixin with live data, please use sample app in dev environment.


nsq-mixin/README.md
```diff
@@ -0,0 +1,156 @@
+# NSQ Mixin
```
rgeyer, line 1: Very thorough README, this should be the standard for all of our mixins. I now have something to aspire to ;)

nsq-mixin/dashboards/instances.libsonnet
```diff
@@ -0,0 +1,213 @@
...
+        title='Show split by channels for ${__field.labels.instance}',
+        url='d/nsq-topics?var-instance=${__field.labels.instance}&%s' % nsqChannelDataLinkQueryParams
+      )
+      .withUnit('ns'),
+
+    'nsq-instances.json':
+      grafana.dashboard.new(
+        '%s Instances' % $._config.dashboardNamePrefix,
+        time_from='%s' % $._config.dashboardPeriod,
+        editable=true,  // TODO change to false
```
rgeyer, line 111: Still a TODO?

nsq-mixin/dashboards/topics.libsonnet
```diff
@@ -0,0 +1,239 @@
...
+    local nsqSelector = 'job=~"$job", instance=~"$instance"',
+    local nsqTopicSelector = nsqSelector + ',topic=~"$topic"',
+    local nsqChannelSelector = nsqTopicSelector + ',channel=~"$channel"',
+    local nsqTopicDataLinkQueryParams = '${datasource:queryparam}&${job:queryparam}&$${__url_time_range}',
+    local nsqChannelDataLinkQueryParams = '${datasource:queryparam}&${job:queryparam}&${__url_time_range}',
+
+    local nsqTopics =
+      nsqgrafonnet.timeseries.new('Topic messages')
+      .addTarget(prometheus.target(expr='avg by (topic) (rate(nsq_topic_message_count{%s}[$__rate_interval]))' % nsqTopicSelector, intervalFactor=1, legendFormat='{{topic}} rps'))
+      .addTarget(prometheus.target(expr='sum by () (rate(nsq_topic_message_bytes{%s}[$__rate_interval]))' % nsqTopicSelector, intervalFactor=1, legendFormat='selected topics Bps rate'))
```
rgeyer, line 18: To my eye, this graph get's a bit difficult to understand, particularly with the mixed units.

If I understand this correctly, the left Y axis is rate of topics per second, while the right Y axis is the Bps rate of select queries?

Does this make more sense as two timeseries? Maybe a gauge for current select Bps, and a pie/donut graph for busiest topic?

nsq-mixin/lib/nsqgrafonnet/timeseries.libsonnet
```diff
@@ -0,0 +1,64 @@
+local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
+local prometheus = grafana.prometheus;
+
+{
+  new(title, description=null)::
```
rgeyer, line 5: This is great, I'll likely use it in some of my future mixins.

Do we anticipate building more of these for other newer visualizations?

Do we think that it makes sense to offer this upstream to grafonnet, or as a separate extension library?

rgeyer (approved):
Couple of small nits about formatting, but nothing that should block this if you do not agree with my comments.

Excellent work. Thank you!

nsq-mixin/lib/nsqgrafonnet/timeseries.libsonnet
v-zhuravlev: Even though there concerns about grafonnet lib(https://github.com/grafana/grafonnet-lib/issues/332), I think we should use some time to update grafonnet upstream.

nsq-mixin/dashboards/topics.libsonnet
v-zhuravlev: I know what you mean, but let's keep it as is for now. I made Bps shady as this metric is more for 'style'.

rgeyer: I'm good with this. We'll see if we get feedback from the field ;)

PR #723, opened 2022-01-12
Create library for loki simple-scalable deployment

trevorwhitney:
Ad library for deploy loki in simple-scalable mode


trevorwhitney:
replaced by https://github.com/grafana/loki/pull/5138

decided to put directly into loki repo as that's where the rest of the loki jsonnet is

PR #725, opened 2022-01-18
feat(postgres-exporter): add lib

Duologic:
Simple lib to configure and install https://github.com/prometheus-community/postgres_exporter/

### DIFF
--- a/postgres-exporter/main.libsonnet
+++ b/postgres-exporter/main.libsonnet
@@ -1,0 +1,43 @@
+local k = import 'ksonnet-util/kausal.libsonnet';
+
+{
+  new(
+    name,
+    data_source_name='$(USER):$(PASSWORD)@tcp($(HOST):$(PORT))/',
+    image='quay.io/prometheuscommunity/postgres-exporter:v0.10.0',
+  ):: {
+    local this = self,
+
+    local container = k.core.v1.container,
+    local containerPort = k.core.v1.containerPort,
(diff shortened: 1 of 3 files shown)


postgres-exporter/main.libsonnet
```diff
@@ -0,0 +1,43 @@
+local k = import 'ksonnet-util/kausal.libsonnet';
+
+{
+  new(
+    name,
+    data_source_name='$(USER):$(PASSWORD)@tcp($(HOST):$(PORT))/',
```
nikosmeds, line 6: First time seeing `@tcp($(HOST):$(PORT))/` - and not finding any references in https://github.com/prometheus-community/postgres_exporter/ or https://github.com/lib/pq.

Curious: how does this differ from `@$(HOST):$(PORT)/` ?

nikosmeds (approved):

postgres-exporter/main.libsonnet
Duologic: Good catch, I based this lib on mysqld_exporter, it parses `unix()` and `tcp()` bits in the data_source_name, I assumed this was standard net/url parser behavior but apparently it is not. Lemme fix.

Duologic:
```suggestion
    data_source_name='$(USER):$(PASSWORD)@$(HOST):$(PORT)/',
```

PR #731, opened 2022-01-21
Hass namespace prefix

v-zhuravlev:
This sets prefix template on the dashboard to 'homeassistant' by default.
image

Default is chosen according to
https://github.com/home-assistant/core/blob/dev/homeassistant/components/prometheus/__init__.py#L63

So, in fact `homeassistant_` namespace is used by default with minimalistic configuration.yaml config like
```yaml
prometheus:
```

...even though documentation states otherwise:
https://www.home-assistant.io/integrations/prometheus/#namespace

This PR also adds mixin makefile.

Hass version: homeassistant/home-assistant:2021.12.8

### DIFF
--- a/hass-mixin/dashboard.libsonnet
+++ b/hass-mixin/dashboard.libsonnet
@@ -5,6 +5,7 @@ local base_matcher = 'job=~"$job", instance=~"$instance"';
 local entity_matcher = base_matcher + ', entity=~"$entity", friendly_name=~"$friendly_name"';
 local available_entity_matcher_decoration = 'and on (entity) entity_available > 0';
 
+local prefix_default = 'homeassistant';
 local queries = {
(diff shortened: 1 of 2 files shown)


hass-mixin/dashboard.libsonnet
```diff
@@ -5,6 +5,7 @@ local base_matcher = 'job=~"$job", instance=~"$instance"';
 local entity_matcher = base_matcher + ', entity=~"$entity", friendly_name=~"$friendly_name"';
 local available_entity_matcher_decoration = 'and on (entity) entity_available > 0';
 
+local prefix_default = 'homeassistant';
```
rgeyer, line 8: Odd.. I thought I had already enabled this. I suppose when I built/tested I was always using a custom prefix.

Thanks for this contribution!

rgeyer (approved):
LGTM. Thanks!

PR #735, opened 2022-01-28
grizzly: remove .json/.yaml in the dashboards names

MrFreezeex:
Mixins typically append .json or .yaml at the end of the name.
Remove those so that Grizzly can deploy them like the jsonnet code embed
into Grizzly does (grizzly.jsonnet).

Signed-off-by: Arthur Outhenin-Chalandre <arthur.outhenin-chalandre@cern.ch>

### DIFF
--- a/grizzly/grafana.libsonnet
+++ b/grizzly/grafana.libsonnet
@@ -5,7 +5,12 @@ local util = import 'util.libsonnet';
   getFolder(main):: util.get(main, 'grafanaDashboardFolder', 'General'),
 
   fromMap(dashboards, folder):: {
-    [k]: util.makeResource('Dashboard', k, dashboards[k], { folder: folder })
+    [k]: util.makeResource(
+      'Dashboard',
+      std.strReplace(std.strReplace(std.strReplace(k, '.json', ''), '.yaml', ''), '.yml', ''),
+      dashboards[k],
+      { folder: folder }
+    )
     for k in std.objectFields(dashboards)
   },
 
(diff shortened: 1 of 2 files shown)


malcolmholmes:
Are you using `fromMixin`?

MrFreezeex:
Indirectly yes. I am using `fromPrometheusKsonnet` and I have all the mixins that I want in a `mixins::` so this lib would eventually call `fromMixin`.

MrFreezeex:
Hi @malcolmholmes, could we settle on something for this PR? It's literally a copy/paste of https://github.com/grafana/grizzly/blob/master/pkg/grizzly/grizzly.jsonnet#L8 and it would remove our internal hack to have grizzly working on mixins so I would really appreciate if this can be merged...

My other PR on grafana/grizzly (https://github.com/grafana/grizzly/pull/199) is not a blocker for us so if you don't have time/you don't want to have it that's fine (see the two last comments for context).

malcolmholmes:
Apologies for my delay.

This fix should not be made in the `makeResource` function, as this is not specific to the prometheus-ksonnet usecase.

The right place should be here:

https://github.com/grafana/jsonnet-libs/blob/857530e2e73565c40a3d06aaf8e2195a77cda55b/grizzly/grafana.libsonnet#L7-L10

e.g.:
```
fromMap(dashboards, folder):: {
    [k]: util.makeResource(
      'Dashboard',
      std.strReplace(std.strReplace(std.strReplace(k, '.json', ''), '.yaml', ''), '.yml', ''),
      dashboards[k],
      { folder: folder }
    )
    for k in std.objectFields(dashboards)
  }
```

If that works, I'll happily merge it!

MrFreezeex:
> Apologies for my delay.

No worries :).

> This fix should not be made in the `makeResource` function, as this is not specific to the prometheus-ksonnet usecase.
> 
> The right place should be here:
> 
> https://github.com/grafana/jsonnet-libs/blob/857530e2e73565c40a3d06aaf8e2195a77cda55b/grizzly/grafana.libsonnet#L7-L10

I did it in the `makeResource` function definition instead because there are several calls to this function that should be fixed (i.e.:  in `fromMap` and `fromMixins`).

malcolmholmes:
> Ah yes you are right thanks! I did it in the makeResource function definition instead because there are several calls to this function that should be fixed (i.e.: in fromMap and fromMixins).

Yeah, but the `makeResource` is intended as a "new generation" thing, whereas the `fromMap` and `fromMixins` are the "migration from legacy". So the fix you're talking about needs to be in the latter so as not to effect the former.

MrFreezeex:
> Yeah, but the `makeResource` is intended as a "new generation" thing, whereas the `fromMap` and `fromMixins` are the "migration from legacy". So the fix you're talking about needs to be in the latter so as not to effect the former.

Ah ok I see. I added those in every fromMap/fromMixins call then!

malcolmholmes (approved):

malcolmholmes:
thanks for persisting!

PR #738, opened 2022-02-02
Add support for specifying MySQL TLS mode to mysqld-exporter

radiohead:
### What

Add support for specifying MySQL TLS mode to `mysqld-exporter`. By default the mode is set to `preferred`, which provides highest compatibility level, i.e. it should work for both SSL and non-SSL setups.

More details on the mode available here - https://github.com/go-sql-driver/mysql#tls.

### Why

Currently there's no way to specify TLS mode without overriding `DATA_SOURCE_NAME` completely and for setups where SSL is required by the server that's the only solution.

### DIFF
--- a/mysqld-exporter/main.libsonnet
+++ b/mysqld-exporter/main.libsonnet
@@ -9,6 +9,7 @@ local container = k.core.v1.container;
     user,
     host,
     port=3306,
+    tlsMode='preferred',
     image='prom/mysqld-exporter:v0.13.0',
   ):: {
     local this = self,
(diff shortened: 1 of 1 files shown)


mysqld-exporter/main.libsonnet
```diff
@@ -9,6 +9,7 @@ local container = k.core.v1.container;
     user,
     host,
     port=3306,
+    tlsMode='preferred',
```
Duologic, line 12: Might be better to put this arg last in the list as some may be using unnamed arts and have the image name it this place.

Duologic (approved):

mysqld-exporter/main.libsonnet
radiohead: Fair point! I'll do that.

PR #742, opened 2022-02-07
Adding Kafka Alerts

gaantunes:
Adding some alerts to the Mixin, based on the ones available at https://github.com/dcos/prometheus-alert-rules/blob/master/rules/kafka/kafka-rules.yml

### DIFF
--- a/kafka-mixin/alerts/KafkaAlerts.yml
+++ b/kafka-mixin/alerts/KafkaAlerts.yml
@@ -1,0 +1,74 @@
+groups:
+- name: Kafka_Alerts
+  rules:
+    - alert: Offline_Partiton_Count
+      expr: sum(kafka_controller_KafkaController_OfflinePartitionsCount{job="integrations/kafka"})by(instance)  > 0
+      for: 1m
+      labels:
+        severity: critical
+      annotations:
+        summary: 'Kafka {{ $labels.instance }}: {{ $value }} partitons offline'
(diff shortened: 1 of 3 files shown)


v-zhuravlev (changes requested):
I think it is better to name alerts using CamelCase, as suggested in mixin gudelines:

https://monitoring.mixins.dev/#guidelines-for-alert-names-labels-and-annotations

kafka-mixin/alerts/KafkaAlerts.yml
```diff
@@ -0,0 +1,74 @@
+groups:
+- name: Kafka_Alerts
+  rules:
+    - alert: Offline_Partiton_Count
+      expr: sum(kafka_controller_KafkaController_OfflinePartitionsCount{job="integrations/kafka"})by(instance)  > 0
```
v-zhuravlev, line 5: whitespace fmt
```suggestion
      expr: sum(kafka_controller_KafkaController_OfflinePartitionsCount{job="integrations/kafka"}) by (instance) > 0
```

v-zhuravlev (commented):
Also, it seems that it may better to swap text in `details` and `summary`, as 
it is suggested that 
summary should be without templating, but details can be with templates:
https://monitoring.mixins.dev/#guidelines-for-alert-names-labels-and-annotations

v-zhuravlev (approved):
I added some more spaces between and after 'by', hope you don't object :)

PR #746, opened 2022-02-15
refactor(postgres-exporter): use DATA_SOURCE_URI

Duologic:
Breaking change, the DATA_SOURCE_NAME is marked as legacy, it also doesn't work well when
non-ascii characters are used in username and password. DATA_SOURCE_URI does support
that but requires a slightly different approach.

### DIFF
--- a/postgres-exporter/main.libsonnet
+++ b/postgres-exporter/main.libsonnet
@@ -3,7 +3,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
 {
   new(
     name,
-    data_source_uri='postgresql://@$(HOSTNAME):$(PORT)/postgres',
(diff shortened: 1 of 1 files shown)


postgres-exporter/main.libsonnet
```diff
@@ -3,7 +3,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
 {
   new(
     name,
-    data_source_name='postgresql://$(USER):$(PASSWORD)@$(HOST):$(PORT)/postgres',
+    data_source_uri='postgresql://@$(HOSTNAME):$(PORT)/postgres',
```
inkel, line 6: Should the `@` still be present in the URI if there are no username and password components?

inkel (approved):

PR #747, opened 2022-02-15
fix(postgres-exporter): correct use of DATA_SOURCE_URI

Duologic:
apparently we need to drop the prefix


inkel (approved):
That is really odd, it stops being an URI in that case 🤔

Duologic:
🤷‍♂️

PR #750, opened 2022-03-03
Allow passing volume mixin to hostVolumeMount()

iainlane:
There are some fields you might want to set on the created volume, for
example `hostPath` volumes can have various [types][0]. It's not that
easy currently to pass these through when using `hostVolumeMount()`, so
add a new kwarg to pass a mixin to the created `volumeMount`.

[0]: https://kubernetes.io/docs/concepts/storage/volumes/#hostpath[](https://github.com/iainlane)

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -162,7 +162,7 @@ local util(k) = {
     ])
     + (if annotations != {} then deployment.mixin.spec.template.metadata.withAnnotationsMixin(annotations) else {}),
 
-  hostVolumeMount(name, hostPath, path, readOnly=false, volumeMountMixin={})::
+  hostVolumeMount(name, hostPath, path, readOnly=false, volumeMountMixin={}, volumeMixin={})::
     local container = k.core.v1.container,
           deployment = k.apps.v1.deployment,
           volumeMount = k.core.v1.volumeMount,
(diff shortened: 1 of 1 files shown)


ksonnet-util/util.libsonnet
```diff
@@ -174,7 +174,8 @@ local util(k) = {
 
     deployment.mapContainers(addMount) +
     deployment.mixin.spec.template.spec.withVolumesMixin([
-      volume.fromHostPath(name, hostPath),
+      volume.fromHostPath(name, hostPath) +
+      if type != '' then volume.hostPath.withType(type) else {},
```
jvrplmlmn, line 178: Given that empty string is the default, would it make sense to remove the conditional logic?

Maybe the question is, if we make that change, does that generated a lot of diff and is something that we are trying to prevent?

iainlane: Ah yeah, I could have said - I tried it like that and a lot of my call sites ended up having `type: ""` added :(

jjo: Indeed I think we need to keep the logic, without assuming implementation's default while also not polluting with "nil" values.

jjo (approved):
LGTM

ksonnet-util/util.libsonnet
jvrplmlmn: @iainlane and I jumped on a call before and my suggestion was to provide a new mixin instead. Something like:

```
hostVolumeMount(name, hostPath, path, readOnly=false, volumeMountMixin={}, volumeMixin={})::
(...)
```

That gives the caller more flexibility, and we are already doing it for the volumeMount

iainlane:
I've updated as requested based on the conversation & have checked it works with code like

```
         k.util.hostVolumeMount('foo',
                                '/foo,
                                '/foo',
                                readOnly=false,
                                volumeMixin=volume.hostPath.withType('DirectoryOrCreate'))
```

Cheers, this is better I think, or at least it's consistent with what the function was already doing.

jvrplmlmn (approved):

PR #755, opened 2022-03-30
Fix CI (+ CircleCI -> Drone)

julienduchesne:
CI is currently on CircleCI and versions are outdated
It also depends on a custom built image that's pushed manually, this PR puts everything in the Makefile so it's easier to manage and moves it to Drone


Makefile
```diff
@@ -1,29 +1,45 @@
...
+	go install github.com/monitoring-mixins/mixtool/cmd/mixtool@ae18e31161ea
+	go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@v0.4.0
+
 fmt:
-		@find . -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
-				xargs -n 1 -- $(JSONNET_FMT) -i
+	@find . -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
+			xargs -n 1 -- $(JSONNET_FMT) -i
+
+lint-fmt:
```
iainlane, line 15: nit: maybe these targets could depend on `install-ci-deps` to help people running locally?

iainlane (approved):

Makefile
julienduchesne: It's a bit rude to run `go install` on people's machines, isn't it? 😆  I'll leave it as-is for now

iainlane: fair enough

PR #756, opened 2022-03-31
Configurable memcached cpu requests

colega:
For small deployments just experimenting with Mimir, 500m is too much for a memcached, but it's not configurable right now so one has to completely override kuberentes requests.

This makes it configurable, just like `memory_requests_mb` is.


pstibrany (approved):
Looks useful.

PR #757, opened 2022-04-04
Replace rbac.authorization.k8s.io/v1beta1 with v1

ivaanko:
rbac.authorization.k8s.io/v1beta1 is removed in k8s 1.22 https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122

### DIFF
--- a/prometheus/prometheus.libsonnet
+++ b/prometheus/prometheus.libsonnet
@@ -18,7 +18,7 @@ local kausal = import 'ksonnet-util/kausal.libsonnet';
           servicePort: kausal.core.v1.service.mixin.spec.portsType,
         } },
         rbac+: { v1+: {
-          policyRule: kausal.rbac.v1beta1.clusterRole.rulesType,
+          policyRule: kausal.rbac.v1.clusterRole.rulesType,
         } },
       }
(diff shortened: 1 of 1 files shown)


xvzf:
Makes it compatible with k8s >= 1.22 👍🏻

Duologic:
This path should not be triggered unless `std.objectHas(kausal, '__ksonnet')`. Are you still using `ksonnet-lib`? Ksonnet hasn't been updated for using with 1.22, I'd strongly discourage using ksonnet-lib, [k8s-libsonnet](https://github.com/jsonnet-libs/k8s-libsonnet) can be used instead.

ivaanko:
Nope we're using `k8s-libsonnet`. Could the path be triggered by some back-compat hacks in `ksonnet-util` or `prometheus` libraries?

Duologic:
Can you grep for `__ksonnet`?

xvzf:
Sorry did not check properly; the v1beta1 reference was not in prometheus but in kube-state-metrics; fix for it here: https://github.com/grafana/jsonnet-libs/pull/757

PR #760, opened 2022-04-06
enterprise-metrics 2.0.0

Logiraptor:
Update enterprise-metrics for the GEM 2.0.0 release (PENDING)

- Update to mimir jsonnet as baseline
- Remove properties set to `null` where no longer necessary
- Tested upgrade from previous jsonnet and GEM 1.7


dimitarvdimitrov (approved):
:shipit:

PR #764, opened 2022-04-11
Upgrade oauth2-proxy version: 7.1.3 -> 7.2.1

julienduchesne:
Also, add an extra args config so that we can set a custom logo and disable the footer


zzehring (approved):
LGTM

Issue #753, opened 2022-03-29
Ceph mixin issues


RutgerKe:
Let me start with: Awesome mixin, it really works well to get an overview of the cluster state, and with debugging issues! There are however a few problems with it.

Context: I'm running the Ceph integration in Grafana Cloud (which I beleive is based on this mixin), my cluster is created with proxmox and consits of three nodes. I'm scraping it from a VM running on one of the nodes.

A few things which are broken:
- A bunch of panels use a `cluster` variable, while only a `ceph_cluster` variable is available. This breaks quite a few panels in the dashboard. Both the dashboard variable and the target 'label name' should change from `cluster` to `ceph_cluster`
- Node statistics don't work, they mix the `cluster` and `ceph_cluster`, but more importantly they use information not coming directly from ceph. For this `node_exporter` is required to be available and configured with a label, which is not documented. I think we should either remove the panels or document the extra dependency.
- The health alert that comes with the mixin fires immediately and keeps firing: `absent(ceph_health_status == 1)` ,  (possible values for that metric: healty=0, warning=1, error=2). So it should only fire if there is a 2 probably.

A few things which might be a little more subjective, but I think these changes will help people:
- We shouldn't be stacking in graphs. While it might sometimes look nice, it basically makes it so you can only see the total in each graph. Two examples:
	- What is the number of objects per pool. Did one pool drop, or did they all?
	- Considering the throughput: 'write looks as high as read!' But no, write is almost 0, but stacked on top of read.
- PGs state (remapped etc) have a log scale that starts at y-min: 0, which is mathematically weird and looks weird with the old graph (unused space) , and rightout fails when converting to the new timeseries panel.

PR #766, opened 2022-04-14
Ceph-mixin: Bugfixes

gaantunes:
This PR implements some bugfixes on the mixin based on @RutgerKe feedback at https://github.com/grafana/jsonnet-libs/issues/753.

- Removing all `cluster` variable in favor of `ceph_cluster`
- Removing node_exporter based panels
- Migrating from graph panel to time series panels


RutgerKe (approved):
I loaded the json in my stack, and it looks great! Thanks :smile:

PR #768, opened 2022-04-14
Shahzeb/docker add logs dashboard

mshahzeb:
- Adds docker logs dashboard
- Minor fixes in docker overview dashboard
- Adds config.jsonnet


mshahzeb:
Logs dashboard screenshot.
docker_logs_1

docker-mixin/config.libsonnet
```diff
@@ -0,0 +1,8 @@
+{
+  _config+:: {
+    dashboardTags: ['docker-integration'],
+    dashboardPeriod: 'now-1h',
+    dashboardTimezone: 'default',
+    dashboardRefresh: '1m',
+  },
```
rgeyer, line 7: Just as a heads-up, these things are all part of the "prepare dashboards" functionality that is executed against all mixins when they become grafana cloud integrations.

It's fine to have these templatized for non grafana cloud users, tho I would probably choose a tag other than `docker-integration`. Perhaps just `docker`?

docker-mixin/dashboards/docker-logs.libsonnet
```diff
@@ -0,0 +1,280 @@
...
+  'job',
+  '$loki_datasource',
+  'label_values(job)',
+  label='Job',
+  refresh='load',
+  multi=true,
+  includeAll=true,
+  allValues='.+',
+  sort=1,
+  regex='.*docker.+'
```
rgeyer, line 51: Is this necessary? Perhaps too opinionated, since users should be free to name their jobs whatever they like, and still find them here.

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
```diff
@@ -0,0 +1,162 @@
+local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
+
+{
+  new(q1, q2, q3)::
```
rgeyer, line 4: This is very domain specific, which seems intentional given the name `custom-barchart`.

It *might* make sense to allow more queries to be added, and you can use the `addTarget` [pattern which exists in grafonnet](https://github.com/grafana/grafonnet-lib/blob/master/grafonnet/bar_gauge_panel.libsonnet#L39-L45).

Not a necessary change, just a suggestion.

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
```diff
@@ -0,0 +1,162 @@
...
+                value: {
+                  fixedColor: 'red',
+                  mode: 'fixed',
+                },
+              },
+            ],
+          },
+        ],
+      },
+      gridPos: {
```
rgeyer, line 103: I suppose the expectation is that he gridpos will always be overriden? Perhaps best to exclude it from this definition.

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
```diff
@@ -0,0 +1,162 @@
...
+        h: 0,
+        w: 0,
+        x: 0,
+        y: 0,
+      },
+      id: 12,
```
rgeyer, line 109: If this gets used in another dashboard, or even if it's order in the existing dashboard gets changed, the ID could end up conflicting.

Best to exclude this as well, and count on the `grafana.dashboard.addPanel` to add the ID.

rgeyer (changes requested):
Looks good, a few nits.

One thing you should resolve is the static ID on your custom barchart panel, that could easily cause some weird issues after subsequent changes.

Good work!

docker-mixin/config.libsonnet
mshahzeb: Thanks! Changed to "docker"

docker-mixin/dashboards/docker-logs.libsonnet
mshahzeb: Yup removed the regex based on our discussion

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
mshahzeb: Good suggestion! Letting this go for now - Will keep in mind next time.

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
mshahzeb: Yup excluded grid pos and id

docker-mixin/lib/custom-barchart-grafonnet/custom-barchart.libsonnet
mshahzeb: Thanks for the catch! Removed the id.

rgeyer (approved):
LGTM, thanks for the quick changes.

mshahzeb:
Thanks! Merging

PR #770, opened 2022-04-22
Add `k.util.secretVolumeMountAnnotated`

colega:
We have configMapVolumeMount that adds config map annotations to the
deployments to ensure that pods are restarted when the contents of the
config map changes. Why not having the same for secrets?

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -209,6 +209,29 @@ local util(k) = {
       volume.mixin.secret.withDefaultMode(defaultMode),
     ]),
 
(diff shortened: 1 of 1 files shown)


colega:
@Duologic asking you for a review as Github suggested that :)

Duologic:
I have been trying to upstream these into the k8s-libsonnet library, would you be interested in contributing it here: https://github.com/jsonnet-libs/k8s/blob/master/libs/k8s/custom/core/volumeMounts.libsonnet#L123 ?

Duologic:
@colega I'm porting it into k8s-libsonnet here: https://github.com/jsonnet-libs/k8s/pull/181

colega:
Thanks, I haven't had time to work on this. I guess we can close this PR now.

PR #771, opened 2022-04-25
Allow kube-state-metrics' 'self metrics' to be scraped using service discovery

iainlane:
kube-state-metrics exposes metrics on two ports:

  - Metrics it gathers about the cluster it is monitoring (port `http-metrics`)
  - 'Self' metrics about itself (port `self-metrics`)

Currently we scrape both of these targets in the same scrape config and deploy
k-s-m with an annotation `prometheus.io.scrape=false` so normal service
discovery doesn't look at it. This is because the standard relabeling would
override the labels to those of k-s-m itself rather than the services being
monitored (e.g. all metrics would get the 'kube-state-metrics' namespace).

This standard service discovery config shipped with jsonnet-libs scrapes all
ports that end with `-metrics`. An alternative approach then would be to rename
the `http-metrics` port to something else, so it is not found by the normal
scrape config, and drop the `prometheus.io.scrape=false` annotation. Then we
can configure k-s-m's scrape config to only look at this port.

### DIFF
--- a/kube-state-metrics/main.libsonnet
+++ b/kube-state-metrics/main.libsonnet
@@ -21,7 +21,7 @@ local kausal = import 'github.com/grafana/jsonnet-libs/ksonnet-util/kausal.libso
         '--telemetry-port=8081',
       ])
       + container.withPorts([
-        containerPort.new('http-metrics', 8080),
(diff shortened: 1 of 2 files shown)


jvrplmlmn (approved):
👍

PR #774, opened 2022-04-29
Allow datasource's regex to be configured

wilfriedroset:
With a lot of datasource it is handy to be able to list only on part of
the prometheus datasource available in the organization.

Signed-off-by: Wilfried Roset <wilfriedroset@users.noreply.github.com>


colega (approved):

colega:
Thank you!

PR #775, opened 2022-05-04
Allow tooltip to be configured

wilfriedroset:
Signed-off-by: Wilfried Roset <wilfriedroset@users.noreply.github.com>

### DIFF
--- a/grafana-builder/grafana.libsonnet
+++ b/grafana-builder/grafana.libsonnet
@@ -178,7 +178,7 @@
     titleSize: 'h6',
   },
 
-  panel(title):: {
+  panel(title, tooltip_shared=true, tooltip_sort=2, tooltip_value_type='individual'):: {
     aliasColors: {},
     bars: false,
     dashLength: 10,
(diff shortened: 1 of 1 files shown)


grafana-builder/grafana.libsonnet
```diff
@@ -1,5 +1,5 @@
 {
-  dashboard(title, uid='', datasource='default', datasource_regex=''):: {
+  dashboard(title, uid='', datasource='default', datasource_regex='', tooltip_shared=true, tooltip_sort=2, tooltip_value_type='individual'):: {
```
pracucci, line 2: I have mixed feelings about this change. The tooltip is not a property of the dashboard, but of each individual panel. It shouldn't be part of dashboard settings.

wilfriedroset: Indeed, however without this patch users will not be able to update the libsonnet.
Example: https://github.com/grafana/mimir/blob/main/operations/mimir-mixin/jsonnetfile.lock.json#L7
The next time `jb update` is run the user will have the new version with `tooltip_shared=true` enable which is not a wishable behavior as you point:

> I just tried the shared tooltip on these dashboards in a large cluster and unfortunately they're unusable. All shared tooltips cover all panels. I would revert back this change, getting back to non shard tooltips.

pracucci:
I still believe it shouldn't be an option of the dashboard, but a `panel()` one. WDYT?

wilfriedroset:
> I still believe it shouldn't be an option of the dashboard, but a `panel()` one. WDYT?
I've moved the tooltip varialbes at the panel definition.

PR #778, opened 2022-05-12
Add apache-http-mixin

v-zhuravlev:
Based on the dashboard from here https://github.com/rfrail3/grafana-dashboards. (https://grafana.com/grafana/dashboards/3894). With @rfrail3 we have decided that this dashboard can land here, and transformed into monitoring-mixin. Future updates would be in this repo. 

As of now, this mixin would generate the same dashboard , but with slight changes (see below and in commit messages). Prometheus alerts are also added. Just run make and follow the README file.

image

- Add stepAfter for Apache worker statuses panel
- Add stepAfter for apache scorecard
- Rename apache dashboard
- Add apache version panel
- Convert to timeseries from old graph
- Change rate panels to Bps and reqps
- Use rate with $__rate_inteval
- Fix uptime panel
- Change host:port to instance and replace apache up/down panel to state timeline
- Add alerts to apache mixin
- Transform dashboard to use grafonnet library
- Rename DS_LOCALHOST to prometheus $datasource
- Convert apache json dahboard to jsonnet mixin

### DIFF
--- a/apache-http-mixin/dashboards/apache-exporter-full.libsonnet
+++ b/apache-http-mixin/dashboards/apache-exporter-full.libsonnet
@@ -1,6 +1,8 @@
 local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
 local dashboard = grafana.dashboard;
 local template = grafana.template;
+local dashboardUid = 'apache-http';
+local matcher = 'job=~"$job", instance=~"$instance"';
 
 {
   grafanaDashboards+:: {
@@ -14,6 +16,7 @@ local template = grafana.template;
         timezone='%s' % $._config.dashboardTimezone,
         refresh='%s' % $._config.dashboardRefresh,
         graphTooltip='shared_crosshair',
+        uid=dashboardUid,
       ).addTemplates(
         [
           {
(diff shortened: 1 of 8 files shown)


apache-http-mixin/dashboards/apache-exporter-full.libsonnet
```diff
@@ -0,0 +1,777 @@
...
+      ).addTemplates(
+        [
+          {
+            current: {
+              text: 'default',
+              value: 'default',
+            },
+            hide: 0,
+            label: 'Data Source',
+            name: 'datasource',
```
mshahzeb, line 29: Should we use prometheus_datasource instead of datasource so we can then use loki_datasource for logs later?

mshahzeb (approved):
LGTM! Great addition to our integrations.

Left 1 comment on datasource naming.

apache-http-mixin/dashboards/apache-exporter-full.libsonnet
rgeyer: The linting rule will *prefer* `datasource` when only one datasource is defined. It's an easy enough search and replace to change when logs are implemented.

v-zhuravlev: thanks for noticing. Yes, let's keep as datasource for now, and we can switch to prometheus_datasource later.

PR #780, opened 2022-05-15
spark-mixin: fix some job name

MrFreezeex:
Some queries didn't have the `integrations/` prefix in the job label.


gaantunes (approved):
Thanks for the fix! Will be merging it.

Issue #726, opened 2022-01-19
Consolidate postgres-mixin


Duologic:
The postgres-mixin has a place in [upstream](https://github.com/prometheus-community/postgres_exporter/tree/master/postgres_mixin), should we consider removing it from grafana/jsonnet-libs before it diverges too much?

cc @kavirajk @jeschkies @tomwilkie

tomwilkie:
Yes please!

kavirajk:
Sure. Thanks @Duologic

jeschkies:
Sure!

PR #783, opened 2022-05-18
Postgres-mixin: Bugfixes and linting

rgeyer:
Some changes to accommodate the dashboard linter, caught several minor (or moderate?) issues.

- Most (all?) queries were not using the `$job` matcher
- The template vars were set with default selected values, which weren't ideal
- The instance template var was not filtering to instances belonging to the selected job(s)

### DIFF
--- a/postgres-mixin/dashboards/postgres-overview.json
+++ b/postgres-mixin/dashboards/postgres-overview.json
@@ -584,7 +584,7 @@
         {
           "alias": "Buffers Allocated",
           "dsType": "prometheus",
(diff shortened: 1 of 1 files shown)


v-zhuravlev:
What do you think, should we switch to upsteam?
https://github.com/grafana/jsonnet-libs/issues/726

rgeyer:
> What do you think, should we switch to upsteam? #726

Ohh yes! Good call.. I totally spaced on that. I'll offer these changes as an upstream PR instead. :)

rgeyer:
Closing in favor of prometheus-community/postgres_exporter#647

PR #790, opened 2022-05-30
Parse windows logs

mshahzeb:
Parses windows logs in json and shows log lines in a formatted human readable form

### DIFF
--- a/windows-mixin/dashboards/windows_logs.libsonnet
+++ b/windows-mixin/dashboards/windows_logs.libsonnet
@@ -288,7 +288,7 @@ local log_full_panel =
         job_template,
         host_template,
         channel_template,
-        source_template
+        source_template,
       ])
 
       .addLink(grafana.link.dashboards(
--- a/windows-mixin/dashboards/windows_logs.libsonnet
+++ b/windows-mixin/dashboards/windows_logs.libsonnet
@@ -4,6 +4,7 @@ local custom_barchart_grafonnet = import '../lib/custom-barchart-grafonnet/custo
 
 local host_matcher = 'job=~"$job", agent_hostname=~"$hostname"';
 local log_channel_matcher = host_matcher + ', channel=~"$channel"';
+local windows_event_parser = '| json | line_format "{{.execution_processId}} Source: {{.source}} EventID: {{.event_id}} Level: {{.levelText}} . {{.message}}"';
 
 local queries = {
   total_log_lines: 'sum(count_over_time({' + log_channel_matcher + '}[$__interval]))',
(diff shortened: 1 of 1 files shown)


mshahzeb:
Screenshot

windows_logs_1

windows-mixin/dashboards/windows_logs.libsonnet
```diff
@@ -4,16 +4,17 @@ local custom_barchart_grafonnet = import '../lib/custom-barchart-grafonnet/custo
 
 local host_matcher = 'job=~"$job", agent_hostname=~"$hostname"';
 local log_channel_matcher = host_matcher + ', channel=~"$channel"';
+local windows_event_parser = '| json | line_format "{{.execution_processId}} Source: {{.source}} EventID: {{.event_id}} Level: {{.levelText}} . {{.message}}"';
 
 local queries = {
   total_log_lines: 'sum(count_over_time({' + log_channel_matcher + '}[$__interval]))',
   total_log_warnings: 'sum(count_over_time({' + log_channel_matcher + '} |= "Warning" [$__interval]))',
   total_log_errors: 'sum(count_over_time({' + log_channel_matcher + '} |= "Error" [$__interval]))',
   error_percentage: 'sum( count_over_time({' + log_channel_matcher + '} |= "Error" [$__interval]) ) / sum( count_over_time({' + log_channel_matcher + '} [$__interval]) )',
```
v-zhuravlev, line 13:
```suggestion
  error_percentage: 'sum(count_over_time({' + log_channel_matcher + '} |= "Error" [$__interval]) ) / sum( count_over_time({' + log_channel_matcher + '} [$__interval]) )',
```
Remove to match other lines style?

windows-mixin/dashboards/windows_logs.libsonnet
```diff
@@ -4,16 +4,17 @@ local custom_barchart_grafonnet = import '../lib/custom-barchart-grafonnet/custo
 
 local host_matcher = 'job=~"$job", agent_hostname=~"$hostname"';
 local log_channel_matcher = host_matcher + ', channel=~"$channel"';
+local windows_event_parser = '| json | line_format "{{.execution_processId}} Source: {{.source}} EventID: {{.event_id}} Level: {{.levelText}} . {{.message}}"';
```
v-zhuravlev, line 7: Two questions, just some thoughts:
1) WDYT about not using dot as separator?
2) I suggest to either add label to processId field or remove processId altogether. Since all extracted fields you don't use while composing line_format would be still available when you click on the line.

v-zhuravlev (commented):
Great addition! Left some small suggestions.

I also have question regarding extracted fields.

I assume you also setup logs parsing on the agent side, since you have fields with suffix _extracted. 

1) WDYT to you use label such as Source as Dashboard filter, so users can narrow down logs search quickly? I used such filter on Linux & Mac dashboards.
2) If you agree on 1) , then it would be great to have README placed for this mixin here with instructions how agent/promtail should be setup in order to receive windows logs with labels as well, so mixin can be used independently of cloud integration. (again, apache mixin could be an example of this)

windows-mixin/dashboards/windows_logs.libsonnet
mshahzeb: @v-zhuravlev Generally in logs thread ID or process ID is usually present that is why I kept it there. I will add a label to it.

Also what do you suggest we use for separator instead of dot?

mshahzeb:
@v-zhuravlev I did not setup log parsing on agent side yet - But yes I was trying to find out how I can add Source as a template variable. For that you are right I would need to process the log on agent side as well.

This is the correct link for doing log parsing in agent config - Correct?
https://grafana.com/docs/loki/latest/clients/promtail/stages/json/

windows-mixin/dashboards/windows_logs.libsonnet
v-zhuravlev: just to be clear. By label here I meant 'text label' like 'Process id: {{.execution_processId}} '. Not Logs stream label.

mshahzeb:
Latest screenshot with source variable added.

Screenshot 2022-05-31 at 7 45 55 PM

PR #793, opened 2022-06-01
nginx-directory: don't accept gzip when subfiltering

colega:
When using subfilter, we need the response to be non-gzipped in order to
be able to inspect the body.

Ref: https://til.hashrocket.com/posts/7ktzgumkmp-subfilter-proxypass-requires-no-gzip-encoding

### DIFF
--- a/nginx-directory/configfile.libsonnet
+++ b/nginx-directory/configfile.libsonnet
@@ -33,13 +33,14 @@ function(services) {
     + (
       if service.subfilter
       then |||
+        proxy_set_header Accept-Encoding "";
         sub_filter 'href="/' 'href="/%(path)s/';
         sub_filter 'src="/' 'src="/%(path)s/';
         sub_filter 'action="/' 'action="/%(path)s/';
         sub_filter 'endpoint:"/' 'endpoint:"/%(path)s/';  # for XHRs.
         sub_filter 'href:"/v1/' 'href:"/%(path)s/v1/';
         sub_filter_once off;
-        sub_filter_types %(rendered_subfilter_content_types)s;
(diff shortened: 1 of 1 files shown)


nginx-directory/configfile.libsonnet
```diff
@@ -33,13 +33,14 @@ function(services) {
...
       then |||
+        proxy_set_header Accept-Encoding "";
         sub_filter 'href="/' 'href="/%(path)s/';
         sub_filter 'src="/' 'src="/%(path)s/';
         sub_filter 'action="/' 'action="/%(path)s/';
         sub_filter 'endpoint:"/' 'endpoint:"/%(path)s/';  # for XHRs.
         sub_filter 'href:"/v1/' 'href:"/%(path)s/v1/';
         sub_filter_once off;
-        sub_filter_types %(rendered_subfilter_content_types)s;
+        sub_filter_types %(rendered_subfilter_content_types)s';
```
pracucci, line 43: Is this correct? Shouldn't have a single quote also at the beginning of the "value"?

pracucci (approved):
Approving to unblock you, but I'm not the best person to review it. If you don't feel comfortable about this change, please wait the approval of someone more experienced.

nginx-directory/configfile.libsonnet
colega: This change shouldn't be there, going to revert it.

colega: Removed it in https://github.com/grafana/jsonnet-libs/pull/793/commits/a4f1cd87c7e2c3bebe1aaf761edefc5c1d5ed09c

PR #803, opened 2022-06-08
Consul-mixin: Improvements from linting

rgeyer:
Make job template var multi-select
Aggregate server queries by instance
Use `$__rate_interval throughout`

### DIFF
--- a/consul-mixin/dashboards.libsonnet
+++ b/consul-mixin/dashboards.libsonnet
@@ -15,13 +15,13 @@ local panel_settings = {
   grafanaDashboards+:: {
     'consul-overview.json':
       g.dashboard('Consul Overview', std.md5('20210205-consul'))
-      .addTemplate('job', 'consul_up', 'job')
-      .addMultiTemplate('instance', 'consul_up{job="$job"}', 'instance')
+      .addMultiTemplate('job', 'consul_up', 'job')
+      .addMultiTemplate('instance', 'consul_up{job=~"$job"}', 'instance')
       .addRow(
         g.row('Up')
         .addPanel(
           g.panel('$instance') +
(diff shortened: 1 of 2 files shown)


consul-mixin/dashboards.libsonnet
```diff
@@ -82,14 +82,14 @@ local panel_settings = {
...
           g.stack
         )
         .addPanel(
           g.panel('Latency') +
-          g.queryPanel('max(consul_http_request{job="$job", quantile="0.99"})', '99th Percentile') +
-          g.queryPanel('max(consul_http_request{job="$job", quantile="0.5"})', '50th Percentile') +
-          g.queryPanel('sum(rate(consul_http_request{job="$job"}[5m])) / sum(rate(consul_http_request{job="$job"}[5m]))', 'Average') +
+          g.queryPanel('max(consul_http_request{job=~"$job", quantile="0.99"}) by (instance)', '99th Percentile', '{{instance}}') +
+          g.queryPanel('max(consul_http_request{job=~"$job", quantile="0.5"}) by (instance)', '50th Percentile', '{{instance}}') +
+          g.queryPanel('sum(rate(consul_http_request{job=~"$job"}[$__rate_interval])) by (instance) / sum(rate(consul_http_request{job=~"$job"}[$__rate_interval])) by (instance)', 'Average') +
```
rgeyer, line 92: @tomwilkie This query was originally written by you. I've changed it to aggregate by instance (like the QPS query), and to use `$__rate_interval`. However, it looks like this is EXACTLY the same value, divided by itself, which will always be 1.

I'm not sure if this was meant to be two different metrics, like the query duration, divided by the query count, or if maybe an offset range was intended?

In any case, I think this query still needs to change. Can you provide insight?

Where do these metrics come from? They aren't in the consul exporter 🤔

PR #810, opened 2022-06-20
Remove the usage of RBAC v1beta1 APIs.

mplzik:
IIUC these have been deprecated since k8s 1.17 and have been removed in
1.22 . The changelog mentions no significant API changes, apart from the
versuin change itself.

This PR right now keeps some references to v1beta1 -- those seem to be
mostly wrappers. I'll be happy to extend this PR with removal of these
-- currently we're well beyond any officially supported k8s version that
doesn't support the v1 RBAC API.

### DIFF
--- a/ksonnet-util/legacy-types.libsonnet
+++ b/ksonnet-util/legacy-types.libsonnet
@@ -18,6 +18,7 @@
       policyRule:: $.rbac.v1.clusterRole.rulesType,
       subject:: $.rbac.v1.clusterRoleBinding.subjectsType,
     },
+    // TODO: the v1beta1 RBAC API has been removed in Kubernetes 1.22 and should get removed once 1.22 is the oldest supported version
     v1beta1+: {
       policyRule:: $.rbac.v1.clusterRole.rulesType,
       subject:: $.rbac.v1.clusterRoleBinding.subjectsType,
--- a/ksonnet-util/legacy-types.libsonnet
+++ b/ksonnet-util/legacy-types.libsonnet
@@ -15,12 +15,12 @@
   },
   rbac+: {
     v1+: {
-      policyRule:: $.rbac.v1beta1.clusterRole.rulesType,
-      subject:: $.rbac.v1beta1.clusterRoleBinding.subjectsType,
+      policyRule:: $.rbac.v1.clusterRole.rulesType,
+      subject:: $.rbac.v1.clusterRoleBinding.subjectsType,
     },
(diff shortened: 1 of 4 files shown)


ksonnet-util/legacy-types.libsonnet
```diff
@@ -15,12 +15,12 @@
...
-      policyRule:: $.rbac.v1beta1.clusterRole.rulesType,
-      subject:: $.rbac.v1beta1.clusterRoleBinding.subjectsType,
+      policyRule:: $.rbac.v1.clusterRole.rulesType,
+      subject:: $.rbac.v1.clusterRoleBinding.subjectsType,
     },
     v1beta1+: {
-      policyRule:: $.rbac.v1beta1.clusterRole.rulesType,
-      subject:: $.rbac.v1beta1.clusterRoleBinding.subjectsType,
+      policyRule:: $.rbac.v1.clusterRole.rulesType,
+      subject:: $.rbac.v1.clusterRoleBinding.subjectsType,
```
jvrplmlmn, line 23: Given that this refers to v1beta1, should this remain unchanged? If you are using `k.rbac.v1beta1` you could migrate to `k.rbac.v1` instead (which you are already patching).

mplzik: The expected end result here is that the users of these libraries should be able to easily migrate to to Kubernetes API 1.22+, which doesn't expose the v1beta1 API at all.  With this in mind, I see these approaches:
- since the changes to RBAC were minimal according to the documentation, we could just start using the `v1` API instead of `v1beta1` directly (implemented in this PR).
- since the v1beta1 API won't be available anymore in 1.22+, we could remove the `v1beta1` key of the RBAC object altogether, making it fail early if something still uses the deprecated API.

jvrplmlmn: > since the changes to RBAC were minimal according to the documentation, we could just start using the v1 API instead of v1beta1 directly (implemented in this PR).

I would prefer to go with this option first.

mplzik: @jvrplmlmn okay, let me add a TODO note about removing the `v1beta1` line at some point in the future.

mplzik: Don; let me know if this PR looks sufficiently okay to give it a green light :-)

jvrplmlmn (approved):

PR #816, opened 2022-07-13
Caddy-Mixin: Report metrics from *all* handlers, not just reverse_proxy

rgeyer:
The mixin previously filtered the metrics to only match those with a handler=reverse_proxy label.

This would exclude direct file_server access, and presumably others as well.


rgeyer:
Reopening this.

@Dasomeone @gaantunes, could I get eyes on this please?

Dasomeone (approved):
LGTM, this is already the version used in the integration so backporting it to the mixin sounds good to me :)

PR #817, opened 2022-07-19
Include inet6 listen directive

xaque208:


nginx-directory/files/nginx.conf
```diff
@@ -18,6 +18,7 @@ http {
   resolver     kube-dns.kube-system.svc.%(cluster_dns_suffix)s;
   server {
     listen 80;
+    listen [::]:80;
```
xaque208, line 21: I'm not clear what the failure scenario is when ipv6 is not available.

xvzf: there's likely always a link local v6 available

PR #822, opened 2022-08-01
Allow including "All" for single template var

colega:
Sometimes it makes sense to be able to select the "all" value for single-valued template vars. This function actually accepts the "allValue" param, but it doesn't allow vars with "all" option.


krajorama (approved):
looks straightforward

PR #824, opened 2022-08-22
Add timeseriesPanel

bboreham:
["The default and primary way to visualize time series data"](https://grafana.com/docs/grafana/latest/visualizations/time-series/)

Introduced under a new name to avoid breaking all dashboards that use `panel()` and override something.


pracucci (approved):
LGTM!

PR #825, opened 2022-08-22
Mixins: draw graphs at full resolution

bboreham:
Remove "intervalFactor: 2", which corresponded to "Resolution 1/2" in the Grafana UI. This feature has now been dropped from the UI, and I can't see why were ever using it - why would you not want the full resolution for your panels?

image

Left `awx-mixin/dashboards/clustertable.libsonnet` untouched because those are table queries not timeseries.


zalegrala:
This seems reasonable to me.  Replicating this change into my own environment didn't break any of the affected charts running a recent grafana.

zalegrala (approved):

dimitarvdimitrov (approved):
Thanks

PR #828, opened 2022-09-06
Windows Mixin - Correct thresholds for C:// Free

rgeyer:
Thresholds for the available disk space on `C://` were correct, but rendering incorrectly.

It turns out that when thresholds are provided, the "first" entry in the `steps` becomes the base, even if it's value is set. I.E. Grafana overrides it's value with `null`.

This resulted in `dark-green` being the base color, and any utilization between 20-0 % would appear as green.

This PR simply changes the order, so that Grafana does not override the value on import.


mshahzeb (approved):
Thanks! Good catch. LGTM!

PR #829, opened 2022-09-07
Grafana Agent: correct overview dashboard queries

RutgerKe:
The queries use metrics that are in seconds, but multiply by a 1000. The panels still have the unit as seconds. I removed the multiplication instead of changing the units, as it reduces complexity.

You can see in the screenshot from [the docs](https://grafana.com/docs/grafana-cloud/data-configuration/integrations/integration-reference/integration-grafana-agent/#dashboards) that the current 'average scrape interval duration' says hours, while it should be talking about seconds.

overview dashboard

After the change, scrape interval is 15s
image

(The target-sync panel is always 0 for me, I'll have to dig into that separately, so I cannot test that. But I think the same logic holds: The metric is already in seconds, multiplying changes that, and the unit should match)

### DIFF
--- a/grafana-agent-mixin/dashboards/grafana-agent.libsonnet
+++ b/grafana-agent-mixin/dashboards/grafana-agent.libsonnet
@@ -89,7 +89,7 @@ local instance_template = grafana.template.new(
           span=6,
         )
         .addTarget(prometheus.target(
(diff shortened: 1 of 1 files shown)


mshahzeb (approved):
Thank you! Great catch.

Once this is merged I will open a PR to update this in the integration as well!

PR #832, opened 2022-09-15
Evenly spread memcached across nodes

salvacorts:
This PR adds support for using a [`TopologySpreadConstraints`][1] to configure Memcached pods to evenly spread across nodes but allow to schedule multiple pod per node. This is similar to https://github.com/grafana/loki/pull/6415 and https://github.com/grafana/loki/pull/6964

We have two new options:
- `use_topology_spread`: False by default.
- `topology_spread_max_skew`: 1 by default. 

If `use_topology_spread` is true, Memcached pods can run on nodes already running Memcached but will be
spread through the available nodes using a `TopologySpreadConstraints` with a max skew of `topology_spread_max_skew`.
If `use_topology_spread` is false, Memcached pods will not be scheduled on nodes already running Memcached.

[1]: https://kubernetes.io/docs/concepts/scheduling-eviction/topology-spread-constraints/


dannykopping (approved):
LGTM

PR #835, opened 2022-09-29
Remove integration status panels from snmp mixin

mshahzeb:
Remove integration status panels from mixin to make them independent of Grafana Cloud integrations.


rgeyer (approved):
LGTM :)

PR #836, opened 2022-10-03
Vault-mixin: Replacing grafana-pie-chart in favor of the core piechart plugin

gaantunes:
This is for fixing a bug where the user has to install the non-default piechart plugin to avoid a plugin not found error.


rgeyer (approved):
LGTM, thanks for taking care of this.

PR #837, opened 2022-10-04
Use nodejs_version_info for alert NodejsDown.

cristiangreco:
Fix NodejsDown firing even when the application is up.


bfmatei (approved):
lgtm! thanks!

rgeyer (approved):
Thank you!

PR #840, opened 2022-10-14
Add Clickhouse mixin

JonathanWamsley:
This PR adds the Clickhouse mixin featuring 3 dashboards with an optional logs error panel and a prometheus alert.


clickhouse-mixin/Makefile
```diff
@@ -0,0 +1,38 @@
...
+dashboards_out: mixin.libsonnet config.libsonnet $(wildcard dashboards/*)
+	@mkdir -p dashboards_out
+	mixtool generate dashboards mixin.libsonnet -d dashboards_out
+
+prometheus_alerts.yaml: mixin.libsonnet alerts/*.libsonnet
+	mixtool generate alerts mixin.libsonnet -a prometheus_alerts.yaml
+
+
+.PHONY: deploy deploy_rules deploy_dashboards
+deploy: deploy_rules deploy_dashboards
```
BinaryFissionGames, line 34: I think this rule should be removed, the rules it depends on doesn't exist

JonathanWamsley: Removed rule

clickhouse-mixin/dashboards/clickhouse-latency.libsonnet
```diff
@@ -0,0 +1,524 @@
...
+        graphTooltip='shared_crosshair',
+        uid=dashboardUid,
+      )
+      .addLink(grafana.link.dashboards(
+        asDropdown=false,
+        title='Other clickhouse dashboards',
+        includeVars=true,
+        keepTime=true,
+        tags=($._config.dashboardTags),
+      )).addTemplates(
```
rgeyer, line 471: 👌

rgeyer (approved):
Looks good! Great work

PR #841, opened 2022-10-17
MongoDb Mixin: Fixing missing cluster labels in Instance and ReplicaSet dashboards.

gaantunes:
This is fixing some errors caused by the absense of the cluster label on Instance and ReplicaSet dashboards, when there are more than one cluster with Replicas or Instances with the same name across them.

It also fixes a few linting errors on alerts and panel positioning problems.


rgeyer (approved):
LGTM 🚀

PR #844, opened 2022-10-19
Add GitLab mixin

BinaryFissionGames:
This PR adds a mixin for GitLab, containing an overview dashboard and 4 alerting policies.


gitlab/README.md
```diff
@@ -0,0 +1,54 @@
...
+
+For linting and formatting, you would also need and `jsonnetfmt` installed. If you
+have a working Go development environment, it's easiest to run the following:
+
+```bash
+go install github.com/google/go-jsonnet/cmd/jsonnetfmt@latest
+```
+
+The files in `dashboards_out` need to be imported
+into your Grafana server.  The exact details will be depending on your environment.
```
schmikei:
```suggestion
into your Grafana server. The exact details will be depending on your environment.
```

nit: extra space.

JonathanWamsley:
LGTM!

gitlab-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,80 @@
+{
+  prometheusAlerts+:: {
```
rgeyer, line 2: Good detailed alert descriptions. 👌

gitlab-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,80 @@
...
+              summary: 'Large percentage of runner authentication failures.',
+              description:
+                ('{{ printf "%%.2f" $value }}%% of GitLab runner authentication attempts are failing on {{$labels.instance}}, ' +
+                 'which is above the threshold of %(alertsWarningRunnerAuthFailures)s%%.') % $._config,
+            },
+          },
+          {
+            alert: 'GitLabHigh5xxResponses',
+            expr: |||
+              sum by (instance) (rate(http_requests_total{status=~"^5.*"}[5m])) > %(alertsCritical5xxResponses)s
```
rgeyer, line 45: For this alert, and the 4xx responses, is there a reason that an absolute value is being used, rather than a percentage?

This would trigger for errors on 1/10 requests or 1/10000 requests.  Both are "bad" but this gives you no sense of scale.

gitlab-mixin/dashboards/gitlab-overview.libsonnet
```diff
@@ -0,0 +1,725 @@
+local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
+local dashboard = grafana.dashboard;
+local template = grafana.template;
+local prometheus = grafana.prometheus;
+
+local dashboardUid = 'gitlab-overview';
+local matcher = 'job=~"$job", instance=~"$instance"';
+local datasourceRegex = '(?!grafanacloud-usage|grafanacloud-ml-metrics).+';
```
rgeyer, line 8: This is grafana cloud specific, and shouldn't be in the OSS mixin.

gitlab-mixin/dashboards/gitlab-overview.libsonnet
```diff
@@ -0,0 +1,725 @@
...
+    showLabels: false,
+    showTime: true,
+    sortOrder: 'Descending',
+    wrapLogMessage: false,
+  },
+  targets: [
+    {
+      datasource: lokiDatasource,
+      editorMode: 'code',
+      expr: '{filename="/var/log/gitlab/gitlab-rails/exceptions_json.log", %s} | json | line_format "{{.severity}} {{.exception_class}} - {{.exception_message}}"' % matcher,
```
rgeyer, line 369: Might want to make the filename a config parameter, in case this is run in a non-standard way, or within a docker container/k8s pod with different volume mounts.

rgeyer (changes requested):
Few minor things.

gitlab-mixin/dashboards/gitlab-overview.libsonnet
BinaryFissionGames: Good catch, made this a config param

gitlab-mixin/dashboards/gitlab-overview.libsonnet
BinaryFissionGames: removed 👍

gitlab-mixin/alerts/alerts.libsonnet
BinaryFissionGames: Good catch; Definitely makes more sense to have these as percentages. Changed to be percentage based for 4xx and 5xx response alerts.

rgeyer (approved):
LGTM, thanks for making the changes 🚀

PR #845, opened 2022-10-21
Use mixins for container resources settings

pr00se:
The existing `resourcesRequests` and `resourcesLimits` [completely replace](https://github.com/jsonnet-libs/k8s-libsonnet/blob/main/1.22/_gen/core/v1/container.libsonnet#L165-L166) the relevant block of the container config, meaning that any value set to `null` completely removes the field from the final result. This means that it is impossible to set the `cpu` requests but leave the `memory` requests unchanged from whatever it might already be.

To allow setting individual resource request and cpu parameters, add `Mixin` versions of the requests and limits functions.

### DIFF
--- a/ksonnet-util/util.libsonnet
+++ b/ksonnet-util/util.libsonnet
@@ -239,8 +239,28 @@ local util(k) = {
        else {})
     ),
 
+  resourcesRequestsMixin(cpu, memory)::
+    k.core.v1.container.mixin.resources.withRequestsMixin(
+      (if cpu != null
+       then { cpu: cpu }
+       else {}) +
+      (if memory != null
+       then { memory: memory }
+       else {})
+    ),
+
   resourcesLimits(cpu, memory)::
     k.core.v1.container.mixin.resources.withLimits(
+      (if cpu != null
+       then { cpu: cpu }
+       else {}) +
+      (if memory != null
+       then { memory: memory }
+       else {})
+    ),
+
+  resourcesLimitsMixin(cpu, memory)::
+    k.core.v1.container.mixin.resources.withLimitsMixin(
       (if cpu != null
        then { cpu: cpu }
        else {}) +


jhesketh (approved):
Seems reasonable to me.

ksonnet-util/util.libsonnet
```diff
@@ -239,6 +239,16 @@ local util(k) = {
        else {})
     ),
 
+  resourcesRequestsMixin(cpu, memory)::
```
grafanalf, line 242: This is ugly. Is rather have a resourcesRequestsCpuMixin and reaourceRequestsMemoryMixin to not have to deal with the combinatory explosion of dealing with optional parameters.

grafanalf (changes requested):

ksonnet-util/util.libsonnet
nikosmeds: @grafanalf please note that the existing `resourcesRequests()` and `resourcesLimits()` set both cpu and memory. I think it's reasonable that the mixin functions accept the same inputs.

pr00se: > This is ugly
Rude

>I would rather have a resourcesRequestsCpuMixin and reaourceRequestsMemoryMixin to not have to deal with the combinatory explosion of dealing with optional parameters.
Happy to explain my reasoning, thanks for asking!

I thought about just changing the existing functions to use `withRequestsMixin`, but we have some places where we use these functions and rely on them overwriting the whole `requests` block to, for example, explicitly not set a CPU limit.

I considered having separate functions for CPU and memory rather than using optional parameters, as you suggest, but ultimately decided against it because it seemed cleaner to keep the API between the mixin and non-mixin versions equivalent. These single parameter functions would still need to deal with `null` inputs, since we likely don't want to blindly set `request.cpu: null`, so the logic inside each separate function would be basically the same as the logic in the multi-parameter function.

I also considered if we could split the existing functions to just operate on CPU or memory, but that isn't possible, since the underlying `withRequests` [operates on the whole `requests` block](https://github.com/jsonnet-libs/k8s-libsonnet/blob/main/1.22/_gen/core/v1/container.libsonnet#L165-L166) -- setting just the CPU requests would have the side-effect of unsetting the memory request, and vice versa.

Since I didn't want to sign up for mitigating all the downstream effects of changing the existing functions, and it didn't seem desirable to have two completely different interfaces for using the mixin vs. non-mixin versions of these function, I opted to just copy the function signatures and change the behavior. This makes it very easy to switch between the two, since it's just a change to the function name with the same inputs.

pr00se: @grafanalf let me know if the above is sufficient, or if you still need changes. I'd like to get this merged soon to unblock dependent work. Thanks!

grafanalf (approved):

ksonnet-util/util.libsonnet
grafanalf: @pr00se sorry, didn't intent to offend you with the "looks ugly". Please accept my apologies. I don't mean your code is ugly, but that Kubernetes grouping CPU and RAM in the same section does not look nice to me. Since I don't want to block you, I have approved the PR. Sorry about the inconveniences.

PR #846, opened 2022-10-25
Pr/tommyp1ckles/improve operator dashboards

tommyp1ckles:


cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -60,14 +40,115 @@
...
+      },
+      "pluginVersion": "9.2.1-cab7049",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${prometheus_datasource}"
+          },
+          "editorMode": "code",
+          "expr": "sum(rate(cilium_operator_number_of_ceps_per_ces_sum{cluster=~\"$cluster\", pod=~\"$pod\", namespace=~\"$namespace\"}[$__interval]) / rate(cilium_operator_number_of_ceps_per_ces_sum{cluster=~\"$cluster\", pod=~\"$pod\", namespace=~\"$namespace\"}[$__interval])) ",
```
Dasomeone, line 136: We should use `$__rate_interval` instead of `$__interval` or the linter will be upset.

You can try running the following commands to lint your changes:
- Within the mixin: `mixtool lint mixin.libsonnet` (requires [mixtool](https://github.com/monitoring-mixins/mixtool) being installed)
- From the root of the repo: `make lint-fmt cilium-enterprise-mixin`

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -186,60 +270,59 @@
...
         }
       },
-      "pluginVersion": "9.2.0",
+      "pluginVersion": "9.1.5-0100a6a",
       "targets": [
         {
           "datasource": {
             "type": "prometheus",
-            "uid": "${prometheus_datasource}"
+            "uid": "grafanacloud-prom"
```
Dasomeone, line 278: Please ensure the uids are not changed from `${prometheus_datasource}` as it will otherwise not work on generic setups using the templated datasource.
This needs to be done throughout

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -330,752 +558,3704 @@
...
   "style": "dark",
   "tags": [],
   "templating": {
     "list": [
       {
         "current": {
           "selected": false,
-          "text": "default",
-          "value": "default"
+          "text": "grafanacloud-isovalent-prom",
```
Dasomeone, line 4257: Similarly here, please don't override the default values otherwise it'll prevent data from loading when used as a mixin

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -1084,19 +4264,20 @@
...
+        "queryValue": "",
         "refresh": 1,
         "regex": "(?!grafanacloud-usage|grafanacloud-ml-metrics).+",
         "skipUrlSync": false,
         "type": "datasource"
       },
       {
         "current": {
-          "selected": false,
+          "selected": true,
```
Dasomeone, line 4275: Same for this, please leave the defaults as `false`, `All`, `$__all`, etc to ensure it functions properly with new datasources and labels

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -1160,14 +4341,14 @@
           "type": "prometheus",
           "uid": "${prometheus_datasource}"
         },
-        "definition": "label_values(cilium_version{cluster=~\"$cluster\", namespace=~\"$namespace\"}, pod)",
+        "definition": "label_values(cilium_operator_identity_gc_entries{cluster=~\"$cluster\", namespace=~\"$namespace\"}, pod)",
```
Dasomeone, line 4344: If we're updating this template query to use `cilium_operator_identity_gc_entries` instead of `cilium_version` we should update them all to be consistent.

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -1178,11 +4359,31 @@
...
+        ],
+        "query": "10",
+        "skipUrlSync": false,
+        "type": "textbox"
       }
     ]
   },
   "time": {
-    "from": "now-30m",
+    "from": "now-6h",
```
Dasomeone, line 4386: Please keep the default time range to 30 minutes

cilium-enterprise-mixin/dashboards/cilium-operator.json
```diff
@@ -1210,8 +4411,8 @@
     ]
   },
   "timezone": "",
-  "title": "Cilium / Operator",
-  "uid": "integrations_cilium_ent_operator",
+  "title": "Cilium Operator",
+  "uid": "integrations_cilium_operator",
```
Dasomeone, line 4415: Please revert this change as the UID needs to stay consistent

Dasomeone (changes requested):
Could you add a screenshot of the updated dashboard please?

cilium-enterprise-mixin/dashboards/cilium-operator.json
tommyp1ckles: this component doesn't have a _version metric, I selected this just because it's always be emitted so we can get the label values for the dashboard.

cilium-enterprise-mixin/dashboards/cilium-operator.json
tommyp1ckles: ran/fixed lint issues, im getting the error `invalid type for field 'datasource': missing uid field` but that seems to happen in the main branch as well.

cilium-enterprise-mixin/dashboards/cilium-operator.json
tommyp1ckles: changed all in cilium-operator, however I notice theres still many instances of these throughout the other dashboard files?

cilium-enterprise-mixin/dashboards/cilium-operator.json
Dasomeone: If you're getting that, you may have some panels with `"datasource": {}` where you need to add the default settings:
```
"datasource": {
  "type": "prometheus",
},
```

cilium-enterprise-mixin/dashboards/cilium-operator.json
Dasomeone: You're quite right, thought I had caught them earlier but must have used a regex that didn't quite capture all the cases. Will update my PR accordingly

cilium-enterprise-mixin/dashboards/cilium-operator.json
Dasomeone: Ah no, I did't mean in respect to the other dashboards, I meant in respect to the templates in this dashboards. The cluster label for example is retrieved via `cilium_version`

Dasomeone:
Screenshot of updated dashboard as provided by Tom:
image (1)

PR #849, opened 2022-10-27
Clickhouse-mixin: update label value to have a clickhouse metric

JonathanWamsley:
By adding this clickhouse specific metric, when a dashboard is loaded for the first time, a clickhouse instance will be automatically selected.

This improves the user experience when adding a clickhouse integration. When a user views a clickhouse dashboard they will immediately see clickhouse metrics without having to select a clickhouse instance.

Added post changes:
- added in a relative path libsonnet imports as of Dasomeone request. 
- corrected query syntax for error log query when running integrations

### DIFF
--- a/clickhouse-mixin/dashboards/clickhouse-overview.libsonnet
+++ b/clickhouse-mixin/dashboards/clickhouse-overview.libsonnet
@@ -815,7 +815,7 @@ local errorLogsPanel =
           uid: '${loki_datasource}',
         },
         editorMode: 'builder',
-        expr: "{filename='/var/log/clickhouse-server/clickhouse-server.err.log'}{' + matcher + '}",
+        expr: '{filename="/var/log/clickhouse-server/clickhouse-server.err.log", %s}' % matcher,
         legendFormat: '',
         queryType: 'range',
         refId: 'A',


Dasomeone (changes requested):
Apologies, I hadn't seen the previous PRs, but I noticed there are some linting issues, regarding libsonnet imports, where you use hardcoded references to github.com instead of relative path to the grafana-builder libs.

Could you please run the following and fix anything highlighted?

Within the mixin: `mixtool lint mixin.libsonnet` (requires [mixtool](https://github.com/monitoring-mixins/mixtool) being installed)
From the root of the repo: `make lint-fmt clickhouse-mixin`

For the specific issue I noticed, you can do:
```
local g = (import 'grafana-builder/grafana.libsonnet');
local grafana = (import 'grafonnet/grafana.libsonnet');
```

Instead of:
```
local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
```

Dasomeone:
> Apologies, I hadn't seen the previous PRs, but I noticed there are some linting issues, regarding libsonnet imports, where you use hardcoded references to github.com instead of relative path to the grafana-builder libs.
> 
> Could you please run the following and fix anything highlighted?
> 
> Within the mixin: `mixtool lint mixin.libsonnet` (requires [mixtool](https://github.com/monitoring-mixins/mixtool) being installed) From the root of the repo: `make lint-fmt clickhouse-mixin`
> 
> For the specific issue I noticed, you can do:
> 
> ```
> local g = (import 'grafana-builder/grafana.libsonnet');
> local grafana = (import 'grafonnet/grafana.libsonnet');
> ```
> 
> Instead of:
> 
> ```
> local grafana = import 'github.com/grafana/grafonnet-lib/grafonnet/grafana.libsonnet';
> ```

Hmm, looks like it's a jsonnet libs issue, but either way we prefer relative paths for when we vendor it into the integration as we also vendor the libs in question

Dasomeone (approved):

PR #850, opened 2022-10-31
gitlab-mixin: Fix GitLabHighRunnerAuthFailure alert not alerting

BinaryFissionGames:
The two metrics used in the query here end up having different labels, so adding them produces an empty set. Summing the failures/successes based on instance allows the alert to function properly.


gaantunes (approved):
LGTM!

PR #851, opened 2022-10-31
Improve Cilium EE Alerts.

tommyp1ckles:
* Fixes lint errors from "mixtool lint ..."
* Removes not superfluous/actionable alerts.
* Add better descriptions


Dasomeone (changes requested):
Missing the alert import in mixin.libsonnet that needs to be uncommented, but otherwise looks great! :D

Dasomeone (approved):

PR #855, opened 2022-11-03
Aligned alert timeframes with Grafana standard

Dasomeone:
Anything below or equal to 5m : Updated to 5m
15minute alerts are kept with the justification that they may take longer to recover on their own, but aren't important enough to alert on unless it's been firing for 15 minutes


gaantunes (approved):
Thanks for the update!

PR #858, opened 2022-11-07
[clickhouse-mixin]:  Fix image urls, alerts and lint file

JonathanWamsley:
- Replaces the images from this repo for the google cloud buckets
- Sets alert duration to default `"5m"`
- Adds `panel-units-rule` for custom units


Dasomeone (approved):
Nice, thank you for updating these :)

PR #866, opened 2022-11-28
Update watch container image version  to `master-0c44bf6`.

mplzik:


Duologic:
Any relevant changes we should be aware of?

PR #867, opened 2022-11-28
Clickhouse:  Update with pr feedback

JonathanWamsley:
This pr solves the following:
- Removes unrelated jobs to Clickhouse in the template drop down
- Updates network units from bytes to bytes/sec(SI) or `Bps` since it is a `rate` 
- Converts panels, legends and dashboard names to Sentence case
- Update the latency expr that uses microseconds to use `increase` instead of `rate` since because a duration / duration cancel and the chart as a percentage is confusing for elapsed time related metrics
- Added `1 /  sec` unit to `Replica operations` panel


JonathanWamsley:
Hey @v-zhuravlev , would you mind taking a look at this! I believe I have address everything that belongs in the mixin.

mshahzeb (approved):
LGTM! But I would suggest waiting for @v-zhuravlev's review as well.

Thank you!

v-zhuravlev:
LGTM, thanks.

PR #868, opened 2022-12-02
Fix nginx queryType bug

lyj0309:
when try to edit the dashboard and not specified queryType, the default value is range, but the true value is instant
image


gaantunes (approved):
Thanks for the changes!

PR #870, opened 2022-12-05
Fix MongoDB replicaset states panel

v-zhuravlev:
1. Mongodb replicaset states panel could have extra rows if external_labels are used on the grafana_agent side. Grouping by only relevant labels fixes this potential problem.
2. Same applies to 'Current ReplSet State' in Mongodb instances dashboard. It is broken if external_labels present. Simplified calculations (removed transformations, changed timeseries to table output) to make it work with extra labels.
3. Also fixes 'SECUNDARY' typo in value mapping
image

### DIFF
--- a/mongodb-mixin/dashboards/MongoDB_Instance.json
+++ b/mongodb-mixin/dashboards/MongoDB_Instance.json
@@ -285,12 +285,23 @@
       "type": "stat"
     },
     {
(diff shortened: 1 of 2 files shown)


gaantunes (approved):
LGTM, thanks for taking the time to check it out!

PR #871, opened 2022-12-05
Dasomeone/cilium enterprise add missing tags makefile

Dasomeone:
As @stevo-f3 mentioned on the initial PR for adding the Cilium Enterprise mixin, two key aspects were not included:
1) The makefile was missing
2) Dashboard tags were used for navigation, but not included in the mixin

This PR addresses these two concerns.


gaantunes (approved):
Thanks for the changes, LGTM!

PR #873, opened 2022-12-06
Mongodb-mixin: Fix replication lag panel

v-zhuravlev:
Fix mongodb lag panel, as each mongodb secondary node returns metrics  `mongodb_mongod_replset_member_replication_lag` for ALL other secondary nodes, not only for itself. So results should be grouped by internal label `name` which properly identifies secondary node insteadof service_name (which identifies node where metrics were taken, not actual replication set member in case of this metric).

telegram-cloud-photo-size-2-5222237850760757785-y


gaantunes (approved):
Nice catch again, thanks for that!

PR #877, opened 2022-12-08
Tweaks TensorFlow dashboard and adds grafonnet dep

StefanKurek:
Changes directly related to feedback from https://github.com/observIQ/cloud-onboarding/pull/47


gaantunes (approved):
Thanks for the changes!

Please exclude the job rule for this panel on the .lint file, since it is not being used anymore.

StefanKurek:
> Thanks for the changes!
> 
> Please exclude the job rule for this panel on the .lint file, since it is not being used anymore.

@gaantunes good catch. Fixed and pushed.

PR #878, opened 2022-12-08
Rabbit MQ: Fixing linting problems

gaantunes:
This is fixing some linting problems:
- Adding job variable to all dashboards
- Renaming the alerts
- Removing templates from alert summaries


Dasomeone (approved):
LGTM, with the caveat that mixtool will hit on the updated Alert names, but as that's on the backlog for being fixed it's all good to me :)

PR #885, opened 2022-12-13
Kafka mixin updates

v-zhuravlev:
- Kafka overview: Show only 0.99 percentile by default. Showing all percentile looks messy.
- Kafka lag: Change table panel to bar chart for partitions per topic panel
- Kafka lag: Stretch kafka lag dashboard to full screen width
- Kafka lag panels: Convert old graph to timeseries (message per sec/per minute)
- Kafka lag: Change delta() to increase() for per minute metrics. Delta could return negative increase if counter resets, while increase() ignores resets().

delta(): "delta should only be used with gauges and native histograms where the components behave like gauges (so-called gauge histograms)."

increase(): "Breaks in monotonicity (such as counter resets due to target restarts) are automatically adjusted for"

https://prometheus.io/docs/prometheus/latest/querying/functions/#delta

### DIFF
--- a/kafka-mixin/dashboards/kafka-lag-overview.json
+++ b/kafka-mixin/dashboards/kafka-lag-overview.json
@@ -73,7 +73,7 @@
       "targets": [
         {
           "exemplar": true,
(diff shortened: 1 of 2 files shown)


gaantunes (approved):
Thanks for the changes, LGTM!

PR #888, opened 2022-12-14
Update vault mixin

v-zhuravlev:
- Change panels to timeseries
- Change old bar charts to new bar gauges
- Add job, instance selectors
- Fix secrets table, add gauge to the table
- Add README

vaultscreen

### DIFF
--- a/vault-mixin/dashboards/vault.json
+++ b/vault-mixin/dashboards/vault.json
@@ -2753,7 +2753,7 @@
         "definition": "label_values(vault_secret_kv_count{job=~\"$job\",instance=~\"$instance\"}, mount_point)",
         "hide": 0,
         "includeAll": true,
-        "label": "",
+        "label": "mountpoint",
         "multi": true,
         "name": "mountpoint",
         "options": [],
--- a/vault-mixin/dashboards/vault.json
+++ b/vault-mixin/dashboards/vault.json
@@ -1,47 +1,11 @@
 {
-  "__requires": [
-    {
-      "type": "grafana",
-      "id": "grafana",
-      "name": "Grafana",
-      "version": "7.0.3"
-    },
-    {
-      "type": "panel",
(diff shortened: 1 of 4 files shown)


vault-mixin/dashboards/vault.json
```diff
@@ -2204,340 +2361,419 @@
...
+        },
+        "datasource": {
+          "type": "prometheus",
+          "uid": "${datasource}"
+        },
+        "definition": "label_values(vault_secret_kv_count{job=~\"$job\",instance=~\"$instance\"}, mount_point)",
         "hide": 0,
         "includeAll": true,
-        "label": "Mount Point:",
+        "label": "",
```
Dasomeone, line 2756: If we have labels for `job` and `instance` it makes sense to also have it for the custom variables

Dasomeone (commented):
Generally looks quite good, I like the additional panels!

I wonder if we could add a bit of context to the readme so there's a clear distinction between the "vault-mixin" and "vault" folders, though. Thoughts?

v-zhuravlev:
> Generally looks quite good, I like the additional panels!
> 
> I wonder if we could add a bit of context to the readme so there's a clear distinction between the "vault-mixin" and "vault" folders, though. Thoughts?

Well, I think README here self describes what this mixin does?

vault-mixin/dashboards/vault.json
v-zhuravlev: updated.

Dasomeone:
> > Generally looks quite good, I like the additional panels!
> > I wonder if we could add a bit of context to the readme so there's a clear distinction between the "vault-mixin" and "vault" folders, though. Thoughts?
> 
> Well, I think README here self describes what this mixin does?

Good point, I guess it's more a case of the other `vault` folder not being properly documented :0

PR #892, opened 2022-12-16
RabbitMQ: small style changes and update alerts

v-zhuravlev:
- Change row names
- Prefix alertgroup with RabbitMQ
- Update alerts (5m grace period for most, removed unused title annotation)


gaantunes (approved):
Thanks for the changes, LGTM!

PR #900, opened 2023-01-10
WSO2 Enterprise Integrator: Adding Job variables to all dashboards

gaantunes:
This is adding a Job template variable to all dashboards, and solving all linting problems when ran againt the current version of mixtool.


Dasomeone (approved):
LGTM :)

PR #909, opened 2023-01-31
Fix lint issues with Envoy Mixin

Dasomeone:
This PR introduces fixes to a number of lint warnings raised by mixtool:

- The `job` template now correctly supports multiple variables
- The `instance` template has been updated with correct label
- All use of `$__interval` has been replaced with `$__rate_interval`
- A lint config has been introduced to suppress noisy and irrelevant warnings


mshahzeb (approved):
LGTM!

PR #910, opened 2023-01-31
Fix lint issues with ElasticSearch mixin

Dasomeone:
This PR introduces a lint config file and addresses some lint warnings raised by Mixtool:
- Ensure all references to template variables are inclusive regexes rather than exact matches
- Fix allValue for templated vars
- Fix labels for templated vars
- Allow templated vars to be multiselect


gaantunes (approved):
LGTM, thanks!

PR #913, opened 2023-01-31
Remove compact URL usage

gelicia:
The Explore team is seeking to deprecate the compact URL format. That format allowed parameters to be passed into the explore URL as an array in a specific order. Due to the increasing number of parameters, this was not going to be a workable, future forward format. This removes the one use of compact URLs from this repository. Search parameters were `orgId=1&left=%5B` and `orgId=1&left=[`


gaantunes:
I will be removing this link since we actually do not ship logs on this integration.

Issue #916, opened 2023-02-01
Traefik dashboards use $__interval and don't show anything by default in Grafana Cloud


regner:
When installing the Traefik dashboards they default to "Last 1 hour" for the time. This combined with the integration defaults having a 60 second scrape default means nothing actually gets displayed. The dashboards look empty. If you expand the time to "Last 3 hours" most of the panels suddenly get data.

I believe the problem is down to the usage of `$__interval` instead of `$__rate_interval` combined with the scape time being 60 seconds.

It probably makes sense to switch all of the `$__interval` values to `$__rate_interval` if I understand things. Which I might now.

v-zhuravlev:
Using $__inteval is done on purpose to make sure that  increase() increments don't **overlap** and Totals calculation is correct. If we change it to $__rate_interval than totals calculation can be too high due to dynamic nature of $__rate_interval and overlapping of increase() increments.

![image](https://user-images.githubusercontent.com/14870891/216453322-756ef0a2-a208-4320-bdc0-10fca19af3eb.png)

Let me see if this can be fixed some other way.

regner:
OK, I definetly did not/still do not understand that. I will be very interested in how to properly solve it.

Issue #915, opened 2023-02-01
Traefik dashboard has hard coded services


regner:
The Traefik dashboard has a panel for "Requests by service" that instead of using `$service` is hard coded with `.*grafana@.*|prometheus.*|gitea.*` which is probably not relevant to most.

https://github.com/grafana/jsonnet-libs/blob/master/traefik-mixin/dashboards/traefikdash.json#L1452

v-zhuravlev:
Thanks for spotting this.

PR #917, opened 2023-02-01
traefik-mixin: dashboard fixes for interval and hardcoded service names

regner:
This fixes both issues #915 and #916.


v-zhuravlev:
Thanks, I left some comments in issues you raised.

PR #918, opened 2023-02-02
nginx-directory: Add optional help text

iainlane:
Sometimes you might want to add additonal information beyond the link text to tell people what a service is. Let services specify a `help` element and render that as a superscript "?" tooltip, if present.


jvrplmlmn (approved):
Nice 👍

PR #919, opened 2023-02-02
oracledb-mixin: fix lint errors for missing units

schmikei:
```sh
jsonnet-libs/oracledb-mixin on fix-oracledb-lint using ☁️  
➜ mixtool lint mixin.libsonnet

jsonnet-libs/oracledb-mixin on fix-oracledb-lint using ☁️  
➜ echo "$?"
0
```


Dasomeone (approved):
LGTM, thanks for fixing these

PR #920, opened 2023-02-02
Traefik mixin: improve errors and load stat panels

v-zhuravlev:
This addresses traefik integration issue with  errors panel. Also updates total calls panel as well.

image


regner:
Thank you very much for sorting out these fixes. :)

Dasomeone (approved):
LGTM, thanks for fixing :)

PR #921, opened 2023-02-03
MSSQL mixin: Set min & max on "Buffer cache hit percentage" panel.

BinaryFissionGames:
* Adds a range of 0-100 on the "Buffer cache hit percentage" panel to make the display match the bounds of the value.

This was already present in the screenshot of the dashboard, but the change seemed to get dropped at some point on my end.


gaantunes (approved):
Thanks for the changes.

PR #922, opened 2023-02-03
Update springboot mixin dashboard

v-zhuravlev:
- Remove application label (can be missing in java app)
- Change to timeseries
- Apply Sentence case
- Remove support for job all value .+ (otherwise 'uptime query' and some metrics queries would return too many unrelated timeseries )

### DIFF
--- a/spring-boot-mixin/dashboards/spring-boot-statistics_rev2.json
+++ b/spring-boot-mixin/dashboards/spring-boot-statistics_rev2.json
@@ -26,7 +26,7 @@
   "fiscalYearStartMonth": 0,
   "gnetId": 6756,
   "graphTooltip": 0,
-  "id": 82,
+  "id": 9,
   "links": [],
   "liveNow": false,
   "panels": [
@@ -118,7 +118,7 @@
         },
         "textMode": "auto"
       },
-      "pluginVersion": "9.3.2",
+      "pluginVersion": "9.3.6",
       "targets": [
         {
           "datasource": {
@@ -206,7 +206,7 @@
         },
         "textMode": "auto"
       },
-      "pluginVersion": "9.3.2",
+      "pluginVersion": "9.3.6",
       "targets": [
         {
           "datasource": {
@@ -303,7 +303,7 @@
         },
         "textMode": "auto"
       },
-      "pluginVersion": "9.3.2",
+      "pluginVersion": "9.3.6",
       "targets": [
         {
           "datasource": {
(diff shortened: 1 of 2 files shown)


spring-boot-mixin/dashboards/spring-boot-statistics_rev2.json
```diff
@@ -3422,8 +4104,8 @@
     ]
   },
   "timezone": "",
-  "title": "Spring Boot Statistics",
-  "version": 15,
+  "title": "SpringBoot statistics",
```
Dasomeone, line 4107: This should be "Spring Boot statistics", as that's how it's officially capitalized and written with a space in between

Dasomeone (changes requested):
Please re-check the layout, because when I loaded it some panels had gotten moved around and there are empty spaces where there shouldn't be: 
image

Additionally, it would be nice to have before/after pics in the PR description for an easier comparison of visual changes :)

Dasomeone:
Forgot to mark this during the review:
> Remove support for job all value .+ (otherwise 'uptime query' and some metrics queries would return too many unrelated timeseries )
For this, it might be worth considering just changing the affected queries to a hardcoded job value, rather than affecting allValue for all uses of `job`?

But in any case, if kept like this, please add a lint config with an exclusion for the allValue rule check for the `job` template

v-zhuravlev:
image

spring-boot-mixin/dashboards/spring-boot-statistics_rev2.json
v-zhuravlev: > Please re-check the layout, because when I loaded it some panels had gotten moved around and there are empty spaces where there shouldn't be: image
> 
> Additionally, it would be nice to have before/after pics in the PR description for an easier comparison of visual changes :)

Empty space is only visible if there is no data due to repeated panels. Since there is always fixed number of panels dynamically generated - row is populated properly without any gaps. I attached the screenshot from sample app to visualize.

v-zhuravlev:
> For this, it might be worth considering just changing the affected queries to a hardcoded job value, rather than affecting allValue for all uses of job?
I think it can lead to unpredictable (or at least complex) behavior, as part of panels would follow one logic (returning all possible series) and some returning only related to specific jobs. Since having special char for ALL is only performance related best practice, I think here it can be skipped as there are no good results with `.+`.

Dasomeone (approved):
Looks good, thanks for answering the comments :)

PR #925, opened 2023-02-09
Add `redis` and `redis-mixin`

colega:
This is some jsonnet used internally at Grafana Labs to operate redis, it also includes the mixin with a dashboard, alerts and some runbooks.

### DIFF
--- a/redis/config.libsonnet
+++ b/redis/config.libsonnet
@@ -1,0 +1,31 @@
+{
+  _config+: {
+    namespace: error 'must specify namespace',
+    redis+: {
+      password: error 'must specify password',
+      port: 6379,
+      sentinel_port: 26379,
+      replicas: 3,
+      down_after_milliseconds: 15000,
+      diskSize: '50Gi',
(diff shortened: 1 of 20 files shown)


redis/config.libsonnet
```diff
@@ -0,0 +1,31 @@
+{
+  _config+: {
+    namespace: error 'must specify namespace',
+    redis+: {
+      password: error 'must specify password',
+      port: 6379,
+      sentinel_port: 26379,
+      replicas: 3,
+      down_after_milliseconds: 15000,
+      diskSize: '50Gi',
```
jesusvazquez, line 10: Is this a sane default? Seems huge to me for a start.

jesusvazquez (approved):
LGTM

redis/config.libsonnet
colega: It's opinionated, yes... but anyone can change it.

colega:
Thank you for the review.

PR #926, opened 2023-02-13
memcached: Relabel Reads/Writes to Received/Transmitted

56quarters:
When talking about a cache server, creating panels based on the number of bytes read from and written to the network is deceptive since:

* When an application reads from the cache, the cache writes bytes to the network.
* When an application writes to the cache, the cache reads bytes from the network.

Changing the panels to "received" and "transmitted" avoids this ambiguity.

Signed-off-by: Nick Pillitteri <nick.pillitteri@grafana.com>


mattmendick (approved):
Had to run this over in my head several times before I understood it, but I agree, this makes much more sense. Thanks for spotting this.

PR #929, opened 2023-02-13
Redis: make resources configurable

colega:
Instead of having to manually override the container requests, this
makes the resources manually configurable.

Also lowers the default requests/limits to some more conservative
numbers.


colega:
After thinking twice, I don't think this code worth the effort, since overriding the limits can now be done by just writing:

```jsonnet
    redis_server_container+:: k.core.v1.container.mixin.resources.withRequestsMixin({
      memory: '512Mi',
      cpu: '100m',
    }),
```

PR #930, opened 2023-02-14
Fix conntrack table full alert.

tommyp1ckles:
Signed-off-by: Tom Hadlaw <tom.hadlaw@isovalent.com>


Dasomeone (approved):
Looks great, thanks for fixing it!

PR #934, opened 2023-02-22
Add Apache Cassandra Mixin

schmikei:
Adding a mixin for [Apache Cassandra](https://cassandra.apache.org/_/index.html) using the [JMX exporter for Prometheus](https://github.com/prometheus/jmx_exporter/)

This includes 3 dashboards `Apache Cassandra overview, Apache Cassandra nodes, Apache Cassandra keyspaces`

It also adds alerts

```HighReadLatency: There is a high level of read latency within the node.
HighWriteLatency: There is a high level of write latency within the node.
HighPendingCompactionTasks: Compaction task queue is filling up.
BlockedCompactionTasksFound: Compaction task queue is full.
HintsStoredOnNode: Hints have been recently written to this node.
UnavailableWriteRequestsFound: Unavailable exceptions have been encountered while performing writes in this cluster.
HighCpuUsage: A node has a CPU usage higher than the configured threshold.
HighMemoryUsage: A node has a higher memory utilization than the configured threshold.
```

node_1
node_2
keyspace
overview_2
overview_1


apache-cassandra/dashboards/cassandra-nodes.libsonnet
```diff
@@ -0,0 +1,1590 @@
...
+            {
+        matcher: {
+          id: 'byFrameRefID',
+          options: 'A',
+        },
+        properties: [
+          {
+            id: 'color',
+            value: {
+              fixedColor: 'green',
```
kuiperda, line 1102: Should this be red?

apache-cassandra/dashboards/cassandra-nodes.libsonnet
```diff
@@ -0,0 +1,1590 @@
...
```
kuiperda, line 1238: Should this be red?

apache-cassandra/dashboards/cassandra-nodes.libsonnet
```diff
@@ -0,0 +1,1590 @@
...
```
kuiperda, line 1372: Should this be red?

apache-cassandra/dashboards/cassandra-overview.libsonnet
```diff
@@ -0,0 +1,1782 @@
...
+  pluginVersion: '9.3.6',
+};
+
+local connectionTimeoutsPanel(matcher) = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'cassandra_connection_timeouts_count{' + matcher + '}',
+      datasource=promDatasource,
+      legendFormat='{{ connection }} - {{ instance }} - {{ cluster }}',
```
Dasomeone, line 177: Not quite sure I agree with this legend format 🤔 

It may just be me, but I feel this could get really overwhelming having both the IP *and* the instance name, as both are quite long strings. 
One option could be to display the legend as a right-aligned table, which would make it a little easier?

@gloriasc What are your thoughts on this? (See below image from Keith for reference "Connection timeouts")
I think I prefer the style below with only `connection` and `cluster`
image

apache-cassandra/config.libsonnet
```diff
@@ -0,0 +1,22 @@
...
+    alertsWarningPendingCompactionTasks15m: 30,
+    alertsCriticalBlockedCompactionTasks5m: 1,
+    alertsWarningHintsStored1m: 1,
+    alertsCriticalUnavailableWriteRequests5m: 1,
+    alertsCriticalHighCpuUsage5m: 80,  //percent: emitted metric has range 0-100
+    alertsCriticalHighMemoryUsage5m: 80,  //percent: calculated as ratio then multiplied by query
+
+    enableLokiLogs: true,
+    enableDatacenterLabel: false,
+    enableRackLabel: false,
```
Dasomeone, line 20: Similar to the readme on how to toggle loki logs, would be great to give datacenter and rack labels a mention too :)

apache-cassandra/dashboards/cassandra-overview.libsonnet
```diff
@@ -0,0 +1,1782 @@
...
+      sort: 'none',
+    },
+  },
+};
+
+local writesPanel(matcher) = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'increase((sum by (cluster) (cassandra_keyspace_writelatency_seconds_count{' + matcher + '})[$__rate_interval:]))',
```
Dasomeone, line 524: I might be remembering wrong, but I think there's a gotcha here about using both `increase()` and `$__rate_interval`. 

I _believe_ you need to use $__interval to have properly spaced increments without messing with backdated entries when inside `increase()`

apache-cassandra/dashboards/cassandra-overview.libsonnet
```diff
@@ -0,0 +1,1782 @@
...
+    ),
+    prometheus.target(
+      'sum(cassandra_keyspace_writelatency_seconds{' + matcher + ', quantile="0.95"} >= 0) by (cluster)',
+      datasource=promDatasource,
+      legendFormat='{{ cluster }} - p99',
+    ),
+  ],
+  type: 'histogram',
+  title: 'Write latency',
+  description: 'Average local write latency for this cluster',
```
Dasomeone, line 995: I think there's been a misunderstanding here, and it's entirely on me 😅 
When we reviewed the spec, I recommended using a histogram, but didn't specify which visualisation style which I definitely should have!

What I was thinking of including is actually two panels, namely a latency histogram *heatmap*, and a Quartile Latency *time series* panel, using the histogram *data*. Again entirely on me, but to get an idea of what I'm thinking, here's a few pictures:
image
image

Taken from this article: https://towardsdatascience.com/prometheus-histograms-with-grafana-heatmaps-d556c28612c7

@gloriasc would love to hear your thoughts on this as well :)

Dasomeone (commented):
Left a couple comments, have to run now for an errand. 

Would it be possible to get a dev environment spun up so we can check out the dashboards with data, similar to what you set up with Snowflake? :D

apache-cassandra/Makefile
```diff
@@ -0,0 +1,34 @@
...
+.PHONY: build
+build: vendor
+
+.PHONY: fmt
+fmt:
+	find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
+		xargs -n 1 -- $(JSONNET_FMT) -i
+
+.PHONY: lint
+lint: build
```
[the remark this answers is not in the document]
schmikei: 🤔 Happy to help any way I can, I just ran reinstalled from mixtool from main and things appear to be working?

Tested by removing a description from a panel. 

image

Definitely strange, but perhaps try reinstalling mixtool?

apache-cassandra/dashboards/cassandra-overview.libsonnet
schmikei: I changed it to be just connection and cluster 👍

apache-cassandra/config.libsonnet
schmikei: Added!

apache-cassandra/dashboards/cassandra-overview.libsonnet
schmikei: Yep, your intuition was correct using the `$__interval` with a specified window of `1m` appeared to resolve the issue!

apache-cassandra/dashboards/cassandra-overview.libsonnet
gloriasc: Agreed! One need to be able to identify the instance but between those options, the shorter the better :)

apache-cassandra/dashboards/cassandra-overview.libsonnet
gloriasc: Yes! Absolutely! 100% agreed! Those graphs would be awesome

gloriasc (approved):
Great job! Thanks for updating requests after review meeting!

Dasomeone (approved):
One or two small things to check, but otherwise looks great! 
Thanks again for adding all the extras from the review call :D

Dasomeone (approved):
LGTM, great work!

PR #939, opened 2023-03-09
Add expectation that apache-cassandra mixin logs panel will work with the rest of the matchers

schmikei:
Expecting people using the integration to have a logs_scrape configuration of this, this was missed when i first implemented the mixin, sorry about that! 

```yaml
 scrape_configs:
    - job_name: integrations/apache-cassandra
      static_configs:
        - targets: [localhost] 
          labels:
            job: integrations/apache-cassandra
            instance: '<your-instance-name>' # <node1>:9000
            cluster: '<your-cluster-name>'
            __path__: /var/log/cassandra/system.log

```

### DIFF
--- a/apache-cassandra/README.md
+++ b/apache-cassandra/README.md
@@ -43,12 +43,25 @@ Cassandra system logs are enabled by default in the `config.libsonnet` and can b
 }
 ```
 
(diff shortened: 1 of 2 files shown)


Dasomeone (approved):
LGTM, nice catch :D

PR #943, opened 2023-03-10
Rename apache-cassandra folder to apache-cassandra-mixin

schmikei:
This was causing drone CI failures because it does not properly match the make lint expectation that all mixin directories suffix with `-mixin`


schmikei:
Closed appropriately in #942

PR #948, opened 2023-03-14
Migrating from old container registry (k8s.gcr.io) to new GA Container registry (registry.k8s.io)

logyball:
As part of the effort to migrate off of the old container registry, we are replacing all of the `k8s.gcr.io` references with `registry.k8s.io`.


Pokom (approved):
:+1: LGTM

PR #951, opened 2023-03-20
fix(ksonnet-util): provide fallback for manifestYaml

Duologic:
`std.manifestYamlDoc` implements the same feature but has a slightly different output,
this change maintains the old Tanka functionality but provides a compatible fallback that
works with plain jsonnet.


JoaoBraveCoding (approved):
/lgtm this is indeed needed if you are not using tanka

PR #953, opened 2023-03-23
Wildfly minor fix

armstrmi:
Some minor fixes caught on some further internal review:
- Dashboard labels updated to sentence case
- `job` -> `Job` selector in `Wildfly datasource` dashboard.
- `Wildfly Datasource` -> `Wildfly datasource` selector in `Wildfly datasource` dashboard.
- Updated some dashboard queries to use the appropriate interval for increase queries


gaantunes:
What is the rational behind the use of `__interval` instead of `__rate_interval`?

schmikei:
> On a previous mixin work with @schmikei it was found that using `__rate_interval` on an `increase` function created a kind of "gotcha" so to speak. Made some issues with how the query was performing. [#934 (comment)](https://github.com/grafana/jsonnet-libs/pull/934#discussion_r1115581388)

I should clarify the irregularity was of a sum of an increase I don't think that applies in this case @armstrmi

gaantunes:
Huum got it... I _think_ this could be solved by the use of min step on the query configuration in the panel, though. Can you try it out?

armstrmi:
For sure, I can try it out

armstrmi:
Alright so upon further inspection theres a few things:
1. Would you be able to clarify what changes to the min step you had in mind?

2. I tried some different things in regards to the min step. 
image
Here is an example of the metric collection with no increase or min step included. In this image it shows a consistent stream of data coming in from the starting point of ~9.

3. Upon placing the `increase` function on the metric with no min step but with a `rate_interval`.
image
It is at this point that we start to see some inconsistencies with the data coming in. Now even updating with a min step of `1m` still remains about the same result.
image

4. we see something similar with `__interval`.
image
And in both cases with `interval` and `rate_interval`, there doesn't seem to be a consistent stream of data coming in until the min step is set to 3m which does not seem like something we would want to do.
image

I'm wondering if you have any suggestions as to which method of approach I should take going forward, and/or if some of these inconsistencies are permissible/ are to be expected. In either sense I think it could help me better understand which interval / min step to use going forward. If you could help shed some light on why the dashboards are having trouble calculating the values for these queries even tho there appears to be a steady stream of data coming through, it would be appreciated.

gaantunes (approved):
Alright, seems like using `increase` with `__interval` is the way to go here. If this worked as expected in your tests, we can go foward and merge.

PR #955, opened 2023-03-28
Tweaks dashboards for CouchDB mixins to use $__interval

StefanKurek:
Tweaks the CouchDB dashboards to be a little more accurate. They are already pushed up to the shared Grafana Cloud instance.

Also switches the "Cluster" selector to "CouchDB cluster". Forgot that "Cluster" was reserved.

### DIFF
--- a/apache-couchdb-mixin/alerts/alerts.libsonnet
+++ b/apache-couchdb-mixin/alerts/alerts.libsonnet
@@ -7,7 +7,7 @@
           {
             alert: 'CouchDBUnhealthyCluster',
             expr: |||
-              min by(job, cluster) (couchdb_couch_replicator_cluster_is_stable) < %(alertsCriticalClusterIsUnstable5m)s
+              min by(job, couchdb_cluster) (couchdb_couch_replicator_cluster_is_stable) < %(alertsCriticalClusterIsUnstable5m)s
             ||| % $._config,
             'for': '5m',
             labels: {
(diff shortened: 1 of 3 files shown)


apache-couchdb-mixin/alerts/alerts.libsonnet
```diff
@@ -7,7 +7,7 @@
           {
             alert: 'CouchDBUnhealthyCluster',
             expr: |||
-              min by(job, cluster) (couchdb_couch_replicator_cluster_is_stable) < %(alertsCriticalClusterIsUnstable5m)s
+              min by(job, couchdb_cluster) (couchdb_couch_replicator_cluster_is_stable) < %(alertsCriticalClusterIsUnstable5m)s
```
Dasomeone, line 10: Won't comment on it throughout, but what do you mean `cluster` is reserved? is this something to do specifically with couchdb also reporting cluster metrics aside from the cluster label we discussed in the spec?

If it's the spec, I don't see how that's reserved as we use it in many other integrations without issue :)

StefanKurek: @Dasomeone Totally fair question as I am deviation from the spec actually. If I recall correctly, we were all talking about what do about cluster vs job selectors. I believe Ryan said that we should definitely keep a cluster selector (instead of relying on job), but we should rename it to something different as "cluster" is a label used by kubernetes in grafana or something like that. I then promptly forgot to take this advice until the Redis Enterprise spec was being created.

Dasomeone (approved):

PR #960, opened 2023-04-05
day of week calculation

aldernero:
I was going through this repo yesterday and realized that it would be useful to have some day of week calculations (specifying actions by day of week, filtering output by day of week, etc). This PR does the following:

- adds a `day_of_week(year, month, day)` function to `date.libsonnet`, which returns a day of the week as an integer (0=Sunday, 1=Monday, etc).
- adds unit tests to verify correctness
- since there are now unit tests for two different functions, I added a `[function_name]` label to each unit test.


aldernero:
After some discussion I'll open up a similar PR in a more general repo

colega:
@aldernero can you link your new PR here please, for reference?

PR #972, opened 2023-04-20
Redis Enterprise: Fix filename filter to accept multiple databases

schmikei:
On a node there is multiple redis logs that have numerics differentiating the different databases i.e `/var/opt/redislabs/log/redis-1.log`

image


v-zhuravlev (approved):

rgeyer (approved):
🚢

PR #977, opened 2023-05-02
Allow excluding mount points using jsonnet.

mplzik:
This will allow overriding the mount point exclusions. A similar
mechanism already exists for filesystems.


Duologic (approved):
Perhaps also add a function that modifies this so that

```
+ { excluded_mount_points:: '...' }
```

can become this:

```
+ nodeExporter.withExcludeMountPoints('...')
```

jvrplmlmn (approved):
👍

PR #981, opened 2023-05-03
OpenSearch Mixin: update alert labels

JonathanWamsley:
Labels are update to the actual metric labels. Alert descriptions will be rendered correctly then.

### DIFF
--- a/opensearch-mixin/README.md
+++ b/opensearch-mixin/README.md
@@ -70,7 +70,7 @@ The OpenSearch search and index overview dashboard provides details on request p
 ## Alerts Overview
 
 
-| Alert                               | Description                                                                     |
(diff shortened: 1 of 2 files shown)


Dasomeone (approved):
Thanks for catching this Jon!
Approving, but I think the table in the readme also needs to be updated, right?

Just give me a ping when you want me to merge it :)

JonathanWamsley:
@Dasomeone I thought a table looked better and used a table template from another mixin that used `Description` as a column header. These column values are actually `Summary` values from the alerts.libsonnet. The summary did not change but I updated the header to match the alerts.libsonnet terminology. Hope that clears things up!

Should be ready to merge!

PR #993, opened 2023-06-05
Add alerts to agent mixin used by Grafana cloud integration

mshahzeb:
The alerts are copied over from the [mixin](https://github.com/grafana/agent/tree/main/production/grafana-agent-mixin) on the agent repo.

### DIFF
--- a/grafana-agent-mixin/alerts/alerts.libsonnet
+++ b/grafana-agent-mixin/alerts/alerts.libsonnet
@@ -1,410 +1,74 @@
-local config = import '../config.libsonnet';
-local _config = config._config;
-
 {
   prometheusAlerts+:: {
     groups+: [
       {
-        name: 'grafana-agent-tracing',
-        rules: [
-          {
-            alert: 'AgentTracingReceiverErrors',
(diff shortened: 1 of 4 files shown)


grafana-agent-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,78 @@
...
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'GrafanaAgentHealthChecks',
+        rules: [
+          {
+            alert: 'GrafanaAgentDown',
+            expr: |||
+              up{
+                job="integrations/agent",
```
Dasomeone, line 11: Minor nitpick, but maybe makes sense to move this to the config and use a variable instead, in case we want to split the grafana-agent and agent-check usage down the road

Dasomeone (approved):
LGTM

grafana-agent-mixin/alerts/alerts.libsonnet
mshahzeb: @Dasomeone The agent check integration already has a different job name `integrations/agent-check`

PR #995, opened 2023-06-08
Update apache mesos cluster filter

JonathanWamsley:
This PR
- Adds a cluster filter so the template selector can filter the dashboard by mesos_cluster type
- Fixes typos

Before selecting a cluster would do nothing because the mesos_cluster was missing from the PromQL queries as shown below:
Screenshot 2023-06-08 at 10 45 53 AM

With this fix the cluster filters applies to the dashboard.
Screenshot 2023-06-08 at 10 38 39 AM


v-zhuravlev (approved):
LGTM

PR #997, opened 2023-06-19
Sort dashboard variable values in ascending (A-Z) order by default

charleskorn:
Currently, drop-down selectors for variables show their values in descending (Z-A) order. A-Z seems like a more reasonable default.


malcolmholmes:
I would suggest this is unwise. This lib has been in existence for 5+ years. We don't know what will break if we change this.

My counter-suggestion is to take a look at the new [Grafonnet](https://github.com/grafana/grafonnet). Perhaps we can make a higher level abstraction on top of it similar to grafana-builder.

Without building on newer, better maintained code, I'd suggest a change such as this is unwise.

charleskorn:
Yep, that was going through my mind as I put this together. I rationalised it as a very small, specific change, but I'm happy to close this if you're not willing to accept it.

PR #998, opened 2023-06-20
Introduce packaging to Docker Mixin

malcolmholmes:
A proposal for a packaging format for observability packages that is back-compatible with
Jsonnet Mixins.


malcolmholmes:
to test, render this:

* First, create a file, `m.jsonnet` that makes the hidden elements in a mixin visible:
```
local mixin = import 'mixin.libsonnet';
{
  grafanaDashboards: {
    [key]: mixin.grafanaDashboards[key]
    for key in std.objectFields(mixin.grafanaDashboards)
  }
}
```

* Next, execute: `jsonnet -J vendor m.jsonnet`
* This will give you the json output from the mixin.
* Alternatively, execute: `jsonnet -J vendor polly.libsonnet` to get the package contents.

docker-mixin/dashboards/docker/variables.libsonnet
```diff
@@ -0,0 +1,66 @@
...
+        sort=1,
+      ),
+    },
+
+    instance: {
+      docs: 'Available instances (nodes)',
+      spec: grafana.template.new(
+        'instance',
+        '$prometheus_datasource',
+        'label_values(machine_scrape_error{job=~"$job"}, instance)',
```
v-zhuravlev, line 42: Sometimes variables depend on config (instead of hardcoded job=~"$job" it could be dynamic label selector). Does this packaging provide a way to provide custom labelsSelector from config?

malcolmholmes: This is entirely up to us to define, so I'm sure it can!

docker-mixin/dashboards/docker/queries.libsonnet
```diff
@@ -0,0 +1,45 @@
+local host_matcher = 'job=~"$job", instance=~"$instance"';
```
v-zhuravlev, line 1: Quite common that query selectors depend on config (instead of hardcoded job=~"$job" it could be user provided label selector). Does this packaging provide a way to provide custom host_matcher from config?

v-zhuravlev:
I think queries are better packed as targets (query + legend)

malcolmholmes:
I've added a lib (in the `/lib` directory), so execution now requires either:
```
jsonnet -J vendor -J lib polly.libsonnet
```

or:

```
jsonnet -J vendor -J lib m.jsonnet
```

Duologic (commented):
I'd love to see a more abstract description of the format, it's hard to untangle the format from the data in this PR.

nikimanoledaki:
Closing older PRs in this repo that seem unmaintained, sorry folks

PR #1002, opened 2023-06-29
Adding ArgoCD mixin

ishanjainn:
The ArgoCD mixin contains the following dashboards :

- ArgoCD

and the following alerts:

- ArgoAppOutOfSync
- ArgoAppSyncFailed
- ArgoAppMissing

Screenshot 2023-06-29 at 11 51 41 AM

### DIFF
--- a/argocd-mixin/alerts/argocd-alerts.yml
+++ b/argocd-mixin/alerts/argocd-alerts.yml
@@ -1,0 +1,31 @@
+groups:
+    - name: ArgoCD                                              
+      rules:
+        - alert: ArgoAppOutOfSync                               
+          expr: argocd_app_info{sync_status="OutOfSync"} == 1   
+          for: 1m                                               
+          labels:                                               
+            severity: warning
+          annotations:                                          
+            summary: '{{ $labels.name }} is OutOfSync' 
+            description: '{{ $labels.name }} Application has sync status as {{ $labels.sync_status }}'
+        
+        - alert: ArgoAppSyncFailed                              
+          expr: argocd_app_sync_total{phase!="Succeeded"} == 1  
+          for: 1m                                               
+          labels:                                               
+            severity: warning
+          annotations:                                          
+            summary: '{{ $labels.name }} Sync Failed' 
+            description: '{{ $labels.name }} Application has sync phase as {{ $labels.phase }}'
+        
+        - alert: ArgoAppMissing                                 
+          expr: absent(argocd_app_info)                         
+          for: 15m
+          labels:                                               
+            severity: critical
+          annotations:                                          
+            summary: "[ArgoCD] No reported applications"
+            description: >
+              ArgoCD has not reported any applications data for the past 15 minutes which
+              means that it must be down or not functioning properly.  
--- a/argocd-mixin/dashboards/argocd-overview.json
+++ b/argocd-mixin/dashboards/argocd-overview.json
@@ -1,0 +1,2909 @@
+{
+    "annotations": {
+      "list": [
+        {
+          "builtIn": 1,
+          "datasource": {
+            "type": "datasource",
+            "uid": "grafanacloud-prom"
+          },
+          "enable": true,
+          "hide": true,
+          "iconColor": "rgba(0, 211, 255, 1)",
+          "name": "Annotations & Alerts",
+          "type": "dashboard"
+        }
+      ]
+    },
+    "description": "",
+    "editable": true,
+    "fiscalYearStartMonth": 0,
+    "gnetId": 14584,
+    "graphTooltip": 1,
+    "id": 55,
+    "links": [],
+    "liveNow": false,
+    "panels": [
+      {
+        "datasource": {
+          "uid": "$datasource"
+        },
+        "gridPos": {
+          "h": 5,
+          "w": 2,
+          "x": 0,
+          "y": 0
+        },
+        "id": 121,
+        "links": [],
+        "options": {
+          "code": {
+            "language": "plaintext",
+            "showLineNumbers": false,
+            "showMiniMap": false
+          },
+          "content": "![argoimage](https://cncf-branding.netlify.app/img/projects/argo/stacked/color/argo-stacked-color.png)",
+          "mode": "markdown"
+        },
+        "pluginVersion": "10.0.1-cloud.3.f250259e",
+        "targets": [
+          {
+            "datasource": {
+              "uid": "$datasource"
+            },
+            "refId": "A"
+          }
+        ],
+        "transparent": true,
+        "type": "text"
+      },
+      {
+        "datasource": {
+          "type": "prometheus",
+          "uid": "$datasource"
+        },
+        "description": "Uptime of ArgoCD",
+        "fieldConfig": {
+          "defaults": {
+            "color": {
+              "mode": "continuous-BlPu"
+            },
+            "mappings": [],
+            "thresholds": {
+              "mode": "absolute",
+              "steps": [
+                {
+                  "color": "green",
+                  "value": null
+                },
+                {
+                  "color": "red",
+                  "value": 80
+                }
+              ]
+            },
+            "unit": "dtdurations"
+          },
+          "overrides": []
+        },
+        "gridPos": {
+          "h": 4,
+          "w": 3,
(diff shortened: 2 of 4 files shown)


olegmayko:
Hi, nice, any ETA on merging it? I'd suggest few alerts in the following PR then.

ishanjainn:
@olegmayko The goal os to get it merged by this week end so feel free to add alerts

argocd-mixin/alerts/argocd-alerts.yml
```diff
@@ -0,0 +1,31 @@
...
+    - name: ArgoCD                                              
+      rules:
+        - alert: ArgoAppOutOfSync                               
+          expr: argocd_app_info{sync_status="OutOfSync"} == 1   
+          for: 1m                                               
+          labels:                                               
+            severity: warning
+          annotations:                                          
+            summary: '{{ $labels.name }} is OutOfSync' 
+            description: '{{ $labels.name }} Application has sync status as {{ $labels.sync_status }}'
```
rgeyer, line 11: Linting rules state that the summary should not have templates in it, and that the description should use sentence case (capital first letter, and end in a period).

argocd-mixin/dashboards/argocd-overview.json
```diff
@@ -0,0 +1,2932 @@
...
+      "type": "timeseries"
+    }
+  ],
+  "refresh": "",
+  "schemaVersion": 38,
+  "style": "dark",
+  "tags": [
+    "argocd-integration"
+  ],
+  "templating": {
```
rgeyer, line 2693: Best practice is to have a template variable for the `job` and `instance` label, allowing you to filter the dashboard to particular instances of the SUO.

We may need to discuss this as I see that you have a couple different job labels in use, and may be using them to target specific processes.

argocd-mixin/dashboards/argocd-overview.json
```diff
@@ -0,0 +1,2932 @@
...
+        },
+        "datasource": {
+          "type": "prometheus",
+          "uid": "${datasource}"
+        },
+        "definition": "label_values(argocd_cluster_info,cluster)",
+        "hide": 0,
+        "includeAll": true,
+        "multi": false,
+        "name": "cluster",
```
rgeyer, line 2790: In Grafana Cloud for the k8s app we have settled on `cluster` being a reserved label for the k8s cluster.

Could you rename this `argocd_cluster` or something similar which specifies ArgoCD?

We'd also want to have a separate template for the label `cluster` which would filter by the k8s cluster.

So, for example, you may want to observe the `dev` ArgoCD cluster, in the `us-east-0` k8s cluster. These template variables would allow that granularity.

argocd-mixin/dashboards/argocd-overview.json
```diff
@@ -0,0 +1,2932 @@
...
+      ],
+      "title": "gRPC stream message rate",
+      "transparent": true,
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "grafanacloud-prom"
+      },
```
rgeyer, line 2494: There are several panels which are not using the templated data source, but are hardcoded like this.

rgeyer (changes requested):
There are several things that the dashboard-linter/mixtool are catching. Some are minor or could be ignored, but some will need to be fixed.

I pointed out a few in separate comments to help with some of the common patterns which are being caught.

If you want to add the following Makefile, and install [mixtool](https://github.com/monitoring-mixins/mixtool) you can run the linter and see what's being called out.

```
JSONNET_FMT := jsonnetfmt -n 2 --max-blank-lines 1 --string-style s --comment-style s

.PHONY: all
all: dashboards_out prometheus_alerts.yaml

.PHONY: fmt
fmt:
        find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
                xargs -n 1 -- $(JSONNET_FMT) -i

.PHONY: lint
lint:
        find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
                while read f; do \
                        $(JSONNET_FMT) "$$f" | diff -u "$$f" -; \
                done
        mixtool lint mixin.libsonnet

dashboards_out: mixin.libsonnet config.libsonnet $(wildcard dashboards/*)
        @mkdir -p dashboards_out
        mixtool generate dashboards mixin.libsonnet -d dashboards_out

prometheus_alerts.yaml: mixin.libsonnet alerts/*.libsonnet
        mixtool generate alerts mixin.libsonnet -a prometheus_alerts.yaml

.PHONY: clean
clean:
        rm -rf dashboards_out prometheus_alerts.yaml
```

argocd-mixin/alerts/argocd-alerts.yml
ishanjainn: Updated, Thanks!

argocd-mixin/dashboards/argocd-overview.json
ishanjainn: I am using the k8s integration in my cluster which is why we see the cluster variable here, ArgoCD on its own isnt exporting a cluster label

argocd-mixin/dashboards/argocd-overview.json
ishanjainn: My bad, Thanks!

argocd-mixin/dashboards/argocd-overview.json
ishanjainn: ArgoCD has quite a few services, I opted to use a different job label for each in my agent scrape config. I can make it scrape all under the same job label although it might not be the best way in my opinion. Open to suggestions here

rgeyer: Ahh, on closer inspection, I see that you are filtering by namespace, and you have a "grouping" option as well.

So, in a case where there was more than one "instance" of ArgoCD in a k8s cluster, they would likely be deployed into different namespaces, and therefore you would be able to filter that way?

The grouping seems to further narrow down which ArgoCD instance you're observing, yes?

If so, the approach you're using is fine, and job/instance are not necessary. When we build the Grafana Cloud integration (or k8s app instructions) we'll want to be explicit about how to setup the scrape jobs to match what the mixin is expecting.

rgeyer: If I'm correct about my assertions above, the following `.lint` file will allow the linter to pass, even with rules pertaining to job and instance failing. Just put it in a file named `.lint` at the root of the project.

```exclusions:
  template-instance-rule:
    reason: ArgoCD instance filtering is accomplished by the namespace and grouping template variables
  template-job-rule:
    reason: ArgoCD instance filtering is accomplished by the namespace and grouping template variables    
  target-instance-rule:
    reason: ArgoCD instance filtering is accomplished by the namespace and grouping template variables
  target-job-rule:
    reason: ArgoCD instance filtering is accomplished by the namespace and grouping template variables
  panel-units-rule:
    reason: Panel units are self descriptive
```

argocd-mixin/makefile
```diff
@@ -0,0 +1,27 @@
...
+
+.PHONY: lint
+lint:
+		find . -name 'vendor' -prune -o -name '*.libsonnet' -print -o -name '*.jsonnet' -print | \
+			while read f; do \
+					$(JSONNET_FMT) "$$f" | diff -u "$$f" -; \
+            done
+		mixtool lint mixin.libsonnet
+
+dashboards_out: mixin.libsonnet config.libsonnet $(wildcard dashboards/*)
```
rgeyer, line 18: My mistake, I left some superfluous config in this file. Thank you for adding it!

```suggestion
dashboards_out: mixin.libsonnet $(wildcard dashboards/*)
```

rgeyer:
I just actually took the time to "get smart" about the ArgoCD deployment model. You are correct, there is only ever one ArgoCD per cluster, with the option for an overarching cluster manager. - https://akuity.io/blog/argo-cd-architectures-explained/

As such, these filters are more than adequate, and allows the most granular filtering. With the `.lint` file, we're good to go.

Thanks for being patient with me getting up to speed on Argo ;)

rgeyer (approved):
LGTM!

PR #1007, opened 2023-07-06
Add `latency` alerts

olegmayko:
Those alerts will help to see if there is an issue with syncying and applying the changes to the cluster.


olegmayko:
@ishanjainn  should i create another PR to `main` or don't you think it's a useful addition to the argocd mixin?
thx

PR #1009, opened 2023-07-06
Add with_otel_resource_attrs option to jaeger-agent-mixin

leveyjam:
Flag causes service.namespace to be included in span attributes so Tempo can generate job labels


jvrplmlmn (approved):

mar4uk:
nice! like the approach

mar4uk (approved):

PR #1013, opened 2023-07-10
Nginx Mixin Update

rgeyer:
This updates the nginx mixin to use the latest revision of the dashboard published by @wardbekker.

The makefile now downloads the specified revision directly from the Grafana dashboards library, and then patches a few things over the top.

The information panels are removed, leaving only the panels with data, which use the loki data source.

This also adds a `cluster` variable selector which allows the mixin to be used with the Grafana Cloud kubernetes monitoring app. The cluster label name is configurable in `config.libsonnet`.

Lastly, a few minor nits which cause errors on initial load of the dashboard are corrected.


v-zhuravlev (approved):
LGTM, nice mods functions

PR #1015, opened 2023-07-11
etcd-operator :: Add RBAC policy to allow managing PodDisruptionBudget

dylwylie:
The upstream etc-operator has been archived but forked versions allow managing PDBs, add the necessary policy rule so the operator can apply these.


jvrplmlmn (approved):
👍

PR #1018, opened 2023-07-12
Add mixin for Harbor

ishanjainn:


harbor-mixin/README.md
```diff
@@ -0,0 +1,39 @@
...
+- Harbor
+
+and the following alerts:
+
+- HarborComponentStatus
+- HarborProjectQuataExceeded
+- HarborHighErrorRate
+
+## Harbor Dashboard Overview
+Harbor dashbaord provides details on the overall status of the Harbor repository including the details about each project and artifcats. The dashboard includes visualizations for Harbor's Components metrics, along with key visualiozation like number of project, images, pulled images etc.
```
rgeyer, line 16:
```suggestion
Harbor dashboard provides details on the overall status of the Harbor repository including the details about each project and artifacts. The dashboard includes visualizations for Harbor's Components metrics, along with key visualization like number of project, images, pulled images etc. 
```

harbor-mixin/dashboards/harbor-overview.json
```diff
@@ -0,0 +1,2485 @@
...
+      },
+      "pluginVersion": "10.0.2-cloud.1.94a6f396",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "sum(harbor_project_total{job=~\"$job\", cluster=~\"$cluster\"})",
```
rgeyer, line 316: Is sum aggregating away a label here? Would it be useful to have a template variable that allowed you to filter on that/those labels instead?

harbor-mixin/dashboards/harbor-overview.json
```diff
@@ -0,0 +1,2485 @@
...
+      },
+      "pluginVersion": "10.0.2-cloud.1.94a6f396",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "harbor_artifact_pulled{job=~\"$job\", project_name=~\"$Project\", cluster=~\"$cluster\"}",
```
rgeyer, line 519: Similar feedback to Velero here. Is it useful to simply show the lifetime totals for these? Would it be better to show the increase for the currently selected time range?

rgeyer (commented):
Looks good, a couple of nits and questions.

harbor-mixin/dashboards/harbor-overview.json
ishanjainn: Sorry, I dont follow, Could you share an example

harbor-mixin/dashboards/harbor-overview.json
ishanjainn: Yup, Totally makes sense, Updated now

harbor-mixin/dashboards/harbor-overview.json
rgeyer: More of a broad question about the metrics, which I am not super familiar with. Are there other labels besides `job` and `cluster` on `harbor_project_total`?

For example is there a `namespace` label (total SWAG, probably a bad example) label which it would be useful to group by for this?

Likely not, but, just checking to see if there's any useful labels/dimensions which are getting aggregated away when you `sum`

ishanjainn: just the `public` label which mentions if the project is public or private

rgeyer: Okay, that *might* be a useful template variable, but I'll leave it to your discretion. You could also use some transforms to display public and private separately there, or just add queries which are explicit. I.E. `sum(harbor_project_total{..., public=m|true})` with a legend of `Public`.

If there isn't any significant value in differentiating, you can just leave this be.

harbor-mixin/dashboards/harbor-overview.json
```diff
@@ -0,0 +1,2485 @@
...
+      },
+      "pluginVersion": "10.0.2-cloud.1.94a6f396",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "$datasource"
+          },
+          "editorMode": "code",
+          "expr": "increase(harbor_artifact_pulled{job=~\"$job\", project_name=~\"$Project\", cluster=~\"$cluster\"}[$__rate_interval])",
```
rgeyer, line 519: I'd actually use `[$__range]` here, so that it shows the number of artifacts pulled for the selected timerange (say 1 hour, or 2 days, whatever).

As written, it will show the number of artifacts pulled in the last ~5min, which might be useful on a timeseries, but for a bargauge showing lastNotNull it's less useful.

```suggestion
          "expr": "increase(harbor_artifact_pulled{job=~\"$job\", project_name=~\"$Project\", cluster=~\"$cluster\"}[$__range])",
```

harbor-mixin/dashboards/harbor-overview.json
ishanjainn: yeah, my thoughts kinda divided on this,  Thinking of keeping it as is for now and later see if there are any particular requests for it and update.

rgeyer (approved):
After a rebase/merge with main, this one looks good to go as well.

Great work! 🎉 🚀

PR #1020, opened 2023-07-13
kubescape mixin

ishanjainn:


rgeyer (commented):
This looks good. Great job integrating the linter and learnings from your previous mixin!

Also appreciate the choice to show stat panels and pie graphs here rather than timeseries which would likely prove less useful.

I don't have technical feedback, but do have some design questions/guidance.

How would an operator use this dashboard?

In the screenshot, for example the "Controls passed" seem to be low percentages <50%. Is that good? Is that bad? Could a threshold be set?

If it *is* bad, what action would the operator take?

Is there an opportunity for a table which shows specifics (if available, in labels) about failing controls, so that an operator can find them? Or is that sort of discovery/troubleshooting better handled in the tool itself?

ishanjainn:
Thanks Ryan!

For the first bit, Controls passed is defineitely low, I had installed kubescape on an open cluster which is the reason. I perosnally would like them to be atleast around the 75-80% mark, I do have an alert setup which triggers at 50% though.

Screenshot 2023-07-18 at 2 00 39 PM

I do have this table at the end for the operator to quickly view the failing control with severity and the CVE link.

rgeyer (approved):
LGTM! Thank you 🚀

PR #1021, opened 2023-07-18
Adds optional k8s related cluster selector to dashboards for Apache Cassandra mixin

StefanKurek:
Setting up the work to handle this [issue](https://github.com/grafana/cloud-onboarding/issues/4217) in the cloud-onboarding repo.

Updates the existing cluster selector to be renamed as "cassandra_cluster". 
Adds a new config to enable the k8s related cluster selector. 
Modifies dashboards to utilize this new selector if the config is enabled.

### DIFF
--- a/apache-cassandra-mixin/README.md
+++ b/apache-cassandra-mixin/README.md
@@ -28,7 +28,7 @@ The Apache Cassandra overview dashboard provides details on number of clusters, 
 
 ## Apache Cassandra Nodes
 
-The Apache Cassandra nodes dashboard provides details on disk/memory/cpu usage, garbage collections, number of pending/blocked compaction tasks, number and latency of reads/writes, and logs for a specific node in the cluster. To get Cassandra system logs, [Promtail and Loki needs to be installed](https://grafana.com/docs/loki/latest/installation/) and provisioned for logs with your Grafana instance. The default Cassandra system log path is `/var/log/cassandra/system.log` on Linux.
+The Apache Cassandra nodes dashboard provides details on disk/memory/cpu usage, garbage collections, number of pending/blocked compaction tasks, number and latency of reads/writes, and logs for a specific node in the Cassandra cluster. To get Cassandra system logs, [Promtail and Loki needs to be installed](https://grafana.com/docs/loki/latest/installation/) and provisioned for logs with your Grafana instance. The default Cassandra system log path is `/var/log/cassandra/system.log` on Linux.
 
 ![First screenshot of the Apache Cassandra nodes dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-cassandra/screenshots/nodes_1.png)
 ![Second screenshot of the Apache Cassandra nodes dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-cassandra/screenshots/nodes_2.png)
@@ -53,7 +53,7 @@ scrape_configs:
         labels:
           job: integrations/apache-cassandra
           instance: '<your-instance-name>'
-          cassandra_cluster: '<your-cluster-name>'
+          cassandra_cluster: '<your-cassandra-cluster-name>'
           __path__: /var/log/cassandra/system.log
 ```
 
--- a/apache-cassandra-mixin/README.md
+++ b/apache-cassandra-mixin/README.md
@@ -43,7 +43,7 @@ Cassandra system logs are enabled by default in the `config.libsonnet` and can b
 }
 ```
 
-In order for the selectors to properly work for system logs ingested into your logs datasource, please also include the matching `instance`, `job`, and `cluster` labels onto the [scrape_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
+In order for the selectors to properly work for system logs ingested into your logs datasource, please also include the matching `instance`, `job`, and `cassandra_cluster` labels onto the [scrape_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
 
 ```yaml
 scrape_configs:
@@ -53,7 +53,7 @@ scrape_configs:
         labels:
           job: integrations/apache-cassandra
           instance: '<your-instance-name>'
-          cluster: '<your-cluster-name>'
+          cassandra_cluster: '<your-cluster-name>'
           __path__: /var/log/cassandra/system.log
 ```
 
--- a/apache-cassandra-mixin/alerts/alerts.libsonnet
+++ b/apache-cassandra-mixin/alerts/alerts.libsonnet
@@ -97,7 +97,7 @@
           {
             alert: 'UnavailableWriteRequestsFound',
             expr: |||
-              sum(cassandra_clientrequest_unavailables_count{clientrequest="Write"}) by (cluster) > %(alertsCriticalUnavailableWriteRequests5m)s
+              sum(cassandra_clientrequest_unavailables_count{clientrequest="Write"}) by (cassandra_cluster) > %(alertsCriticalUnavailableWriteRequests5m)s
             ||| % $._config,
             'for': '5m',
             labels: {
--- a/apache-cassandra-mixin/config.libsonnet
+++ b/apache-cassandra-mixin/config.libsonnet
@@ -1,5 +1,9 @@
 {
   _config+:: {
+    enableMultiCluster: false,
+    cassandraSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
+    multiclusterSelector: 'job=~"$job"',
+
     dashboardTags: ['apache-cassandra-mixin'],
     dashboardPeriod: 'now-1h',
     dashboardTimezone: 'default',
--- a/apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
+++ b/apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
@@ -708,13 +708,13 @@ local keyspaceReadLatencyPanel(matcher) = {
       'sum(cassandra_keyspace_readlatency_seconds{' + matcher + ', quantile="0.95"} >= 0) by (keyspace)',
       datasource=promDatasource,
       legendFormat='{{ keyspace }} - p95',
-      format='heatmap',
+      format='time_series',
     ),
     prometheus.target(
       'sum(cassandra_keyspace_readlatency_seconds{' + matcher + ', quantile="0.99"} >= 0) by (keyspace)',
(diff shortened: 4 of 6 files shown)


apache-cassandra-mixin/README.md
```diff
@@ -53,7 +53,7 @@ scrape_configs:
         labels:
           job: integrations/apache-cassandra
           instance: '<your-instance-name>'
-          cluster: '<your-cluster-name>'
+          cassandra_cluster: '<your-cluster-name>'
```
schmikei, line 56: Seems like the rename of `cluster` => `cassandra_cluster` is a good move. 

I understand that the `cluster` label is reserved and that is the motivating factor behind this effort, do we have a good understanding of when the label gets populated?

apache-cassandra-mixin/config.libsonnet
```diff
@@ -1,5 +1,9 @@
 {
   _config+:: {
+    enableMultiCluster: true,
```
schmikei, line 3: This will probably have inclinations of the integration, just a heads up by default we will have a cluster label when importing and using this mixin when its updated.

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
```diff
@@ -708,13 +708,13 @@ local keyspaceReadLatencyPanel(matcher) = {
...
       legendFormat='{{ keyspace }} - p95',
-      format='heatmap',
+      format='time_series',
     ),
     prometheus.target(
       'sum(cassandra_keyspace_readlatency_seconds{' + matcher + ', quantile="0.99"} >= 0) by (keyspace)',
       datasource=promDatasource,
       legendFormat='{{ keyspace }} - p99',
-      format='heatmap',
+      format='time_series',
```
schmikei, line 717: Same question here

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
```diff
@@ -882,7 +896,7 @@ local getMatcher(cfg) = 'job=~"$job", instance=~"$instance", keyspace=~"$keyspac
             template.new(
               'rack',
               promDatasource,
-              'label_values(cassandra_cache_size, rack)',
+              'label_values(cassandra_cache_size{%(cassandraSelector)s, datacenter="$datacenter"}, rack)' % $._config,
```
schmikei, line 899: We will want to add the check for the `config.enableDatacenterLabel` value before putting datacenter here.

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
```diff
@@ -708,13 +708,13 @@ local keyspaceReadLatencyPanel(matcher) = {
       'sum(cassandra_keyspace_readlatency_seconds{' + matcher + ', quantile="0.95"} >= 0) by (keyspace)',
       datasource=promDatasource,
       legendFormat='{{ keyspace }} - p95',
-      format='heatmap',
```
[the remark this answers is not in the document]
StefanKurek: @schmikei yeah. I forgot to list that (the original dashboard was giving me an error until i made this match up with the "writelatency" version of this panel. I can't imagine this hurting anything as the next line basically states that it's a timeseries anyways.

apache-cassandra-mixin/alerts/alerts.libsonnet
```diff
@@ -97,7 +97,7 @@
           {
             alert: 'UnavailableWriteRequestsFound',
             expr: |||
-              sum(cassandra_clientrequest_unavailables_count{clientrequest="Write"}) by (cluster) > %(alertsCriticalUnavailableWriteRequests5m)s
+              sum(cassandra_clientrequest_unavailables_count{clientrequest="Write"}) by (cassandra_cluster) > %(alertsCriticalUnavailableWriteRequests5m)s
```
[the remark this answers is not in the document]
StefanKurek: @schmikei was considering it...but i don't actually have valid data right now so i wasn't confident specifying all of the labels to "not" include

apache-cassandra-mixin/README.md
StefanKurek: @schmikei i believe when you use the integration through the kubernetes plugin then it gets added automatically (i could be wrong).

apache-cassandra-mixin/config.libsonnet
StefanKurek: @schmikei good catch. this was never meant to be pushed as "true". i switched it while testing, but forgot to switch it back to "false". i'll change.

apache-cassandra-mixin/README.md
schmikei: I guess my concern is communicating for people not using the k8s plugin if this will work OOTB. Will look into that a tad bit closer on my 2nd pass :)

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
StefanKurek: @schmikei i guess you're right. good catch. it would be pretty dumb to enable the rack label but not the datacenter label. but i guess it's possible.

StefanKurek: updated on all 3 dashboards

apache-cassandra-mixin/config.libsonnet
```diff
@@ -15,7 +19,7 @@
     alertsCriticalHighCpuUsage5m: 80,  //percent: emitted metric has range 0-100
     alertsCriticalHighMemoryUsage5m: 80,  //percent: calculated as ratio then multiplied by query
 
-    enableLokiLogs: true,
+    enableLokiLogs: false,
```
schmikei, line 22: I think we want logs enabled by default?

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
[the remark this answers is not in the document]
schmikei: I think we liked the look with the heatmap approach if I recall correctly. Could you clarify what the error you were encountering was?

apache-cassandra-mixin/README.md
schmikei: I guess for people not using the plugin was more of my concern. Which should be more on the integration side I believe.

apache-cassandra-mixin/config.libsonnet
StefanKurek: @schmikei wow, that's just a dum dum mistake from moving too fast. thank you for noticing 😄 . the correct config value should be updated now

apache-cassandra-mixin/alerts/alerts.libsonnet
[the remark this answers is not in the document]
schmikei: SGTM 👍

apache-cassandra-mixin/README.md
schmikei: After re-reading this I think it could benefit a little from rephrasing 😅 

My question was more of people who are using cluster currently will be affected by this. It feels like this **IS** a correct approach on renaming to use `cassandra_cluster` and that is what we want users to do. So does that warrant a documentation change somewhere was the primary thing I wanted to get clarification on? Or if we are okay sending the correct implementation with an update to the integration's documentation?

schmikei (approved):

apache-cassandra-mixin/README.md
StefanKurek: Keeping this visible for Grafana reviewers.

Dasomeone (approved):
Generally looks good, just some minor comments :)

apache-cassandra-mixin/dashboards/cassandra-keyspaces.libsonnet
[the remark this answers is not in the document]
StefanKurek: @Dasomeone to be fair, it has not. I changed this because I was seeing this error for this panel on the original dashboard with no current data being pulled.
image
Looking at the fields for the sister "writelatency" panel, I noticed that the format was set to "time_series" which matched the panel "type" field. For some reason this one had this inconsistency (it's supposed to be a timeseries and not a heatmap).

apache-cassandra-mixin/config.libsonnet
```diff
@@ -1,5 +1,9 @@
 {
   _config+:: {
+    enableMultiCluster: false,
+    cassandraSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
```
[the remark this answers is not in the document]
Dasomeone: Yeah no need to remove it, just wanted to point it out in case you ran into issues later :)

apache-cassandra-mixin/README.md
gloriasc: it would definitely need an docs update - however this would happen automatically as long as we make the changes in the integrations files in cloud-onboarding (agent config), right ? @Dasomeone

gloriasc (approved):
Looks good from what I can see. - @Dasomeone do you think we should get a second dev to review? 
Also we'd need to update the cloud-onboarding integration files as @schmikei identified, @StefanKurek can you add a PR for that here?

StefanKurek:
> 

@gloriasc I was assuming that when the 2nd part of this (making an update to the cloud-onboarding repo to add a ServiceMonitor) would be the right time to add any sort of documentation changes there (because I thought that the integration was locked to a specific version of the mixin and wouldn't automatically get updated with mixin changes). Or am I wrong there?

apache-cassandra-mixin/README.md
Dasomeone: The change would also need to be made in the scrape configuration, at which point the CI would demand new docs generation as well :)

Dasomeone:
> > 
> 
> @gloriasc I was assuming that when the 2nd part of this (making an update to the cloud-onboarding repo to add a ServiceMonitor) would be the right time to add any sort of documentation changes there (because I thought that the integration was locked to a specific version of the mixin and wouldn't automatically get updated with mixin changes). Or am I wrong there?

Correct, the integration does **not** automatically update based on mixin changes, so any integration specific docs should be done then.
However, mixin specific documentation should and can be done at this stage :)

Dasomeone:
> Looks good from what I can see. - @Dasomeone do you think we should get a second dev to review? Also we'd need to update the cloud-onboarding integration files as @schmikei identified, @StefanKurek can you add a PR for that here?

As for a second dev, I think we're good to go. Worst case we catch something while testing the integration changes and have to do a small fix here then

StefanKurek:
@Dasomeone added the README change to try and more explicitly have references to Cassandra clusters. Hopefully should be good to go now

gloriasc:
Good work everyone :)

PR #1028, opened 2023-07-24
Integrations common lib

v-zhuravlev:
1) This commonlib implements 
the following panel groups:
`generic` as base group, then:
  - `system`
  - `cpu`
  - `memory`
  - `disk`
  - `network`

Inside each group, the following panel types can be found:
  - timeSeries
  - table
  - statusHistory
  - stat
2) Also implements common grafana annotations prototypes (such as reboot event)
3) Adds some util functions to manipulate labels and variables

All panels have `new()` method to create this type of panel and `stylize()` method now that can be used to 'mutate' old panels or change style of existing panels.
You can also use it as `stylize(allLayers=false)` to ignore base styles and apply only top layer. This can be handy when you want to cherry-pick different style sets from different panels to compose new combination.

For actual use of this lib see #1050.


common-lib/common/g.libsonnet
```diff
@@ -0,0 +1 @@
+import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet'
```
malcolmholmes, line 1: I would suggest pinning to a specific Grafana version and let Grafana handle upgrading.

common-lib/common/panels.libsonnet
```diff
@@ -0,0 +1,6 @@
+local g = import 'g.libsonnet';
+
+{
+  stat: import 'panels/stat/main.libsonnet',
```
malcolmholmes, line 4: does this import work? Won't it be looking in the `JSONNET_PATH` for this lib? I'd say `./panels/...` would be better.

common-lib/common/panels/all/base.libsonnet
```diff
@@ -0,0 +1,13 @@
...
+local timeSeries = g.panel.timeSeries;
+local fieldOverride = g.panel.timeSeries.fieldOverride;
+local custom = timeSeries.fieldConfig.defaults.custom;
+local defaults = timeSeries.fieldConfig.defaults;
+local options = timeSeries.options;
+{
+  new(title, targets, description=''):
+    timeSeries.queryOptions.withInterval('1m')
+    + timeSeries.queryOptions.withTargets(targets)
+    + timeSeries.panelOptions.withDescription(description),
```
rgeyer, line 12: These feel out of place to me. Tho I may be struggling with the grafonnet object model more than anything.

It seems like we're applying these properties to the entire panel, where they should be applied to each query/target individually.

Why are we dictating 1m interval? I suppose the purpose of this library is to be opinionated. 🤔 If we *do* set an interval for an entire panel, it should be a `Mixin` type function that recursively and retroactively applies to all queries/targets which have been added.

common-lib/common/panels/stat/base.libsonnet
```diff
@@ -0,0 +1,10 @@
+local g = import '../../g.libsonnet';
+local stat = g.panel.stat;
+local base = import '../all/base.libsonnet';
+
+stat + base {
+  new(title, targets, description=''):
+    stat.new(title)
+    + super.new(title, targets, description),
```
rgeyer, line 8: `title` doesn't seem to ever be used in the base class, since the panel type is not yet known.

Probably makes sense to drop title from the all panels base.

common-lib/common/panels/stat/percentage.libsonnet
```diff
@@ -0,0 +1,16 @@
+local g = import '../../g.libsonnet';
+local base = import './base.libsonnet';
+local stat = g.panel.stat;
+// Panels to display metrics that can go from 0 to 100%.
+// (cpu utilization, memory utilization etc).
+base {
+  new(title, targets, description=''):
+    super.new(title, targets, description)
+    + stat.standardOptions.withDecimals(1)
+    + stat.standardOptions.withUnit('percent')
```
rgeyer, line 10: nit; Would probably want a parameter to allow for either percent, or percentunit.

common-lib/common/panels/timeSeries/network/base.libsonnet
```diff
@@ -0,0 +1,26 @@
...
+  new(
+    title,
+    targets,
+    description=''
+  ):
+    super.new(title, targets, description)
+    + self.standardOptions.withDecimals(1)
+    + self.standardOptions.withUnit('pps'),
+
+  withNegateOutPackets(regexp='/transmit|tx|out/'):
```
rgeyer, line 18: Love this. It's such a visually expressive way to show network packets, and it's a lot of copy/paste or UI toil to set these up generally. Having it in a reusable library/pattern is great.

rgeyer (commented):
Several questions, mostly open-ended for the sake of discussion.

I generally like the idea that we're being opinionated, as that should be the purpose of this library. This will require fairly comprehensive documentation to explain the actual result of these assumptions/opinions. For example, in the stats info panel, I agree with the assertions, but the description of "simple info panel with text or count" falls short in fully describing what type of panel you get, and when you'd use it.

An illustrated guide with screenshots and use-cases would likely be appropriate.

v-zhuravlev:
> Several questions, mostly open-ended for the sake of discussion.
> 
> I generally like the idea that we're being opinionated, as that should be the purpose of this library. This will require fairly comprehensive documentation to explain the actual result of these assumptions/opinions. For example, in the stats info panel, I agree with the assertions, but the description of "simple info panel with text or count" falls short in fully describing what type of panel you get, and when you'd use it.
> 
> An illustrated guide with screenshots and use-cases would likely be appropriate.

Agree, more detailed descriptions are necessary.

common-lib/common/panels/all/base.libsonnet
v-zhuravlev: This is a tricky part.
1. If interval is not defined at all, it will be sort of 'none' because panel's common datasource is 'Mixed'. Setting it one 1m here, is sort of hack, that i'll try to address.

rgeyer: Ahh, I see! I think I experienced this on my hackathon project where my timeseries panel was not rendering or behaving as expected. This was very likely the cause.

Maybe having interval as an optional parameter, with documentation describing the reason it needs to be set (as you described above) would be the best approach.

common-lib/common/panels.libsonnet
```diff
@@ -0,0 +1,14 @@
+local g = import 'g.libsonnet';
```
rgeyer, line 1: Not strictly necessary, but the more explicit `import './g.libsonnet'` will ensure that the correct file is always used, and gives the user a hint that this is a local file in this library.

```suggestion
local g = import './g.libsonnet';
```

common-lib/common/panels/all/stat/info.libsonnet
```diff
@@ -0,0 +1,19 @@
+local g = import '../../../g.libsonnet';
+local base = import './base.libsonnet';
+local stat = g.panel.stat;
+//simple info panel with text or count
+base {
+  new(title, targets, description=''):
+    super.new(title, targets, description)
+    // Style choice: Make it bigger
```
rgeyer, line 8: Can these style choices be overridden? Perhaps they could be encapsulated in an additional `stylize` function which adds them all, which could then also be used on dashboards/panels created with grafonnet directly?

v-zhuravlev: well, everything can be overriden, since it is jsonnet. We just providing defaults we think best.

common-lib/common/panels.libsonnet
v-zhuravlev: yes, thanks!

common-lib/common/panels/all/stat/info.libsonnet
v-zhuravlev: @rgeyer , good idea with stylize. Added in the latest commit.

Here is example how to transform panel created by other libs:
```
      idleCPU::
        nodePanels.timeseries.new(
          'CPU Usage',
          description='Total CPU utilisation percent.'
        )
        .withStacking('normal')
        .addTarget(c.commonPromTarget(
          expr=c.queries.cpuUsagePerCore,
          legendFormat='cpu {{cpu}}',
        ))
        + commonlib.all.timeSeries.base.stylize(),
```

common-lib/common/panels/all/base.libsonnet
v-zhuravlev: this one solved, see logic in panels/generic/base (tries to set panel's datasource based on target's datasource)

common-lib/common/panels/stat/base.libsonnet
v-zhuravlev: Dropped!

common-lib/common/panels/stat/percentage.libsonnet
v-zhuravlev: I'll keep it for now hardcoded and implement it later. For now,  you can always multiply metric by 100.

common-lib/common/panels/all/stat/info.libsonnet
v-zhuravlev: Implemented stylize().

All panels have `stylize()` method now that can be used to 'mutate' old panels or change style of existing panels.
You can also use it as `stylize(allLayers=false)` to ignore base styles and apply only top layer.  This can be handy when you want to cherry-pick different style sets from different panels to compose new combination.

common-lib/common/panels/disk/table/usage.libsonnet
```diff
@@ -0,0 +1,144 @@
...
+    // validate inputs
+    std.prune(
+      {
+        checks: [
+          if !(std.objectHas(totalTarget, 'format') && std.assertEqual(totalTarget.format, 'table')) then error 'totalTarget format must be "table"',
+          if !(std.objectHas(totalTarget, 'instant') && std.assertEqual(totalTarget.instant, true)) then error 'totalTarget must be type instant',
+          if !(std.objectHas(usageTarget, 'format') && std.assertEqual(usageTarget.format, 'table')) then error 'usageTarget format must be "table"',
+          if !(std.objectHas(usageTarget, 'instant') && std.assertEqual(usageTarget.instant, true)) then error 'usageTarget must be type instant',
+          // if std.length(std.findSubstr(groupLabel, totalTarget.expr)) == 0 then error 'totalTarget expression must be grouped by groupLabel "%s", current expression is %s' % [groupLabel, totalTarget.expr],
+          // if std.length(std.findSubstr(groupLabel, usageTarget.expr)) == 0 then error 'usageTarget expression must be grouped by groupLabel "%s", current expression is %s' % [groupLabel, totalTarget.expr],
```
rgeyer, line 27: Are these not necessary?

common-lib/common/panels/generic/base.libsonnet
```diff
@@ -0,0 +1,27 @@
...
+
+// This is base of ALL panels in the common lib
+{
+  new(targets, description=''):
+    // hidden field to hold styles modifiers
+
+    timeSeries.queryOptions.withTargets(targets)
+    + timeSeries.panelOptions.withDescription(description)
+    // set first target's datasource
+    // to panel's datasource if only sinlge type of
```
rgeyer, line 17:
```suggestion
    // to panel's datasource if only single type of
```

common-lib/common/panels/generic/timeSeries/percentage.libsonnet
```diff
@@ -0,0 +1,17 @@
+local g = import '../../../g.libsonnet';
+local timeSeries = g.panel.timeSeries;
+local base = import './base.libsonnet';
+// This panel can be used to display gauge metrics with possible values range 0-100%.
+// Examples: cpu utilization, memory utilization etc.
+base {
+  stylize(allLayers=true):
+    (if allLayers then super.stylize() else {})
+    // (if allLayers then super.stylize() else {})
```
rgeyer, line 9:
```suggestion
```

common-lib/common/panels/generic/timeSeries/topk_percentage.libsonnet
```diff
@@ -0,0 +1,69 @@
...
+    ),
+  stylize(allLayers=true):
+    (if allLayers then super.stylize() else {})
+    + generic.percentage.stylize(allLayers=false)
+    // make dots cloud
+    // + fieldConfig.defaults.custom.withLineStyleMixin(
+    // {
+    //   fill: 'dot',
+    //   dash: [0,40],
+    // })
```
rgeyer, line 45: Is this a WIP, or can it be removed?

rgeyer (commented):
Looking very good.. I have a few nits/questions. Enough that I'll refrain from approving just yet.

The biggest concern is that there seem to be several commits/changed files which don't relate to this PR.

Tremendous work! Thank you for the dedication to getting this finished!

aerospike-mixin/.lint
```diff
@@ -11,6 +11,7 @@ exclusions:
     reason: "The dashboard is a 'cluster' dashboard where the instance refers to nodes, this dashboard focuses only on the cluster view."
     entries:
       - dashboard: "Aerospike overview"
+      - dashboard: "Aerospike namespace overview"
   target-promql-rule:
     reason: "Linter does not support selector variable value as a scalar in top-k PromQL queries."
   template-label-promql-rule:
```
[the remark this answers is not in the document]
v-zhuravlev: Rebase helped. Should be clean now.

common-lib/common/annotations/fatal.libsonnet
```diff
@@ -0,0 +1,14 @@
...
+local annotation = g.dashboard.annotation;
+local base = import './base.libsonnet';
+
+// Show fatal or critical events as annotations
+base {
+  new(
+    title,
+    target,
+  ):
+    super.new(title, target)
```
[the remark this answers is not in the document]
v-zhuravlev: yes, you got it right. We just provide this annotation.fatal as prototype, where color choice is enforced for now.

common-lib/common/annotations/fatal.libsonnet
```diff
@@ -0,0 +1,14 @@
...
+    + annotation.withIconColor('light-purple')
```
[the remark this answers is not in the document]
v-zhuravlev: Just following colors used in grafana itself for autocoloring loki logs: https://github.com/grafana/grafana/blob/fe9f52097bb4717a7ebfac17a62f399a2ce0ba1b/docs/sources/explore/index.md#log-level

common-lib/common/annotations/reboot.libsonnet
```diff
@@ -0,0 +1,16 @@
...
+local annotation = g.dashboard.annotation;
+local base = import './base.libsonnet';
+
+base {
+  new(
+    title,
+    target,
+    instanceLabels,
+  ):
+    super.new(title, target)
```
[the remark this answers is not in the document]
v-zhuravlev: In windows yes! And it would be in linux and so on.

common-lib/common/panels/disk/table/usage.libsonnet
v-zhuravlev: updated

common-lib/common/panels/generic/timeSeries/topk_percentage.libsonnet
v-zhuravlev: Removed, thnks

common-lib/common/annotations/fatal.libsonnet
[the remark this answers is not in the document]
rgeyer: Ahh perfect! Still worth documenting it (again, in a future PR).

I have a bias with purple. It's my favorite color, so I don't immediately associate it with "bad things". But I'm always reminded that it's "darker" than red, and thus more severe. :)

rgeyer (approved):
Thank you for the regular updates and getting this sorted out. I think we're ready to merge this!

Again tremendous work! Great job!

PR #1029, opened 2023-07-24
Fix(Apache Tomcat): Add job and instance to apache-tomcat logs filter

schmikei:
If multiple tomcat logs are ingested outside of the context of the job, we will most likely want to not include them.


rgeyer (approved):
LGTM, thank you for jumping on this quickly!

PR #1031, opened 2023-07-25
Update kafka lag dashboard

v-zhuravlev:
Change basic metric in variables discovery: 
from
`kafka_consumergroup_current_offset`
to 
`kafka_topic_partition_current_offset`

There might be cases where topics are available but not consumergroups.


gaantunes (approved):
Makes sense, thanks for taking this!

PR #1039, opened 2023-07-31
Nginx: adding '{}' as stream filter to avoid 'cluster' being identified as a metric

gaantunes:


rgeyer (approved):
LGTM, thank you for handling this one!

PR #1042, opened 2023-08-09
Include all option for job in mongodb mixin

v-zhuravlev:
Should include all, otherwise hidden 'job' label can 'stuck' to first job value found in the list.


gaantunes (approved):
Good catch! LGTM.

PR #1044, opened 2023-08-14
Add MongoDB Atlas mixin

dpaasman00:
Adds 5 dashboards and 5 alerts for MongoDB Atlas

Dashboards
- MongoDB Atlas cluster overview
- MongoDB Atlas operations overview
- MongoDB Atlas performance overview
- MongoDB Atlas elections overview
- MongoDB Atlas sharding overview 

Alets
- MongoDBAtlasHighNumberOfDeadlocks
- MongoDBAtlasHighNumberOfSlowNetworkRequests
- MongoDBAtlasDiskSpaceLow
- MongoDBAtlasSlowHardwareIO
- MongoDBAtlasHighNumberOfTimeoutElections

**The sharding dashboard is specific to 2 kinds of nodes, and each node has different panels that relate to it, so the two different views for the dashboard are included below**

## Atlas cluster overview
cluster1
cluster2

## Atlas operations overview
operations1
operations2
operations3

## Atlas performance overview
performance1
performance2

## Atlas elections overview
elections1
elections2

## Atlas sharding overview (shardsvr node)
sharding1(shardsvr)
sharding2(shardsvr)

## Atlas sharding overview (mongos node)
sharding1(mongos)
sharding2(mongos)


atlas-mongodb-mixin/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+- Atlas cluster overview
+- Atlas performance overview
+- Atlas operations overview
+- Atlas elections overview
+- Atlas sharding overview
+
+and the following alerts:
+
+- AtlasHighNumberOfDeadlocks
```
schmikei, line 14: I think these alerts Should be `MongoDBAtlas` prefixed (linter considered)

Atlas just seems like it could be another technology not related to MongoDB.

atlas-mongodb-mixin/README.md
```diff
@@ -0,0 +1,103 @@
+# MongoDB Atlas Mixin
+The MongoDB Atlas mixin is a set of configurable Grafana dashboards and alerts.
+
+The MongoDB Atlas  mixin contains the following dashboards:
+
+- Atlas cluster overview
```
schmikei, line 6: Prefix these with MongoDB

atlas-mongodb-mixin/README.md
```diff
@@ -0,0 +1,103 @@
+# MongoDB Atlas Mixin
+The MongoDB Atlas mixin is a set of configurable Grafana dashboards and alerts.
+
+The MongoDB Atlas  mixin contains the following dashboards:
```
schmikei, line 4:
```suggestion
The MongoDB Atlas mixin contains the following dashboards:
```

atlas-mongodb-mixin/README.md
```diff
@@ -0,0 +1,103 @@
...
+- The Atlas sharding overview dashboard is controlled by a boolean flag in the config file. See more in the [environment configuration](#environment-configuration) section.
+- Within the Atlas sharding overview dashboard certain panels only relate to `shardsvr` nodes, other panels relate only to `mongos` nodes, and other relate to both nodes.
+
+## Atlas cluster overview
+The Atlas cluster overview dashboard provides an at a glance view into an Atlas cluster. Metrics such as hardware I/O's, network requests, memory usage, hardware disk usage, connections, latency, operations, deadlocks, and inventory tables for the nodes in the cluster. Please note that lock related panels such as deadlocks and wait times will show "no data" until load is generated for their respective metrics.
+
+![First screenshot of the MongoDB Atlas cluster overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/atlas-mongodb/screenshots/atlas-cluster-overview-1.png)
+![Second screenshot of the MongoDB Atlas cluster overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/atlas-mongodb/screenshots/atlas-cluster-overview-2.png)
+
+## Atlas performance overview
```
schmikei, line 32: Similar sentiments to brand guidelines I think should apply to dashboard names. It seems this was also a diversion from the spec?

mongodb-atlas-mixin/config.libsonnet
```diff
@@ -0,0 +1,20 @@
+{
+  _config+:: {
+    // sharding dashboard flag
+    enableShardingOverview: false,
```
schmikei, line 4: Just making sure that this is specced to be false by default?

mongodb-atlas-mixin/dashboards/mongodb-atlas-sharding-overview.libsonnet
```diff
@@ -0,0 +1,1330 @@
...
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+{
+  grafanaDashboards+::
+    if $._config.enableShardingOverview then {
```
schmikei, line 1234: This is clever, I like it 👍

schmikei (approved):
Think it looks good enough to proceed, albeit with the caveat that we should get better screenshots before we finalizing the integration 🚀

mongodb-atlas-mixin/config.libsonnet
dpaasman00: not really specced anywhere, but set it to false by default because as far as i can tell sharded clusters are not the default for atlas, you have to use their advanced config page to make a cluster sharded

mongodb-atlas-mixin/dashboards/mongodb-atlas-sharding-overview.libsonnet
```diff
@@ -0,0 +1,1374 @@
...
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+
+{
+  grafanaDashboards+:: {
+    'mongodb-atlas-sharding-overview.json':
```
kuiperda, line 1287: Commit [56ff7ba](https://github.com/grafana/jsonnet-libs/pull/1044/commits/56ff7ba36996a51016f8e547b0038b0750b9198f) removed the if check for this dashboard, need to put it in manually again. Also, this is missing a dashboard link.

mongodb-atlas-mixin/README.md
```diff
@@ -0,0 +1,103 @@
...
+  },
+}
+```
+
+## Alerts overview
+
+- MongoDBAtlasHighNumberOfDeadlocks: There is a high number of deadlocks occurring.
+- MongoDBAtlasHighNumberOfSlowNetworkRequests: There is a high number of slow network requests.
+- MongoDBAtlasDiskSpaceLow: Hardware is running out of disk space.
+- MongoDBAtlasSlowHardwareIO: Read and write I/O's are taking too long to complete.
```
kuiperda, line 72:
```suggestion
- MongoDBAtlasSlowHardwareIO: Read and write I/Os are taking too long to complete.
```

mongodb-atlas-mixin/README.md
```diff
@@ -0,0 +1,103 @@
...
+![First screenshot of the MongoDB Atlas sharding overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/mongodb-atlas/screenshots/mongodb-atlas-sharding-overview-1.png)
+![Second screenshot of the MongoDB Atlas sharding overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/mongodb-atlas/screenshots/mongodb-atlas-sharding-overview-2.png)
+
+## Environment configuration
+This mixin includes the MongoDB Atlas sharding overview dashboard, however the metrics used report data only when there is a sharded cluster in the project the prometheus integration is reporting on. Because of this, the MongoDB Atlas sharding overview dashboard is disabled by default. However, if you do have a sharded cluster and want to use the MongoDB Atlas sharding overview dashboard, you can set the `enableShardingDashboard` parameter in `config.libsonnet` to `true`.
+
+```
+{
+  _config+:: {
+    enableShardingDashboard: true,
```
kuiperda, line 62:
```suggestion
    enableShardingOverview: true,
```
to match config.libsonnet

mongodb-atlas-mixin/README.md
```diff
@@ -0,0 +1,103 @@
...
+
+![First screenshot of the MongoDB Atlas sharding overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/mongodb-atlas/screenshots/mongodb-atlas-sharding-overview-1.png)
+![Second screenshot of the MongoDB Atlas sharding overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/mongodb-atlas/screenshots/mongodb-atlas-sharding-overview-2.png)
+
+## Environment configuration
+This mixin includes the MongoDB Atlas sharding overview dashboard, however the metrics used report data only when there is a sharded cluster in the project the prometheus integration is reporting on. Because of this, the MongoDB Atlas sharding overview dashboard is disabled by default. However, if you do have a sharded cluster and want to use the MongoDB Atlas sharding overview dashboard, you can set the `enableShardingDashboard` parameter in `config.libsonnet` to `true`.
```
kuiperda, line 57:
```suggestion
This mixin includes the MongoDB Atlas sharding overview dashboard, however the metrics used report data only when there is a sharded cluster in the project the prometheus integration is reporting on. Because of this, the MongoDB Atlas sharding overview dashboard is disabled by default. However, if you do have a sharded cluster and want to use the MongoDB Atlas sharding overview dashboard, you can set the `enableShardingOverview` parameter in `config.libsonnet` to `true`.
```
to match config.libsonnet

kuiperda (commented):
Dashboards look great! A few small things, then run `make fmt` and `make lint` before pushing.

mongodb-atlas-mixin/dashboards/mongodb-atlas-sharding-overview.libsonnet
dpaasman00: oh yes thank you, good call, I had to reconvert the json to libsonnet bc dashboards_out was having an issue with the microseconds symbol

mongodb-atlas-mixin/README.md
dpaasman00: good catch

mongodb-atlas-mixin/README.md
dpaasman00: good catch

kuiperda (approved):
Nice!

gaantunes:
It looks very good overall, great work!
Only 2 considerations:

- At the cluster overview dashboard, I would place the tables before the time series panels, since they give more context to what I am looking at.

- If possible, I would like to see some sharding operation happening, so we know how it looks like on the dashboards. I know this is probably not under our control since it is SaaS, but if possible, would be great :)

dpaasman00:
> * If possible, I would like to see some sharding operation happening, so we know how it looks like on the dashboards. I know this is probably not under our control since it is SaaS, but if possible, would be great :)

There is some sharding operation load occurring for the mongos nodes which appear under the replica set "none". It isn't very clear that is where the mongos nodes "live", but that's just how MongoDB Atlas identifies their mongos nodes in their Prometheus integration: by assigning them the "none" replica set. 

Your call on this, but should a callout for this bit of info be added to the mixin? I thought about it originally but wasn't too sure.

gaantunes:
> > * If possible, I would like to see some sharding operation happening, so we know how it looks like on the dashboards. I know this is probably not under our control since it is SaaS, but if possible, would be great :)
> 
> There is some sharding operation load occurring for the mongos nodes which appear under the replica set "none". It isn't very clear that is where the mongos nodes "live", but that's just how MongoDB Atlas identifies their mongos nodes in their Prometheus integration: by assigning them the "none" replica set.
> 
> Your call on this, but should a callout for this bit of info be added to the mixin? I thought about it originally but wasn't too sure.

Got it, thanks for explaining! I guess one thing we could do is add this information on the panel info, just so it is clear.

gaantunes (approved):
LGTM. Thanks for the great work.

v-zhuravlev (changes requested):
Looks good. Some minor suggestions about first dashboard. I'll review others and post comments in next message, if any.

1. MongoDB Cluster Overview:
a. Hardware I/o panel - please add units?
b. Network bytes - most likely should be Bps not just Bytes?
c. Hardware I/O wait time - Just checking. is this correct that screenshot showing write wait 
time as 1 min?
d. please add units, most likely 'rps'?
e. Read/Write latency. 400s latency looks pretty high. Is this correct?
f. Read/Write opertaions - units missing
g. Operations - are those totals for time range opened? Can we rename 'Value' to Totals in legend?
h. Disk space - I would suggest to set min:0 and max:100 here, to see when we are near the limit. Also, consider to rename maybe to Disk space usage or Disk space free, dependending on what you have here.
i. Legends styles are mixed on the dashboard. Maybe position all to 'bottom', so we have less empty space on the screen. I would also drop min/max/mean in legend alltogher here, instead, some really imporant numbers can be presented as 'Stat' panels on top of the dashboard.

mongodb-atlas-mixin/config.libsonnet
```diff
@@ -0,0 +1,20 @@
...
+    dashboardRefresh: '1m',
+
+    // alerts thresholds
+    alertsDeadlocks: 10,  // count
+    alertsSlowNetworkRequests: 10,  // count
+    alertsHighDiskUsage: 90,  // percentage: 0-100
+    alertsSlowHardwareIO: 3,  // seconds
+    alertsHighTimeoutElections: 10,  // count
+
+    enableLokiLogs: false,
```
v-zhuravlev, line 18:
```suggestion
```
Worth dropping if we don't any logs?

dpaasman00:
> Looks good. Some minor suggestions about first dashboard. I'll review others and post comments in next message, if any.
> 
> 1. MongoDB Cluster Overview:
>    a. Hardware I/o panel - please add units?
>    b. Network bytes - most likely should be Bps not just Bytes? UPD. I see that you use increase(), but is there a reason to show per $interval rate for network usage not per second?
>    c. Hardware I/O wait time - Just checking. is this correct that screenshot showing write wait
>    time as 1 min?
>    d. please add units, most likely 'rps'?
>    e. Read/Write latency. 400s latency looks pretty high. Is this correct?
>    f. Read/Write operations - units missing
>    g. Operations - are those totals for time range opened? Can we rename 'Value' to Totals in legend?
>    h. Disk space - I would suggest to set min:0 and max:100 here, to see when we are near the limit. Also, consider to rename maybe to Disk space usage or Disk space free, depending on what you have here.
>    i. Legends styles are mixed on the dashboard. Maybe position all to 'bottom', so we have less empty space on the screen. I would also drop min/max/mean in legend altogether here, instead, some really important numbers can be presented as 'Stat' panels on top of the dashboard.
>    j. I would consider placing tables with cluster members on top, as they provide some good overview on what we have in cluster. Performance row can follow
>    l. It would be good to highlight member states in table with some coloring, for example like 'light-blue': Primary, 'light-yellow": Secondary.

Hey Vitaly, thanks for taking a look. To answer some of your feedback...
a. Right now I have the unit as just "number" since it is more of a count, but if you'd prefer custom units like "inputs" and "outputs" I can do so. I had thought "number" as the unit would have been fine.
b. Yea I can just change these queries to use rates instead and change the unit as a result
c. yea, that metric is reported in ms and i have the unit set as ms, not entirely sure why the performance is so poor but it is correct.
d. not entirely sure which panel you're referring to here
e. yea, similar story to "c", this metric is in microseconds and that's what i have the unit set as
f. same comment here as for "a"
g. ya no problem changing to "Value"
h. no problem, i'll add the min,max and rename to "Disk space usage"
i. no problem dropping the min/max/mean stuff from the legends. as for positioning the legends on the bottom, with the new guidelines on "stacking" metrics when applicable, i thought that also applied to setting the legends on the right side of the panel. if that isn't the case then i have no problem dropping them to the bottom of the panel, I'm more just unsure about when legends should be on the right side vs on the bottom.
j. yup, the screenshots in the pr don't reflect it but that was part of Gabriel's feedback and the pr is up to date with this change
l. no problem, this is something i can look into

mongodb-atlas-mixin/config.libsonnet
dpaasman00: can do, wasn't sure if that needed to be there regardless

dpaasman00:
> 2. For MongoDB Atlas Operations
> 
> a. Dashboard looks long, is it worth placing R/w operations, r/w latency on same row? Same for locks types, can be on same row. 
> b. R/w latency - 500s. Looks like a lot. Are proper units used? 
> c. if you use increase() instead of rate(), you can also provide 'Total' for total number of operations happened on legend or as separate 'Stat' panel. I would go for second option and drop extended legend. 
> d. if you use increase, you basically showing numbes per $interval, where $interval could be 1m or more. If you open dashboard on very large interval (90d?), numbers can rise dramatically due different interval would be in use. To avoid confusion, try putting $__interval into panel titles, like 'Read and write operations / $interval' or switch to rates() where applicable. you can check our traefik integration where increase() is also use instead of rate().

a. definitely an option, my concern with doing so is i'd need to realign the legends to the bottom of the panels and in doing so if/when a user decides to view these dashboards from a replica set point of view the legend would be helpful to have aligned on the right side. i could drop the min/mean/max stats from the legends and that may allow for the r/w latency and ops panels to be placed on the same row. However I don't think the lock metrics would work for condensing further
b. yep
c. ya, i can remove the min/mean/max stats and replace with just total
d. I hadn't realized this was a potential problem? How can the dashboard be opened on a large interval? I have the intervals set to `1m` in the queries rather than the default `auto`. Would that make a difference?

v-zhuravlev:
> a. Right now I have the unit as just "number" since it is more of a count, but if you'd prefer custom units like "inputs" and "outputs" I can do so. I had thought "number" as the unit would have been fine.
You may try  "unit": "iops" for this panel.

v-zhuravlev:
> d. I hadn't realized this was a potential problem? How can the dashboard be opened on a large interval? I have the intervals set to `1m` in the queries rather than the default `auto`. Would that make a difference?

This '1m' interval is not really fixed. It is minimum interval, therefore, if you change time range to 6 months - it would change to something bigger.

v-zhuravlev:
> i. no problem dropping the min/max/mean stuff from the legends. as for positioning the legends on the bottom, with the new guidelines on "stacking" metrics when applicable, i thought that also applied to setting the legends on the right side of the panel. if that isn't the case then i have no problem dropping them to the bottom of the panel, I'm more just unsure about when legends should be on the right side vs on the bottom.

I would say go for 'bottom' where possible. Still, this rule is not strict. My concern was that legend styles are mixed, I think they look better when consistent.

v-zhuravlev:
> g. ya no problem changing to "Value" but not sure what you mean by "time range opened"

Time interval of the dashboard: Last 1h, Last 7d...

dpaasman00:
> > g. ya no problem changing to "Value" but not sure what you mean by "time range opened"
> 
> Time interval of the dashboard: Last 1h, Last 7d...

testing it and changing the time interval of the dashboard has no effect on the numbers, so not sure if this still needs to become "Total" or if staying as "Value" is okay

mongodb-atlas-mixin/dashboards/mongodb-atlas-sharding-overview.libsonnet
```diff
@@ -960,44 +941,39 @@ local manyShardsPanel = {
...
-      'increase(mongodb_shardingStatistics_numHostsTargeted_update_manyShards{job=~"$job",cl_name=~"$cl_name",rs_nm=~"$rs",instance=~"$instance"}[$__interval:])',
+      'rate(mongodb_shardingStatistics_numHostsTargeted_update_manyShards{job=~"$job",cl_name=~"$cl_name",rs_nm=~"$rs",instance=~"$instance"}[$__rate_interval:])',
       datasource=promDatasource,
       legendFormat='{{instance}} - update',
       format='time_series',
-      interval='1m',
     ),
     prometheus.target(
-      'increase(mongodb_shardingStatistics_numHostsTargeted_delete_manyShards{job=~"$job",cl_name=~"$cl_name",rs_nm=~"$rs",instance=~"$instance"}[$__interval:])',
+      'rate(mongodb_shardingStatistics_numHostsTargeted_delete_manyShards{job=~"$job",cl_name=~"$cl_name",rs_nm=~"$rs",instance=~"$instance"}[$__rate_interval:])',
```
v-zhuravlev, line 962: I think you don't need subquery in case of rate instead of increase():
`$__rate_interval:` -> `$__rate_interval`

v-zhuravlev (approved):
LGTM, one last small nit, in comment above.

v-zhuravlev:
LGTM

PR #1045, opened 2023-08-18
Refactor opensearch to use as submixin (modular observ lib)

v-zhuravlev:
To use as submodule, helping to avoid overlaps in alert rules or dashboard results.
1) Adds config params:
- filteringSelector
- groupLabels
- instanceLabels
- uid (used in dashboard uids, alert group uids)
- dashboardNamePrefix

2) Use commonlib for CPU/memory/disk/network panel styles
3) Change aggregations from sum to avg on cluster level (to avoid miscalculations where cluster-wide data is gathered multiple times (from different nodes))

In other words, this mixin was refactored, bringing it closer to modular-observ-lib format, and is ready to be instantiated twice in single Grafana instance.

### DIFF
--- a/opensearch-mixin/.lint
+++ b/opensearch-mixin/.lint
@@ -3,6 +3,7 @@ exclusions:
     reason: Uses --mixed-- (generated from grafonnet)
     entries:
       - panel: 'OpenSearch cluster overview'
+      - panel: 'Roles'
   panel-units-rule:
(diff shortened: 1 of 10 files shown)


rgeyer (approved):

rgeyer:
Looks like there are still a couple small jsonnetfmt issues, but looks good otherwise.

PR #1046, opened 2023-08-21
Add Aerospike Mixin

Caleb-Hurshman:
**Note for reviewers**: for the logs dashboard on the keithschmitty cloud instance, it will default to an incorrect loki datasource. Switch to `grafanacloud-keithschmitty-logs` to see log data.

Adds dashboards and alerts for Aerospike.

Dashboards:
- Aerospike overview
- Aerospike instance overview
- Aerospike namespace overview
- Aerospike logs

Alerts:
- AerospikeNodeHighMemoryUsage
- AerospikeNamespaceHighDiskUsage
- AerospikeUnavailablePartitions
- AerospikeDeadPartitions
- AerospikeNamespaceRejectingWrites
- AerospikeHighClientReadErrorRate
- AerospikeHighClientWriteErrorRate
- AerospikeHighClientUDFErrorRate

Aerospike overview
aerospike_overview_1
aerospike_overview_2

Aerospike instance overview
aerospike_instance_overview_1
aerospike_instance_overview_2
aerospike_instance_overview_3

Aerospike namespace overview
aerospike_namespace_overview_1
aerospike_namespace_overview_2

Aerospike logs
aerospike_logs_1


aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+            includeAll=true,
+            multi=true,
+            allValues='',
+            sort=0
+          ),
+          template.new(
+            'k',
+            promDatasource,
+            '5, 10, 20, 50',
+            label='K',
```
dpaasman00, line 1152: Just curious if having a selector for K is something we've done before/want to do here?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
+local g = (import 'grafana-builder/grafana.libsonnet');
```
dpaasman00, line 1: not 100% sure where/how, but i think the panels with tables will need another filter/transfrom/override on them so that the value shown in them is something like "lastNonNull" or something else...currently they seem to show infinite entries for the same instance/namespaces

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+      },
+      decimals: 0,
+      mappings: [],
+      thresholds: {
+        mode: 'absolute',
+        steps: [
+          {
+            color: 'green',
+            value: null,
+          },
```
dpaasman00, line 329: I think you could add in a threshold here for 1 similar to other unavailable partition panels throughout the mixin, unless i'm missing something and it doesn't make sense here

dpaasman00 (commented):
Pretty solid, documentation looks good, just some questions about some of the dashboards

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: No and yes. This is from Gabriel's feedback, but not something we've ever done befora afaik.

aerospike-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,154 @@
...
+              description:
+                (
+                  'Namespace {{$labels.ns}} on node {{$labels.instance}} on cluster {{$labels.aerospike_cluster}} is currently rejecting all client-originated writes.'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'AerospikeHighClientReadErrorRate',
+            expr: |||
+              sum without(service) (rate(aerospike_namespace_client_read_error[5m])) / (clamp_min(sum without(service) (rate(aerospike_namespace_client_read_error[5m])) + sum without(service) (rate(aerospike_namespace_client_read_success[5m])), 1)) > %(alertsWarningHighClientReadErrorRate)s
```
StefanKurek, line 99: Nothing wrong with this, but this also should have worked just fine if you took out all of the "rates" (although it would mean something slightly different than with the rates)

aerospike-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,154 @@
...
+                (
+                  '{{ printf "%%.0f" $value }} percent of client read transactions are resulting in errors for namespace {{$labels.ns}}, on node {{$labels.instance}}, on cluster {{$labels.aerospike_cluster}}, ' +
+                  'which is above the threshold of %(alertsWarningHighClientReadErrorRate)s.'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'AerospikeHighClientWriteErrorRate',
+            expr: |||
+              sum without(service) (rate(aerospike_namespace_client_write_error[5m])) / (clamp_min(sum without(service) (rate(aerospike_namespace_client_write_error[5m])) + sum without(service) (rate(aerospike_namespace_client_write_success[5m])), 1)) > %(alertsWarningHighClientWriteErrorRate)s
```
StefanKurek, line 117: same comment about the rates here.

aerospike-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,154 @@
...
+                (
+                  '{{ printf "%%.0f" $value }} percent of client write transactions are resulting in errors for namespace {{$labels.ns}}, on node {{$labels.instance}}, on cluster {{$labels.aerospike_cluster}}, ' +
+                  'which is above the threshold of %(alertsWarningHighClientWriteErrorRate)s.'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'AerospikeHighClientUDFErrorRate',
+            expr: |||
+              sum without(service) (rate(aerospike_namespace_client_udf_error[5m])) / (clamp_min(sum without(service) (rate(aerospike_namespace_client_udf_error[5m])) + sum without(service) (rate(aerospike_namespace_client_udf_complete[5m])), 1)) > %(alertsWarningHighClientUDFErrorRate)s
```
StefanKurek, line 135: same general comment

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek, line 1152: @Caleb-Hurshman Just commenting that it seems like we should be able to have a better label name than "K"? like "Top percentile" or whatever? What do you think?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+
+local promDatasource = {
+  uid: '${%s}' % promDatasourceName,
+};
+
+local nodesPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(aerospike_cluster, instance) (aerospike_namespace_ns_cluster_size{job=~"$job", aerospike_cluster=~"$aerospike_cluster"})',
```
StefanKurek, line 19: this doesn't seem to take into account whether the `cluster` selector should be used. I'm guessing this applies to all panels and all dashboards. take a look at https://github.com/grafana/jsonnet-libs/blob/62aec8403a5c38d5dc97ba596703753289b1c33b/apache-cassandra-mixin/dashboards/cassandra-overview.libsonnet#L1781 for an example

StefanKurek, line 19: any reason why you aren't using "job" to be safe in your `by`s? I'm guessing this applies to all dashboards/panels?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+
+local dashboardUid = 'aerospike-overview';
+
+local promDatasourceName = 'prometheus_datasource';
+
+local promDatasource = {
+  uid: '${%s}' % promDatasourceName,
+};
+
+local nodesPanel = {
```
StefanKurek, line 15: i still have doubts that gabriel was asking for a table with only the most basic information for nodes (no metrics). i know you show'd him your spec (which updated the panels types in the image to "table"), but was it made clear to him that your plan was to only have the list of nodes with no actual metric data?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+            desc: true,
+            field: 'Time',
+          },
+        ],
+      },
+    },
+  ],
+};
+
+local namespacesPanel = {
```
StefanKurek, line 154: same general concern about a table with no metrics

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(job, aerospike_cluster) (aerospike_namespace_unavailable_partitions{job=~"$job", aerospike_cluster=~"$aerospike_cluster"})',
+      datasource=promDatasource,
+      legendFormat='{{aerospike_cluster}}',
+      format='time_series',
+    ),
+  ],
+  type: 'bargauge',
```
StefanKurek, line 313: seems like this would be better as a horizontal bar gauge in case there were multiples showing (text for each wouldn't be bumping into each other)

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+local deadPartitionsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(job, aerospike_cluster) (aerospike_namespace_dead_partitions{job=~"$job", aerospike_cluster=~"$aerospike_cluster"})',
+      datasource=promDatasource,
+      legendFormat='{{aerospike_cluster}}',
+    ),
+  ],
+  type: 'bargauge',
```
StefanKurek, line 380: same comment about horizontal

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek: yeah, if it's really bad, then we should have anything thats > 0 show up as red maybe? for both partition panels

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+      },
+    },
+  ],
+};
+
+local topNodesByMemoryUsagePanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'topk($k, 100 - sum by(aerospike_cluster, instance) (avg_over_time(aerospike_node_stats_system_free_mem_pct{job=~"$job", aerospike_cluster=~"$aerospike_cluster"}[$__interval])))',
```
StefanKurek, line 441: i'm probably a little behind on this. Is this supposed to be showing the top % of "worst nodes"? or all nodes that have a metric value > some %?

either way, can you help explain how to me how this calculation does either of those things, because my brain isn't processing it right now?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+            color: 'green',
+            value: null,
+          },
+          {
+            color: 'red',
+            value: 80,
+          },
+        ],
+      },
+      unit: 'ops',
```
StefanKurek, line 788: there are grafana units for "reads per second". not sure if you think that would be more appropriate

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+        insertNulls: false,
+        lineInterpolation: 'linear',
+        lineWidth: 1,
+        pointSize: 5,
+        scaleDistribution: {
+          type: 'linear',
+        },
+        showPoints: 'never',
+        spanNulls: false,
+        stacking: {
```
StefanKurek, line 766: so is this a stacked timeseries. if so, are we sure that there is no overlap between any of these read metrics? are they definitely all unique and added together make up the total reads?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local clientWritesPanel = {
```
StefanKurek, line 806: same questions about this panel as for read version of it

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+        insertNulls: false,
+        lineInterpolation: 'linear',
+        lineWidth: 1,
+        pointSize: 5,
+        scaleDistribution: {
+          type: 'linear',
+        },
+        showPoints: 'never',
+        spanNulls: false,
+        stacking: {
```
StefanKurek, line 956: same question about stacked timeseries here

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1176 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local connectionsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(aerospike_cluster, job) (aerospike_node_stats_client_connections{aerospike_cluster=~"$aerospike_cluster", job=~"$job"})',
```
StefanKurek, line 1000: so are these all gauge metrics and not counters? i notice you're not doing an `increase` or `rate` here

StefanKurek (changes requested):

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+local unavailablePartitionsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(instance, aerospike_cluster, job) (aerospike_namespace_unavailable_partitions{job=~"$job", aerospike_cluster=~"$aerospike_cluster", instance=~"$instance"})',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+    ),
+  ],
+  type: 'bargauge',
```
StefanKurek, line 29: should it be horizontal?

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+      fields: '',
+      values: false,
+    },
+    showUnfilled: true,
+    text: {
+      valueSize: 100,
+    },
+    valueMode: 'color',
+  },
+  pluginVersion: '10.2.0-59542pre',
```
StefanKurek, line 73: is this needed?

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+local deadPartitionsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(instance, aerospike_cluster, job) (aerospike_namespace_dead_partitions{job=~"$job", aerospike_cluster=~"$aerospike_cluster", instance=~"$instance"})',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+    ),
+  ],
+  type: 'bargauge',
```
StefanKurek, line 165: horizontal?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: We're actually not using a percentile anymore. I think renaming it would be good as 'k' isn't super clear.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: I actually showed Gabriel the exact implementation and got his approval on it.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: It takes the free pct for each node, subtracts 100 to get the usage, then takes the top-k usages.

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local connectionsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'aerospike_node_stats_client_connections{aerospike_cluster=~"$aerospike_cluster", job=~"$job", instance=~"$instance"}',
```
StefanKurek, line 376: just double checking that these are gauges

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local cacheReadUtilizationPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(job, aerospike_cluster, instance) (aerospike_namespace_cache_read_pct{job=~"$job", aerospike_cluster=~"$aerospike_cluster", instance=~"$instance"})',
```
StefanKurek, line 753: what labels are you summing away here? is sum actually appropriate for a cache hit percentage? would average be better? just double checking

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,964 @@
...
+  },
+};
+
+local systemLogsPanel = {
+  datasource: lokiDatasource,
+  targets: [
+    {
+      datasource: lokiDatasource,
+      editorMode: 'code',
+      expr: '{job=~"$job", aerospike_cluster=~"$aerospike_cluster", instance=~"$instance"}',
```
StefanKurek, line 834: don't we usually have a "filename" in this expression as well?

StefanKurek (changes requested):

aerospike-mixin/dashboards/aerospike-namespace-overview.libsonnet
```diff
@@ -0,0 +1,754 @@
...
+
+local dashboardUid = 'aerospike-namespace-overview';
+
+local promDatasourceName = 'prometheus_datasource';
+
+local promDatasource = {
+  uid: '${%s}' % promDatasourceName,
+};
+
+local unavailablePartitionsPanel = {
```
StefanKurek, line 15: these panels all look to be the same as other dashboards, so any comments applies to panels from other dashboards would apply here

StefanKurek (changes requested):

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: Yep, [AS metric reference](https://docs.aerospike.com/reference/metrics) calls them out as instantaneous (aka gauges)

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
Caleb-Hurshman: yep
[AS metric reference](https://docs.aerospike.com/reference/metrics)

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
Caleb-Hurshman: I got feedback on Couchbase to do the opposite.

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
Caleb-Hurshman: Avg is probably better here.

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
Caleb-Hurshman: This appears in 32 files across `jsonnet-libs`, probably added during dashboard conversion.

Caleb-Hurshman: This seems to appear pretty regularly across `jsonnet-libs`

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
Caleb-Hurshman: This is set to auto, so I think it should adjust between horizontal and vertical depending on how much space is available? I personally prefer vertical for lower values, but maybe this doesn't work the way I think it does.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: Just to clarify, Gabriel directed me to use top-k here instead of a percentile like he suggested earlier.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek: oooh, now i think i see the value. it looks like you have links to the node dashboard? do they automatically select the correct instance selector as well?

Caleb-Hurshman: Not currently, it wasn't clear if he wanted that, so I kept it like this for now while I work on the agent integration.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek: @Caleb-Hurshman I still think "Top K" might be unclear for some customers. How about something more along the lines of "Top node count" or something like that?

StefanKurek: To be clear I'm referring to the dashboard where this selector still exists

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek: Feel free to resolve

aerospike-mixin/dashboards/aerospike-overview.libsonnet
StefanKurek: Ok, feel free to resolve....but given the way the tables are, this addition would be worthwhile

StefanKurek (changes requested):
2nd pass not fully done, but figured i'd get these comments out there

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
StefanKurek: @Caleb-Hurshman just if there were multiple things showing (not just one like in your current environment), horizontal would be better to show the label for each bar

aerospike-mixin/alerts/alerts.libsonnet
Caleb-Hurshman: I don't think this alert is nearly as valuable without the rates

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: Will change, but not sure what you mean by your clarifying comment as this selector has only ever existed on one dashboard.

dpaasman00 (commented):
2nd pass looking good, as of now just a nit and a couple questions

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1209 @@
...
+            refresh=2,
+            includeAll=true,
+            multi=true,
+            allValues='',
+            sort=0
+          ),
+          template.new(
+            'k',
+            promDatasource,
+            '5,10,20,50',
```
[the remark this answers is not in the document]
Caleb-Hurshman: These last two questions will be updated on the cloud instance shortly.

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1209 @@
+local g = (import 'grafana-builder/grafana.libsonnet');
```
dpaasman00, line 1: Pretty nit and just a stylistic choice, but I feel like the tables on this dashboard should have their columns in order of descending order/cardinality ... so like the columns should go "Job", "Aerospike cluster", "Instance", (then "Namespace" for the 1 panel with namespaces), but again a stylistic choice so your call on this

dpaasman00: But I did notice while making the above comment that the column names aren't sentence case and afaik I think they need/should be sentence case. So the Job, Instance, Namespace, and Usage columns may need to changed... pretty sure that can be done with an override or transformation.

dpaasman00 (approved):
LGTM, just a couple nits/style choices that are ultimately your call

aerospike-mixin/dashboards/aerospike-overview.libsonnet
Caleb-Hurshman: I've already showed this to Gabriel and he has yet to comment on any of these things. I'll resolve this convo for now and wait for external review here.

aerospike-mixin/dashboards/aerospike-instance-overview.libsonnet
```diff
@@ -0,0 +1,987 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local heapMemoryEfficiencyPanel(matcher) = {
```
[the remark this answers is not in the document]
Caleb-Hurshman: I think I like having percent metrics using a max in general, and I don't believe this one can go over 100. More complex environments will see values lower than mine.

aerospike-mixin/jsonnetfile.json
```diff
@@ -0,0 +1,15 @@
+{
+  "version": 1,
+  "dependencies": [
+      {
+          "source": {
+              "git": {
+                  "remote": "https://github.com/grafana/grafonnet-lib.git",
```
BominRahmani, line 7: I believe we are not using this library anymore. Its deprecated and they asked us to use grafonnet 
image
Dan repinged it on the grafana documentations channel.

StefanKurek (approved):

aerospike-mixin/jsonnetfile.json
Caleb-Hurshman: Good call out. Going to continue working on this. Since it doesn't seem to be a blocker for MongoDB Atlas so far, I'll open this up externally to not block too long.

aerospike-mixin/config.libsonnet
```diff
@@ -0,0 +1,24 @@
+{
+  _config+:: {
+    enableMultiCluster: true,
+    aerospikeSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
+    multiclusterSelector: 'job=~"$job"',
+
+    dashboardTags: ['aerospike-mixin'],
+    dashboardPeriod: 'now-1h',
```
gaantunes, line 8: Any specific reason for 1h?

aerospike-mixin/dashboards/aerospike-overview.libsonnet
```diff
@@ -0,0 +1,1209 @@
...
+
+local promDatasource = {
+  uid: '${%s}' % promDatasourceName,
+};
+
+local nodesPanel(matcher) = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by(job, aerospike_cluster, instance) (aerospike_namespace_ns_cluster_size{' + matcher + '})',
```
gaantunes, line 19: I would get rid of the job column from all the tables in this dashboard. This most of the times will have a single value, so exlcuing it will open up space for more relevant information.

aerospike-mixin/config.libsonnet
Caleb-Hurshman: I've always been told 1hr is standard.

gaantunes: We normally use 30m. If you do not set this manually, the `util.prepare-dashboards` function set it to the default `now-30m`. This avoids uneeded long running queries.

gaantunes (approved):
LGTM, just needs to update the time frame to `now-30m` as we discussed.

aerospike-mixin/config.libsonnet
Caleb-Hurshman: Good to know, I'll make sure to radiate this to the team.

StefanKurek (changes requested):

StefanKurek (approved):

aerospike-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,154 @@
...
+                  'which is above the threshold of %(alertsCriticalDeadPartitions)s.'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'AerospikeNamespaceRejectingWrites',
+            expr: |||
+              sum without(service) (aerospike_namespace_stop_writes + aerospike_namespace_clock_skew_stop_writes) > %(alertsCriticalSystemRejectingWrites)s
+            ||| % $._config,
+            'for': '1m',
```
v-zhuravlev, line 84:
```suggestion
            'for': '5m',
```
Consider to increase to 5m. Basically, In 1 minute, in Grafana Cloud only single point would be collected. But we need to get some confirmation that problem is there before bothering people with paging.

v-zhuravlev (changes requested):
Looks very good, thank you! 

1) Better to keep pending duration of alerts for at least 5m by default
2) If logs return severity level, best if parse it out to `level` label by using regex,logfmt, or json parsers, depending on format you have for you logs. 
  - Later on, this label can be used to search logs by severity(`level`), for instance shwoing only errors at specific case.
3) In our latest integraitons/mixins we tend to use [logs-lib](https://github.com/grafana/jsonnet-libs/tree/master/logs-lib) to generate dedicated logs dashboard, which then can be attached to the mixin.

You can check some examples in README: 
https://github.com/grafana/jsonnet-libs/blob/master/logs-lib/README.md
or [here](https://github.com/grafana/jsonnet-libs/pull/1006/files#diff-15814e586790cf38e37a6807094998a29cdf8d15c8c648119ce3a068dea7f2f5) 
or [here](https://github.com/grafana/cloud-onboarding/pull/4347/files#diff-ed4082fa30d5e39ae59297a1b0d09430364c7368a16d34e8e38e0f9605d4749a)
or [here](https://github.com/grafana/cloud-onboarding/pull/4329/files#diff-b180e79282c8609c77965107c896ee37d481bcc17fb2dc1c02974cddb1d7b430R8-R26)

Together with `level` from 2) you can generate such dashboard with `job,cluster,aerospike_cluster,instance,level` selectors.

aerospike-mixin/config.libsonnet
```diff
@@ -0,0 +1,25 @@
...
+    alertsCriticalNodeHighMemoryUsage: 80,  // %
+    alertsCriticalNamespaceHighDiskUsage: 80,  // %
+    alertsCriticalUnavailablePartitions: 0,  // count
+    alertsCriticalDeadPartitions: 0,  // count
+    alertsCriticalSystemRejectingWrites: 0,  // count
+    alertsWarningHighClientReadErrorRate: 25,  // %
+    alertsWarningHighClientWriteErrorRate: 25,  // %
+    alertsWarningHighClientUDFErrorRate: 25,  // %
+
+    enableLokiLogs: false,
```
v-zhuravlev, line 23: Can you use this parameter to hide/show logs dashboard?

aerospike-mixin/README.md
```diff
@@ -57,6 +57,13 @@ scrape_configs:
           instance: "<your-instance-name>"
           aerospike_cluster: "<your-cluster-name>"
           __path__: /var/log/aerospike/aerospike.log
+    pipeline_stages:
+        - multiline:
+            firstline: '\w{3} \d{2} \d{4}'
+        - regex:
+            expression: '\w{3} \d{2} \d{4} \d{2}:\d{2}:\d{2} \w{3}: (?P<level>\w+) (?P<context>\(\w+\)): (?P<trace>\(\S+\))\s+(?P<message>.*)'
```
v-zhuravlev, line 64: when you have multiline, 
(?P<message>.*)' should be 
(?P<message>(?s:.*))$'
https://grafana.com/docs/loki/latest/send-data/promtail/stages/multiline/

aerospike-mixin/README.md
```diff
@@ -57,6 +57,13 @@ scrape_configs:
...
+        - labels:
+            level:
 ```
 
 ## Aerospike Namespace Overview
```
v-zhuravlev, line 69: ## Aerospike namespace overview?

Caleb-Hurshman: Sorry, could you clarify the meaning of this comment? I've moved the log config to be under the logs dashboard and updated the regex, let me know if there's anything else.

v-zhuravlev: Sentence case should be used everywhere in Grafana where applicable. 
https://grafana.com/docs/writers-toolkit/write/style-guide/capitalization-punctuation/#sentence-case-capitalization

aerospike-mixin/dashboards/aerospike-logs-overview.libsonnet
```diff
@@ -0,0 +1,32 @@
...
+{
+  grafanaDashboards+::
+    if $._config.enableLokiLogs then {
+      local aerospikeLogs =
+        logsDashboard.new(
+          'Aerospike logs',
+          datasourceName='loki_datasource',
+          datasourceRegex='',
+          filterSelector=$._config.filterSelector,
+          labels=['job', 'aerospike_cluster', 'instance', 'level'],
```
v-zhuravlev, line 12:
```suggestion
          labels=['job', 'aerospike_cluster', 'instance', 'context', 'level'],
```
Looks like context is another one useful label we can extract.

v-zhuravlev (commented):
Looks like context is another one useful label we can extract.

aerospike-mixin/README.md
```diff
@@ -0,0 +1,138 @@
...
+          instance: "<your-instance-name>"
+          aerospike_cluster: "<your-cluster-name>"
+          __path__: /var/log/aerospike/aerospike.log
+    pipeline_stages:
+        - multiline:
+            firstline: '\w{3} \d{2} \d{4}'
+        - regex:
+            expression: '\w{3} \d{2} \d{4} \d{2}:\d{2}:\d{2} \w{3}: (?P<level>\w+) (?P<context>\(\w+\)): (?P<trace>\(\S+\))\s+(?P<message>(?s:.*))$'
+        - labels:
+            level:
```
v-zhuravlev, line 79:
```suggestion
scrape_configs:
  - job_name: integrations/aerospike
    static_configs:
      - targets: [localhost]
        labels:
          job: integrations/aerospike
          instance: "<your-instance-name>"
          aerospike_cluster: "<your-cluster-name>"
          __path__: /var/log/aerospike/aerospike.log
    pipeline_stages:
        - multiline:
            firstline: '\w{3} \d{2} \d{4}'
        - regex:
            expression: '\w{3} \d{2} \d{4} \d{2}:\d{2}:\d{2} \w{3}: (?P<level>\w+) (?P<context>\(\w+\)): (?P<trace>\(\S+\))\s+(?P<message>(?s:.*))$'
        - labels:
            level:
            context
```

v-zhuravlev: Looks like context is another one useful label we can extract.

v-zhuravlev (approved):
Thank you

PR #1053, opened 2023-09-08
Update `mixtool` to the latest version

julienduchesne:
We are severely out of date with `mixtool` (Current version is from January 2021). This PR updates it and adds exclusions for all non-compliant mixins (to be fixed later)

This was done in three steps (commits):
- I updated mixtool and fixed all non-Grafana dashboard errors. This was trivial since there weren't that many
- Added exclusions to all mixins for newly failing rules using a script that merges a base yaml file in each dir for the linting config
- Removed all unneeded linting exclusions using a script that removes each exclusion one by one and checks if linting is still passing

Asked for review some recent contributors, feel free to add more, or remove yourself


.lint.base.merge.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
```
iainlane, line 2: `set -e`

.lint.base.merge.sh
iainlane, line 1: Could you expose a Makefile target to run this script? Either one which depends on `lint-mixins` or do it in that target, not sure.

iainlane (approved):

PR #1055, opened 2023-09-11
Get rid of `std.native("parseYaml")`

julienduchesne:
This is in the jsonnet std lib now. 
I also updated the go-jsonnet versions in the makefile to match the version in mixtool


v-zhuravlev (approved):
Very useful!

Issue #1043, opened 2023-08-10
Kafka mixin should have configurable kafkaSelector for alerts


v-zhuravlev:
Kafka mixin has hardcoded job=integrations/kafka in all alerts

Instead, it should have configurable kafkaSelector that can be injected into alerts (and dashboards too).

v-zhuravlev:
For example:
```
      expr: sum(kafka_server_ReplicaManager_UnderReplicatedPartitions{job="integrations/kafka"}) by (instance) > 0
```

PR #1061, opened 2023-09-18
Kafka Mixin: Introducing kafka_cluster label to each dashboard, and filter selectors to dashboards and alerts

gaantunes:
This in introducing a kafka_cluster label to each dashboard.

Also introducing filter selectors to all dashboards and alerts, fixing https://github.com/grafana/jsonnet-libs/issues/1043.


Dasomeone (approved):
Looks good to me, though I did note that some of the plugins used were from version 7, so it might make sense also upgrade the dashboards a little. 
Doesn't have to be in this PR, but something we can add to the backlog :)

Dasomeone (approved):

gaantunes:
> Looks good to me, though I did note that some of the plugins used were from version 7, so it might make sense also upgrade the dashboards a little. Doesn't have to be in this PR, but something we can add to the backlog :)

Yes, this definitely deserves more attention, I think it is a good candidate for a common lib refactoring. Will create a ticket.

kafka-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,123 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'Kafka_Alerts',
+        rules: [
+          {
+            alert: 'KafkaOfflinePartitonCount',
+            expr: |||
+              sum without(kafka_cluster) (kafka_controller_KafkaController_OfflinePartitionsCount{%(FilterSelector)s}) > 0
```
v-zhuravlev, line 10: sum without (kafka_cluster) would drop `kafka_cluster` making it impossible to use in description below.
What do we want to here?

gaantunes: My bad, I misinterpreted the documentation. It should be correct now.

kafka-mixin/alerts/alerts.libsonnet
```diff
@@ -63,21 +59,21 @@
           {
             alert: 'KafkaISRExpandRate',
             expr: |||
-              sum by(instance, kafka_cluster) (rate(kafka_server_ReplicaManager_IsrExpandsPerSec{%(kafkaFilteringSelector)s}[5m])) != 0
+              sum (rate(kafka_server_replicamanager_isrexpandspersec{%(kafkaFilteringSelector)s}[5m])) != 0
```
v-zhuravlev, line 62: sum() will likely drop all labels, so instance/kafka_cluster would not be available in descriptions.

v-zhuravlev: - if you do sum(<metric>) you would get just sum with all labels dropped
- if you do sum by (<labe1l>) (<metric>) you would get just sum by (<label1>), thus only <label1> would present in result, others dropped.
- if you do sum without (<label1>) (<metric>) you would get sum by all labels, except <label1>
- you also just use expression like  `<metric> > x` if you don't need any aggregation at all.

"
These operators can either be used to aggregate over all label dimensions or preserve distinct dimensions by including a without or by clause. These clauses may be used before or after the expression.

`<aggr-op> [without|by (<label list>)] ([parameter,] <vector expression>)`

or

`<aggr-op>([parameter,] <vector expression>) [without|by (<label list>)]`

`without` removes the listed labels from the result vector, while all other labels are preserved in the output. `by` does the opposite and drops labels that are not listed in the by clause, even if their label values are identical between all elements of the vector.
"
https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

gaantunes: Best course of action here is to use sum without(), with empty without clause. This way sum() don't drop any label.

gaantunes: image

v-zhuravlev (approved):
LGTM

PR #1063, opened 2023-09-19
ActiveMQ alerts updates

BominRahmani:
A couple of small changes for alerts + usage of k_selector variable on queries it was missed in previously.


gaantunes (approved):
Good catches, thanks for Fixing it.

PR #1064, opened 2023-09-20
Apache ActiveMQ increase addition

BominRahmani:
- Addition of minimum interval on all increase queries.
- Changed line interpolation on single panel to smooth
- Added instance and cluster information onto tables.


apache-activemq-mixin/dashboards/apache-activemq-topic-overview.libsonnet
```diff
@@ -775,11 +775,15 @@ local topicSummaryPanel(matcher) = {
           'Time 4': true,
           '__name__ 1': true,
           '__name__ 2': true,
-          'activemq_cluster 1': true,
```
gaantunes, line 778: what is this change for?

BominRahmani: image
On the summary table, it will show the cluster and instance. I thought that this would be nice to have on the table, unless you disagree.

gaantunes (approved):

PR #1067, opened 2023-09-26
Update MongoDB Atlas mixin: fix allValue for selectors

dpaasman00:
quick fix to set `allValue = '.+'` for all variable selectors


gaantunes (approved):
LGTM.

PR #1072, opened 2023-09-27
Update logs-lib to pass mixtool linter

v-zhuravlev:
Add panel description
Add unit
Sentence case variables' labels
Change mixed DS to Loki
Name DS as 'Loki data source' by default
image

### DIFF
--- a/logs-lib/logs/panels.libsonnet
+++ b/logs-lib/logs/panels.libsonnet
@@ -9,6 +9,7 @@ local fieldConfig = timeSeries.fieldConfig;
 function(
   logsVolumeTarget,
   logsTarget,
+  logsVolumeGroupBy,
 )
 
   {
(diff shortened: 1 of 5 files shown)


v-zhuravlev:
This PR should fix logs lib warnings:
https://github.com/grafana/jsonnet-libs/pull/1072

logs-lib/logs/panels.libsonnet
```diff
@@ -17,12 +18,19 @@ function(
...
+      // set type to first target's type
+      + timeSeries.queryOptions.withDatasource(
+        logsVolumeTarget.datasource.type, logsVolumeTarget.datasource.uid
+      )
       + custom.withDrawStyle('bars')
       + custom.stacking.withMode('normal')
       + custom.withFillOpacity(50)
-      + timeSeries.queryOptions.withInterval('30s')  // must be set , otherwise interval is around 1ms
+      // should be set, otherwise interval is around 1s by default
+      + timeSeries.queryOptions.withInterval('30s')
```
rgeyer, line 30: Should this be a parameter somewhere? Or, perhaps in the common-lib we add a generic "panels utility" function that can update this on all panels you pass it.

Given how this "broke" so many dashboards for me, and how crucial it can be in getting the right display for a given query, it shouldn't be as hidden.

I do agree with setting a sane default value tho.

logs-lib/utils.libsonnet
```diff
@@ -10,6 +10,9 @@
 
   labelsToPanelLegend(labels): std.join('/', ['{{%s}}' % [label] for label in labels]),
 
+  toSentenceCase(string)::
```
rgeyer, line 13: Nice useful util!

rgeyer (approved):
Approval, with one observation/nit.

I think the solution isn't necessarily to change this lib, but to provide a top level utility function. LMK what you think.

logs-lib/utils.libsonnet
v-zhuravlev: I'll try to move those utils to grafonnet some time

logs-lib/logs/panels.libsonnet
v-zhuravlev: Indeed.  I'll try to address this hardcode next time I change something in this lib.

Dasomeone: I think an approach similar to Shahzeb's for the status panels, with how they ingest the integration version from the metadata might make sense? 

I wouldn't be against setting the default update interval in either the metadata or config, but we'd have to be consistent in using it

PR #1073, opened 2023-09-27
F5 BIG-IP:  add interval

JonathanWamsley:
This PR Adds a rate interval in. Running the k3d dashboards, the traffic panels did not have a set Interval / Min interval so the charts are not showing any data on load. This PR fixes this.

Before:
Screenshot 2023-09-27 at 1 54 54 PM

After new dashboard with changes:
Screenshot 2023-09-27 at 1 58 48 PM


v-zhuravlev:
I think It is best to fix this on k3d.

rgeyer (approved):

rgeyer:
> I think It is best to fix this on k3d.

It's probably fine to set a sane default. I did approve a bit impulsively, but perhaps we could meet in the middle and have this be a config parameter of the mixin, set to a sane value (30s or 1m) by default?

v-zhuravlev:
Best is if we set agent's scrape interval to 1m in k3d. And also set min interval in mimir datasource of k3d ops cluster to 1m as well. That should fix it

JonathanWamsley:
I just update to include a default rate interval  @rgeyer  . It seems like @v-zhuravlev has another solutions which I believe are outside of the mixin. I am guessing this would be changing default options on the k3d / mimir which I would not take apart of.  Let me know how to proceed, thanks.

rgeyer:
> Best is if we set agent's scrape interval to 1m in k3d. And also set min interval in mimir datasource of k3d ops cluster to 1m as well. That should fix it

@v-zhuravlev I agree that this should also have defaults set in the k3d environment. However, is there any harm in having this set as a default in the mixin, particularly since it's now configurable? I believe this should be fine, but I know you've been deeper into these specifics, so I wanted to make sure there isn't a concern I'm not considering.

v-zhuravlev:
I just think we should fix root cause, not symptoms.
Here, we only overcomplicating things, as it might be not obvious later on, why we set any interval here for two panels.

If you agree that new k3d defaults would fix this problem with this mixin as well, we can open an issue to track it, if we can't do it right now.

rgeyer:
Yep, that makes sense. Lets close this in favor of a fix in our dev environment. Sorry for any confusion @JonathanWamsley

PR #1076, opened 2023-09-29
Apache ActiveMQ Tweaks

BominRahmani:
- Added Drill down links onto queue and topics summary table.
image

- Simplified JVM queries and made it compatible with sample app
- Altered "Broker" -> "Brokers" on cluster dashboard
- Fixed linting issues caused by having empty allValues='' on template variables
- Added min and max on bar gauges on cluster dashboard to ensure bar gauges grow correctly.
- Got rid of sum without (job) on alerts because job labels are needed on alerts panel.

### DIFF
--- a/apache-activemq-mixin/alerts/alerts.libsonnet
+++ b/apache-activemq-mixin/alerts/alerts.libsonnet
@@ -7,7 +7,7 @@
           {
             alert: 'ApacheActiveMQHighTopicMemoryUsage',
             expr: |||
-              sum (activemq_topic_memory_percent_usage{destination!~"ActiveMQ.Advisory.*"}) > %(alertsHighTopicMemoryUsage)s
+              sum without (destination) (activemq_topic_memory_percent_usage{destination!~"ActiveMQ.Advisory.*"}) > %(alertsHighTopicMemoryUsage)s
             ||| % $._config,
             'for': '5m',
             labels: {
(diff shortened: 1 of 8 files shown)


apache-activemq-mixin/alerts/alerts.libsonnet
```diff
@@ -7,7 +7,7 @@
           {
             alert: 'ApacheActiveMQHighTopicMemoryUsage',
             expr: |||
-              sum without (job) (activemq_topic_memory_percent_usage{destination!~"ActiveMQ.Advisory.*"}) > %(alertsHighTopicMemoryUsage)s
+              sum (activemq_topic_memory_percent_usage{destination!~"ActiveMQ.Advisory.*"}) > %(alertsHighTopicMemoryUsage)s
```
v-zhuravlev, line 10: In this case you may likely want this:
```suggestion
              activemq_topic_memory_percent_usage{destination!~"ActiveMQ.Advisory.*"} > %(alertsHighTopicMemoryUsage)s
```

- if you do sum(<metric>) you would get just sum with all labels dropped
- if you do sum by (<labe1l>) (<metric>) you would get just sum by (<label1>), thus only <label1> would present in result, others dropped.
- if you do sum without (<label1>) (<metric>) you would get sum by all labels, except <label1>
- you also just use expression like  `<metric> > x` if you don't need any aggregation at all.

"
These operators can either be used to aggregate over all label dimensions or preserve distinct dimensions by including a without or by clause. These clauses may be used before or after the expression.

`<aggr-op> [without|by (<label list>)] ([parameter,] <vector expression>)`

or

`<aggr-op>([parameter,] <vector expression>) [without|by (<label list>)]`

`without` removes the listed labels from the result vector, while all other labels are preserved in the output. `by` does the opposite and drops labels that are not listed in the by clause, even if their label values are identical between all elements of the vector.
"
https://prometheus.io/docs/prometheus/latest/querying/operators/#aggregation-operators

v-zhuravlev: you can try to experiment in Grafana->`Explore` to see resulting vectors.

v-zhuravlev (approved):

PR #1078, opened 2023-10-04
Apache ActiveMQ Table + Stat panel changes

BominRahmani:
- aggregates all data into one number for stat panels
- fixes table having multiple extra labels present


v-zhuravlev (commented):
Instead of excluding, try to include fields we actually need in tables. That would more reliable.

v-zhuravlev (approved):

PR #1079, opened 2023-10-09
Fixing enableLokiLogs flags

gaantunes:
This is fixing a few mixins that were not considering the enableLokiLogs flag correctly.


v-zhuravlev (approved):

asterisk-mixin/config.libsonnet
```diff
@@ -0,0 +1,5 @@
+{
+  _config+:: {
+    enableLokiLogs: true,
+  },
+}
```
Dasomeone, line 5: The newline has escaped 😨

asterisk-mixin/mixin.libsonnet
```diff
@@ -1,7 +1,15 @@
+local config = (import 'config.libsonnet');
+
 {
-  grafanaDashboards: {
+  grafanaDashboards: 
+  if config._config.enableLokiLogs then
+  {
+    'asterisk-overview.json': (import 'dashboards/asterisk-overview.json'),
+    'asterisk-logs.json': (import 'dashboards/asterisk-logs.json')
```
Dasomeone, line 8: Minor nitpick/question here, as I don't recall seeing this pattern before.

Do we have a style of trending towards clearly distinct conditional blocks like here, where overview is imported in both conditions?

Asking because my gut feel is that the conditional should only affect the logs dashboard? Or maybe this is a quirk of jsonnet making this easier?

Dasomeone (approved):
Couple minor comments, but overall looks great :D
Thanks for catching these!

asterisk-mixin/mixin.libsonnet
gaantunes: It is more of a jsonnet syntax thing, I was not able to do it another way unfortunately. We normally define these at the dashboard level, but these are json dashboards so it does not accept jsonnet logic.

Dasomeone: Got it, thank you for explaining :D

PR #1080, opened 2023-10-11
Clickhouse add multicluster support for K8s

JonathanWamsley:
This PR adds the cluster selector to support k8s plugin


mshahzeb (approved):

clickhouse-mixin/config.libsonnet
```diff
@@ -1,9 +1,15 @@
 {
   _config+:: {
+    enableMultiCluster: false,
+    clickhouseSelector: if self.enableMultiCluster then 'job=~"$job", instance=~"$instance", cluster=~"$cluster"' else 'job=~"$job", instance=~"$instance"',
+
     dashboardTags: ['clickhouse-mixin'],
     dashboardPeriod: 'now-30m',
     dashboardTimezone: 'default',
     dashboardRefresh: '1m',
+    dashboardClickhouseErrorFilename: '/var/log/clickhouse-server/clickhouse-server.err.log',
```
v-zhuravlev, line 10: Do we need static filename in query? Chances are path or filename would by different on some clickhouse setups.

v-zhuravlev: IMHO, selector `job=~"$job", instance=~"$instance"` or `cluster=~"$cluster", job=~"$job", instance=~"$instance"` should be enough for logs.

JonathanWamsley: True. Now that there is an actual `ClickHouse logs` dashboard with great filtering, I wonder if this errors log panel should still be around then?

v-zhuravlev (approved):

PR #1082, opened 2023-10-12
Add parameter for adding a separate row for status panels in status panels lib

mshahzeb:
Adds a parameter that controls if a new row is to be added for the status panels or not.


v-zhuravlev (approved):
LGTM

PR #1085, opened 2023-10-16
Add Apache HBase mixin

Caleb-Hurshman:
Adds dashboards and alerts for the Apache HBase mixin.

### Note
The vm environment that is currently sending to Grafana cloud is highly unreliable due to Hadoop consuming large amounts of disk space. For a representative image of what these dashboards will look like under load, refer to the time the screenshots were taken.
See also an example of a dead RegionServer at 9:15-9:44am EST on 10/17.

Dashboards:
- Apache HBase cluster overview
- Apache HBase RegionServer overview
- Apache HBase logs overview

Alerts:
- ApacheHBaseHighHeapMemUsage
- ApacheHBaseHighNonHeapMemUsage
- ApacheHBaseDeadRegionServer
- ApacheHBaseOldRegionsInTransition
- ApacheHBaseHighMasterAuthFailureRate
- ApacheHBaseHighRSAuthFailureRate

## Apache HBase cluster overview
apache_hbase_cluster_overview_1
apache_hbase_cluster_overview_2

## Apache HBase RegionServer overview
apache_hbase_regionserver_overview_1
apache_hbase_regionserver_overview_2
apache_hbase_regionserver_overview_3

## Apache HBase logs overview
apache_hbase_logs_overview_1

### DIFF
--- a/apache-hbase-mixin/README.md
+++ b/apache-hbase-mixin/README.md
@@ -10,12 +10,12 @@ The Apache HBase mixin contains the following dashboards:
 
 and the following alerts:
 
-- ApacheHBaseHighHeapMemUsage
-- ApacheHBaseHighNonHeapMemUsage
-- ApacheHBaseDeadRegionServer
-- ApacheHBaseOldRegionsInTransition
-- ApacheHBaseHighMasterAuthFailureRate
-- ApacheHBaseHighRSAuthFailureRate
+- HBaseHighHeapMemUsage
+- HBaseHighNonHeapMemUsage
+- HBaseDeadRegionServer
+- HBaseOldRegionsInTransition
+- HBaseHighMasterAuthFailureRate
+- HBaseHighRSAuthFailureRate
 
 ## Apache HBase overview
 The Apache HBase cluster overview dashboard provides details on integration status/alerts, current RegionServers, JVM memory usage, cluster connections, master queue performance, and transitioning regions.
--- a/apache-hbase-mixin/README.md
+++ b/apache-hbase-mixin/README.md
@@ -54,8 +54,8 @@ scrape_configs:
       - targets: [localhost]
         labels:
           job: integrations/apache-hbase
+          hbase_cluster: "<your-cluster-name>"
           instance: "<your-instance-name>"
-          apache_hbase_cluster: "<your-cluster-name>"
           __path__: {hbase-home}/logs/*.log
     pipeline_stages:
         - multiline:
@@ -64,7 +64,7 @@ scrape_configs:
             expression: '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},\d{3} (?P<level>\w+)  \[(?P<context>.*)\] (?P<message>(?s:.*))$'
         - labels:
             level:
-            context:
+            logger:
 ```
 
 ## Alerts overview
--- a/apache-hbase-mixin/README.md
+++ b/apache-hbase-mixin/README.md
@@ -1,0 +1,122 @@
+# Apache HBase mixin
+
+The Apache HBase mixin is a set of configurable Grafana dashboards and alerts.
+
+The Apache HBase mixin contains the following dashboards:
+
+- Apache HBase cluster overview
+- Apache HBase RegionServer overview
+- Apache HBase logs
+
+and the following alerts:
+
+- ApacheHBaseHighHeapMemUsage
+- ApacheHBaseHighNonHeapMemUsage
+- ApacheHBaseDeadRegionServer
+- ApacheHBaseOldRegionsInTransition
+- ApacheHBaseHighMasterAuthFailureRate
+- ApacheHBaseHighRSAuthFailureRate
+
+## Apache HBase overview
+The Apache HBase cluster overview dashboard provides details on integration status/alerts, current RegionServers, JVM memory usage, cluster connections, master queue performance, and transitioning regions.
+
+![First screenshot of the Apache HBase cluster overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-hbase/screenshots/apache_hbase_cluster_overview_1.png)
+![Second screenshot of the Apache HBase cluster overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-hbase/screenshots/apache_hbase_cluster_overview_2.png)
+
+## Apache HBase RegionServer overview
+The Apache HBase RegionServer overview dashboard provides details on data regions, storage, connections, and request handling performance for a RegionServer node.
+
+![First screenshot of the Apache HBase RegionServer overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-hbase/screenshots/apache_hbase_region_server_overview_1.png)
+![Second screenshot of the Apache HBase RegionServer overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-hbase/screenshots/apache_hbase_region_server_overview_2.png)
+
+
+## Apache HBase logs
+The Apache HBase logs dashboard provides details on incoming system logs.
+
+![First screenshot of the Apache HBase logs dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/apache-hbase/screenshots/apache_hbase_logs_1.png)
+
+Apache HBase system logs are enabled by default in the `config.libsonnet` and can be removed by setting `enableLokiLogs` to `false`. Then run `make` again to regenerate the dashboard:
+
+```
+{
+  _config+:: {
+    enableLokiLogs: false,
+  },
+}
+```
+
+In order for the selectors to properly work for system logs ingested into your logs datasource, please also include the matching `instance`, `job`, and `apache_hbase_cluster` labels onto the [scrape_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
+
+```yaml
+scrape_configs:
+  - job_name: integrations/apache-hbase
+    static_configs:
+      - targets: [localhost]
+        labels:
+          job: integrations/apache-hbase
+          instance: "<your-instance-name>"
+          apache_hbase_cluster: "<your-cluster-name>"
+          __path__: {hbase-home}/logs/*.log
+    pipeline_stages:
+        - multiline:
+            firstline: '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},\d{3}'
+        - regex:
+            expression: '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},\d{3} (?P<level>\w+)  \[(?P<context>.*)\] (?P<message>(?s:.*))$'
+        - labels:
+            level:
+            context:
+```
+
+## Alerts overview
+
+- ApacheHBaseHighHeapMemUsage: There is a limited amount of heap memory available to the JVM.
+- ApacheHBaseHighNonHeapMemUsage: There is a limited amount of non-heap memory available to the JVM.
+- ApacheHBaseDeadRegionServer: One or more RegionServer(s) has become unresponsive.
+- ApacheHBaseOldRegionsInTransition: RegionServers are in transition for longer than expected.
+- ApacheHBaseHighMasterAuthFailureRate: A high percentage of authentication attempts to the master are failing.
+- ApacheHBaseHighRSAuthFailureRate: A high percentage of authentication attempts to a RegionServer are failing.
+
+Default thresholds can be configured in `config.libsonnet`.
+
+```js
+{
+  _config+:: {
+    alertsHighHeapMemUsage: 80        // percentage
+    alertsHighNonHeapMemUsage: 80     // percentage
+    alertsDeadRegionServer: 0         // count
+    alertsOldRegionsInTransition: 50  // percentage
+    alertsHighMasterAuthFailRate: 35  // percentage
+    alertsHighRSAuthFailRate: 35      // percentage
+  },
+}
+```
+
+## Install tools
+
+```bash
+go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb@latest
+go install github.com/monitoring-mixins/mixtool/cmd/mixtool@latest
+```
+
+For linting and formatting, you would also need `jsonnetfmt` installed. If you
+have a working Go development environment, it's easiest to run the following:
+
+```bash
+go install github.com/google/go-jsonnet/cmd/jsonnetfmt@latest
+```
+
+The files in `dashboards_out` need to be imported
+into your Grafana server. The exact details will be depending on your environment.
+
+`prometheus_alerts.yaml` needs to be imported into Prometheus.
+
+## Generate dashboards and alerts
+
+Edit `config.libsonnet` if required and then build JSON dashboard files for Grafana:
+
+```bash
+make
+```
+
+For more advanced uses of mixins, see
+https://github.com/monitoring-mixins/docs.
--- a/apache-hbase-mixin/alerts/alerts.libsonnet
+++ b/apache-hbase-mixin/alerts/alerts.libsonnet
@@ -1,0 +1,113 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'apache-hbase-alerts',
+        rules: [
+          {
+            alert: 'HighHeapMemUsage',
+            expr: |||
+              100 * sum without(context, hostname, processname) (jvm_metrics_mem_heap_used_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"} / clamp_min(jvm_metrics_mem_heap_committed_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"}, 1))  > %(alertsHighHeapMemUsage)s
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'There is a limited amount of heap memory available to the JVM.',
+              description:
+                (
+                  'The heap memory usage for the JVM on instance {{$labels.instance}} in cluster {{$labels.apache_hbase_cluster}} is {{printf "%%.0f" $labels.value}} percent, which is above the threshold of %(alertsHighHeapMemUsage)s'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'HighNonHeapMemUsage',
+            expr: |||
+              100 * sum without(context, hostname, processname) (jvm_metrics_mem_non_heap_used_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"} / clamp_min(jvm_metrics_mem_non_heap_committed_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"}, 1))  > %(alertsHighNonHeapMemUsage)s
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'There is a limited amount of non-heap memory available to the JVM.',
+              description:
+                (
+                  'The non-heap memory usage for the JVM on instance {{$labels.instance}} in cluster {{$labels.apache_hbase_cluster}} is {{printf "%%.0f" $labels.value}} percent, which is above the threshold of %(alertsHighNonHeapMemUsage)s'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'DeadRegionServer',
+            expr: |||
+              server_num_dead_region_servers{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster"} > %(alertsDeadRegionServer)s
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'One or more RegionServer(s) has become unresponsive.',
+              description:
+                (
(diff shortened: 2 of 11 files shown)


apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1255 @@
...
+            includeAll=true,
+            multi=true,
+            allValues='',
+            sort=0
+          ),
+          template.new(
+            'instance',
+            promDatasource,
+            'label_values(server_region_count{hbase_cluster=~"$hbase_cluster"},instance)',
+            label='Instance',
```
dpaasman00, line 1226: If this selector is supposed to be region servers, does it make sense to rename the label to be "Region server"? It can still be the `instance` selector, but the user facing name is "Region server"... what do you think?

Caleb-Hurshman: Went over this during spec review, Grafana still wants 'instance' to be the main secondary label. In this case, the relationship between a RS and an instance is one to one, so not a big deal to use instance.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+              description:
+                (
+                  'The non-heap memory usage for the JVM on instance {{$labels.instance}} in cluster {{$labels.hbase_cluster}} is {{printf "%%.0f" $labels.value}} percent, which is above the threshold of %(alertsHighNonHeapMemUsage)s'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'DeadRegionServer',
+            expr: |||
+              server_num_dead_region_servers > %(alertsDeadRegionServer)s
```
dpaasman00, line 44: this metric has a lot of labels associated with it... not sure if it's worthwhile removing some of them with a sum without or if you're happy with as is? also applies to the other non JVM alerts

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'RegionServers are in transition for longer than expected.',
+              description:
+                (
+                  '{{printf "%%.0f" $labels.value}} percent of RegionServers in transition in cluster {{$labels.hbase_cluster}} are transitioning for longer than expected, which is above the threshold of %(alertsOldRegionsInTransition)s'
```
dpaasman00, line 71: I'd either add the word "percent" or `%%s` to the end of this description since it is a percent threshold...also applies to the other alerts with percentage thresholds

apache-hbase-mixin/alerts/alerts.libsonnet
Caleb-Hurshman: Yeah there are quite a few, but I don't think any of them will alter the calculation, so they should be fine to keep around in case they provide some specific context to the user that could be useful.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+              description:
+                (
+                  '{{printf "%%.0f" $labels.value}} percent of authentication attempts to the master are failing in cluster {{$labels.hbase_cluster}}, which is above the threshold of %(alertsHighMasterAuthFailRate)s'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'HighRSAuthFailRate',
+            expr: |||
+              100 * rate(region_server_authentication_failures[5m]) / (clamp_min(rate(region_server_authentication_successes[5m] + clamp_min(rate(region_server_authentication_failures[5m]), 1)))) > %(alertsHighRSAuthFailRate)s
```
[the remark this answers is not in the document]
Caleb-Hurshman: Just pushed a change for this haha

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
```diff
@@ -0,0 +1,1034 @@
...
+      ],
+      fields: '',
+      values: false,
+    },
+    textMode: 'value',
+  },
+  pluginVersion: '10.2.0-61719',
+};
+
+local regionserversPanel = {
```
dpaasman00, line 115: Do you think it makes sense to rework the "Is master" column? I had a very similar column in a table for Atlas, and it got worked to report the type rather than true or false and to be color coded. Without knowing enough, looking at this table, I imagine it'd be similar to renaming the column to be "Context" and the value is either "Master" or "Region server" and one of them has the text colored yellow and the other green. 

Just something to keep in mind that a similar change may be requested, unless grafana specified the column to be done the current way.

dpaasman00 (commented):
2nd pass, dashboards looking good just 1 comment(not really a requested change, just a headsup). Few more comments on the alerts

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
Caleb-Hurshman: That's a good idea, but I think 'context' is less obvious than 'Is master'. I'll look into some other options for naming that column. There are also a bunch of other changes to this table that I'm going to go over with Vitaly, so I'll bring that up with him at that point.

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
```diff
@@ -0,0 +1,1032 @@
...
+      datasource=promDatasource,
+      legendFormat='{{hbase_cluster}}',
+      format='time_series',
+    ),
+  ],
+  type: 'alertlist',
+  title: 'Alerts',
+  description: 'Panel to report on the status of integration alerts.',
+  options: {
+    alertInstanceLabelFilter: '{job="integrations/apache-hbase"}',
```
v-zhuravlev, line 299: Looks like hardcoded value

Caleb-Hurshman: Currently yes. This is our first round of integrations implementing this new alert panel, so if there is an example implementation to go off of that would be very helpful.

v-zhuravlev: You can try using same dynamic filter you use for that dashboard:
`{job=~"${job:regex}", hbase_cluster=~"${hbase_cluster:regex}", instance=~"${instance:regex}"}`
note usage of 'advanced syntax'  though as it doesn't work otherwise.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'apache-hbase-alerts',
+        rules: [
+          {
+            alert: 'HighHeapMemUsage',
+            expr: |||
+              100 * sum without(context, hostname, processname) (jvm_metrics_mem_heap_used_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"} / clamp_min(jvm_metrics_mem_heap_committed_m{job=~"$job", apache_hbase_cluster=~"$apache_hbase_cluster", instance=~"$instance"}, 1))  > %(alertsHighHeapMemUsage)s
```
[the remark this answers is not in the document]
Caleb-Hurshman: It looks like this expression was changed to remove the selectors. Let me know if this feedback still applies and how that would look.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'RegionServers are in transition for longer than expected.',
+              description:
+                (
+                  '{{printf "%%.0f" $labels.value}} percent of RegionServers in transition in cluster {{$labels.hbase_cluster}} are transitioning for longer than expected, which is above the threshold of %(alertsOldRegionsInTransition)s'
```
dpaasman00, line 71: missed one description to add percent to

dpaasman00 (approved):
LGTM after the 1 alert description

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
```diff
@@ -0,0 +1,1032 @@
...
+          '__name__ 2': true,
+          clusterid: true,
+          'clusterid 1': true,
+          'clusterid 2': true,
+          context: true,
+          'context 1': true,
+          'context 2': true,
+          hbase_cluster: true,
+          'hbase_cluster 1': true,
+          'hbase_cluster 2': true,
```
JonathanWamsley, line 233: Should the RegionServers chart be larger and display which HBase cluster it belongs to? I am guessing each cluster has 1 master, which the Hbase RegionServer overview looks at those master stats then?

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1309 @@
...
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'server_region_count{job=~"$job", hbase_cluster=~"$hbase_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'stat',
```
JonathanWamsley, line 25: The RegionServer overview I am a little concerned with scalability but I am not familiar with how many instances are expected in an environment. I notice you have 3 stat panels that may scale weirdly if you have multiple instances and clusters. How many instances are expected with hbase at a time and would it make sense to change that to a scrollable table showing the instance and values for the first 2 panels, and maybe horizontal scrollable bar for the `Store file size` chart?

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1309 @@
...
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'server_store_file_count{job=~"$job", hbase_cluster=~"$hbase_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'stat',
```
JonathanWamsley, line 79: same question from before

JonathanWamsley (commented):
Overall looks pretty good with a couple of minor comments

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
Caleb-Hurshman: I can definitely make the bar gauge horizontal, but the rest of the panels in the top row were specifically requested as stat panels by Grafana.

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
Caleb-Hurshman: I'd be in favor of expanding the table, just need to figure out what else to re-size. Master stats are actually shown on the cluster overview, not on the RegionServer dashboard.

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
JonathanWamsley: I just did a quick check, and saw that you might start with a minimum of three RegionServer instances for fault tolerance and scalability. A larger clusters may have tens or even hundreds of RegionServer instances. So that dashboard may only be usable when selecting for a small amount of instance anyways.

JonathanWamsley (approved):
LGTM 👍 Nice work @Caleb-Hurshman

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1306 @@
...
+    prometheus.target(
+      'server_region_count{job=~"$job", hbase_cluster=~"$hbase_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'stat',
+  title: 'Regions',
+  description: 'The number of regions hosted by the RegionServer.',
```
mshahzeb, line 27: In descriptions and panel titles I think we should use the term as "Region Server" instead of "RegionSever" since these are display names and they should not be formatted like variables IMO

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1306 @@
...
+    prometheus.target(
+      'server_store_file_count{job=~"$job", hbase_cluster=~"$hbase_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'stat',
+  title: 'Store files',
+  description: 'The number of store files on disk currently managed by the RegionServer.',
```
mshahzeb, line 81: Same comment as above

apache-hbase-mixin/dashboards/apache-hbase-regionserver-overview.libsonnet
```diff
@@ -0,0 +1,1306 @@
...
+    prometheus.target(
+      'server_store_file_size{job=~"$job", hbase_cluster=~"$hbase_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'bargauge',
+  title: 'Store file size',
+  description: 'The total size of the store files on disk managed by the RegionServer.',
```
mshahzeb, line 135: Same comment as above

mshahzeb (approved):
LGTM!

Just one suggestion on the word "RegionServer" showing up in description & titles

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
```diff
@@ -0,0 +1,1035 @@
...
+      firing: true,
+      noData: true,
+      normal: true,
+      pending: true,
+    },
+    viewMode: 'list',
+  },
+};
+
+local jvmMemoryUsagePanel = {
```
Caleb-Hurshman, line 320: @v-zhuravlev @mshahzeb I'm considering removing the non-heap component of this dashboard, as it is always reporting >90%. That would also mean probably removing the related alert.

apache-hbase-mixin/dashboards/apache-hbase-logs-overview.libsonnet
Caleb-Hurshman, line 1: HBase has a lot of options that can show up in the 'context' field for the current regex, I'll be looking into different ways to make this selector more succinct, open to suggestions.

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
v-zhuravlev, line 1: I tried to use link in the table to jump to specific host. But looks like it just points to another dashboard, without instance variable nor datetime. Can you please add more contextual jump? (add instance variable and datetime selected). Thanks!

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'apache-hbase-alerts',
+        rules: [
+          {
+            alert: 'HighHeapMemUsage',
```
v-zhuravlev, line 8: Best if we prefix alerts  with 'HBase'.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,113 @@
...
+            expr: |||
+              100 * sum without(context, hostname, processname) (jvm_metrics_mem_heap_used_m / clamp_min(jvm_metrics_mem_heap_committed_m, 1))  > %(alertsHighHeapMemUsage)s
```
v-zhuravlev, line 10: These metrics look generic too me. To make sure we don't leave chances that there could be other integration having same java metrics (instrumented same way as Hbase), best if we add filtering selector to each alert rule. That we we could be confident that there would no potential overlaps. And Hbase alerts woudn't fire on some other JVMs unrelated to Hbase.

apache-hbase-mixin/dashboards/apache-hbase-logs-overview.libsonnet
```diff
@@ -0,0 +1,32 @@
+local g = import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet';
+local logsDashboard = import 'github.com/grafana/jsonnet-libs/logs-lib/logs/main.libsonnet';
+{
+  grafanaDashboards+::
+    if $._config.enableLokiLogs then {
+      local apacheHBaseLogs =
+        logsDashboard.new(
+          'Apache HBase logs overview',
+          datasourceName='loki_datasource',
+          datasourceRegex='(?!grafanacloud.+usage-insights|grafanacloud.+alert-state-history).+',
```
v-zhuravlev, line 10: This is not required here. Such filters are applied automatically on integrations phase, by one of the utils functions.

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
```diff
@@ -0,0 +1,1035 @@
...
+      },
+    },
+  ],
+};
+
+local alertsPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'master_num_open_connections{job=~"$job", hbase_cluster=~"$hbase_cluster"}',
```
v-zhuravlev, line 291: Is target needed in alertsPanel?

v-zhuravlev (changes requested):
Looks cool, thank you! Some suggestions below:

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
Caleb-Hurshman: @v-zhuravlev I can do this for region servers, but not for the master. As we discussed during spec review, since the master is technically not a RegionServer, it won't have a link to another dashboard. You mentioned possibly using separate columns?

apache-hbase-mixin/alerts/alerts.libsonnet
Caleb-Hurshman: I understand what you mean, but couldn't find any examples of this in other mixins. Made an attempt in the most recent commit, let me know if it needs to be changed again.

apache-hbase-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,96 @@
+{
+  prometheusAlerts+:: {
+    groups+: [
+      {
+        name: 'apache-hbase-alerts',
+        rules: [
+          {
+            alert: 'HBaseHighHeapMemUsage',
+            expr: |||
+              100 * sum without(context, hostname, processname) (jvm_metrics_mem_heap_used_m{$._config.filterSelector} / clamp_min(jvm_metrics_mem_heap_committed_m{$._config.filterSelector}, 1))  > %(alertsHighHeapMemUsage)s
```
v-zhuravlev, line 10:
```suggestion
              100 * sum without(context, hostname, processname) (jvm_metrics_mem_heap_used_m{%(filterSelector)s} / clamp_min(jvm_metrics_mem_heap_committed_m{%(filterSelector)s}, 1))  > %(alertsHighHeapMemUsage)s
```

apache-hbase-mixin/README.md
```diff
@@ -0,0 +1,122 @@
...
+          __path__: {hbase-home}/logs/*.log
+    pipeline_stages:
+        - multiline:
+            firstline: '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},\d{3}'
+        - regex:
+            expression: '\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2},\d{3} (?P<level>\w+)  \[(?P<context>.*)\] (?P<message>(?s:.*))$'
+        - labels:
+            level:
+            context:
+```
```
v-zhuravlev, line 68: This needs to be updated with `logger` instead of `context`

apache-hbase-mixin/dashboards/apache-hbase-cluster-overview.libsonnet
v-zhuravlev: yes, could work with two rows.

apache-hbase-mixin/dashboards/apache-hbase-logs-overview.libsonnet
v-zhuravlev: 1) I think logger should be not `regionserver` but `regionserver.HStore`. can you update parser?
2) in test grafana, `Instance` label is not filled
3) in test grafana, there is `context` label among templated variables, but not `logger`.

v-zhuravlev (changes requested):

Caleb-Hurshman:
@v-zhuravlev Changes to the logs dashboard and JVM alert have been pushed. Still need guidance on how to implement the links and status history for the RegionServer table on the cluster overview dashboard.

v-zhuravlev (approved):

Dasomeone:
Approved the CI :)

PR #1086, opened 2023-10-17
Gitlab: update multi select instance panel & have multicluster to be false

JonathanWamsley:
- update multi select instance panel. This is needed because only one instance can be selected at once, which makes it awkward when selecting different cluster selectors because you also have to adjust the instance selector to currently.
- multicluster is set to false


rgeyer (approved):
LGTM, good catch!

PR #1087, opened 2023-10-17
Presto Mixin

BominRahmani:
The following PR adds dashboards and alerts for the Presto mixin

The Presto mixin contains the following dashboards:

- Presto overview
- Presto coordinator overview
- Presto worker overview
- Presto logs overview

and the following alerts:

- PrestoHighInsufficientResources
- PrestoHighTaskFailuresWarning
- PrestoHighTaskFailuresCritical
- PrestoHighQueuedTaskCount
- PrestoHighBlockedNodes
- PrestoHighFailedQueriesWarning
- PrestoHighFailedQueriesCritical

## Presto overview
image
image

## Presto coordinator overview
image
image

## Presto worker overview
image
image

## Presto logs overview
image


dpaasman00 (commented):
initial comments along with what we talked about earlier

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,837 @@
...
+    prometheus.target(
+      'rate(com_facebook_presto_execution_TaskManager_OutputPositions_OneMinute_Count{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}[$__rate_interval])',
+      datasource=promDatasource,
+      legendFormat='{{presto_cluster}} - {{instance}}',
+      format='time_series',
+    ),
+  ],
+  type: 'timeseries',
+  title: 'Output positions',
+  description: 'The number of rows (or records) produced by an operation.',
```
dpaasman00, line 262:
```suggestion
  description: 'The rate of rows (or records) produced by an operation.',
```

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,837 @@
...
+    ),
+    prometheus.target(
+      'com_facebook_presto_execution_executor_TaskExecutor_ProcessorExecutor_PoolSize{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{presto_cluster}} - {{instance}} - process executor',
+    ),
+  ],
+  type: 'timeseries',
+  title: 'Executor pool size',
+  description: 'The pool size of the task notification executor, and process executor.',
```
dpaasman00, line 350:
```suggestion
  description: 'The pool size of the task notification executor and process executor.',
```

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,837 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local memoryPoolPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'sum by (instance, presto_cluster) (com_facebook_presto_memory_MemoryPool_FreeBytes{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"})',
```
dpaasman00, line 420: do these queries need to be changed? What's the point of filtering on `instance` and `presto_cluster` if we're summing by them?

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,837 @@
...
+          ),
+          template.new(
+            'presto_cluster',
+            promDatasource,
+            'label_values(com_facebook_presto_failureDetector_HeartbeatFailureDetector_ActiveCount{job=~"$job"},presto_cluster)',
+            label='Presto cluster',
+            refresh=2,
+            includeAll=true,
+            multi=true,
+            allValues='.*',
```
dpaasman00, line 806: do you know the difference between setting this as `.*` vs `.+`?

dpaasman00 (commented):
End of first pass - some nits

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,837 @@
...
+      thresholds: {
+        mode: 'absolute',
+        steps: [
+          {
+            color: 'green',
+            value: null,
+          },
+        ],
+      },
+      unit: 'cps',
```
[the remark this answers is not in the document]
BominRahmani: yes, I can change it to Ops/s as discussed.

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1218 @@
...
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'com_facebook_presto_execution_QueryManager_CompletedQueries_OneMinute_Count{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}',
+      datasource=promDatasource,
+      legendFormat='{{presto_cluster}} - {{instance}} - completed',
+      format='time_series',
+    ),
+    prometheus.target(
+      'com_facebook_presto_execution_QueryManager_RunningQueries{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}',
```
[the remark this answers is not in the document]
BominRahmani: no one_minute for this metric.

presto-mixin/dashboards/presto-worker-overview.libsonnet
BominRahmani: we still need to filter based off instance and presto cluster to keep the results as those chosen by the selector no?

presto-mixin/dashboards/presto-logs-overview.libsonnet
```diff
@@ -0,0 +1,32 @@
...
+        {
+          panels+:
+            {
+              logs+:
+                // presto logs already have timestamp
+                g.panel.logs.options.withShowTime(false),
+            },
+          dashboards+:
+            {
+              logs+: g.dashboard.withLinksMixin($.grafanaDashboards['presto-overview.json'].links)
```
Caleb-Hurshman, line 25: Links aren't showing up right now on the logs dashboard on the cloud.

presto-mixin/dashboards/presto-worker-overview.libsonnet
dpaasman00: yea no you're good, i think i was confusing sum and sum without when i wrote that

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,943 @@
...
+};
+
+local alertsPanel = {
+  datasource: promDatasource,
+  targets: [],
+  type: 'alertlist',
+  title: 'Alerts',
+  description: 'Reports firing alerts.',
+  options: {
+    alertInstanceLabelFilter: '',
```
Caleb-Hurshman, line 531: Be sure to add your filter here

presto-mixin/dashboards/presto-worker-overview.libsonnet
dpaasman00: feedback I've gotten of late is to have it set to `.+`

BominRahmani: set this as non multi. But previously i had set this to .* because thats what i've been told on a previous integration.

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,943 @@
...
+      thresholds: {
+        mode: 'absolute',
+        steps: [
+          {
+            color: 'green',
+            value: null,
+          },
+        ],
+      },
+      unit: 'err / s',
```
Caleb-Hurshman, line 506: Is this a custom unit? just wondering about the spacing around the '/'. It seems like it should be `err/s` to fit with other units.

presto-mixin/README.md
```diff
@@ -0,0 +1,129 @@
...
+- PrestoHighInsufficientFailures
+- PrestoHighTaskFailuresWarning
+- PrestoHighTaskFailuresCritical
+- PrestoHighQueuedTaskCount
+- PrestoHighBlockedNodes
+- PrestoHighFailedQueriesWarning
+- PrestoHighFailedQueriesCritical
+
+## Presto overview
+The Presto cluster overview dashboard provides details on integration status/alerts, workers/coordinators, error failures, data throughput, blocked nodes, and distributed bytes.
```
Caleb-Hurshman, line 23: I'd probably stick to just calling this 'Presto overview' as I don't see the 'cluster' being included elsewhere.

Caleb-Hurshman (commented):
A few small comments, looks good overall

presto-mixin/dashboards/presto-logs-overview.libsonnet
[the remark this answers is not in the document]
Caleb-Hurshman: I'm also not seeing these links

BominRahmani: Yea, will need to re import this dashboard so the uid changes from random string to name of dashboard. Code is correct, just need to update shared instance dashboard.

presto-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,130 @@
...
+              description:
+                (
+                  'The number of task failures on {{$labels.instance}} is {{ printf "%%.0f" $value }} which is above the threshold of %(alertsHighTaskFailuresCritical)s%%s.'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'PrestoHighQueuedTaskCount',
+            expr: |||
+              increase(com_facebook_presto_execution_QueryManager_QueuedTaskCount[5m]) > %(alertsHighQueuedTaskCount)s
```
dpaasman00, line 61: going as far back as 7 days, no data showing up for this query (no threshold). have you been able to get load on this metric? Looks like a pretty simple query, so either you just haven't been able to get load on it or is there a typo in the metric name?

presto-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,130 @@
...
+              description:
+                (
+                  'The number of queued tasks on {{$labels.instance}} is {{ printf "%%.0f" $value }} which is greater than the threshold of %(alertsHighQueuedTaskCount)s'
+                ) % $._config,
+            },
+          },
+          {
+            alert: 'PrestoHighBlockedNodes',
+            expr: |||
+              increase(com_facebook_presto_memory_ClusterMemoryPool_BlockedNodes[5m]) > %(alertsHighBlockedNodesCount)s
```
dpaasman00, line 78: this metric has an additional label on it called `name` which had 2 values - "general" and "reserved". Does that label need to be summed out or maybe included in the description?

dpaasman00 (approved):
Tentative approval - couple alerts may or may not need to be reworked, otherwise LGTM

Caleb-Hurshman (approved):
Approved, pending some fixes to alerts that Dakota brought up

gaantunes:
@BominRahmani what is the current scrape interval set for this sample app? I want to make sure the JMX configuration we provide with this integration is optimized for a reasonable amount of time, so the SUO does not loose too much performance from gererating those metrics.

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local userErrorFailuresOneMinuteRatePanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'com_facebook_presto_execution_QueryManager_UserErrorFailures_OneMinute_Rate{job=~"$job", presto_cluster=~"$presto_cluster"}',
```
gaantunes, line 372: Could you give us an update in regards to any difference between results from pre calculate rate against Grafana calculating it?

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
Dasomeone, line 1: For all the panels on this dashboard, can you please remove the `{{presto_cluster}}` from all legends?

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+      fields: '',
+      values: false,
+    },
+    showThresholdLabels: false,
+    showThresholdMarkers: true,
+  },
+  pluginVersion: '10.2.0-62263',
+};
+
+local normalQueryOneMinuteCountPanel = {
```
Dasomeone, line 127: So for this dashboard and the four panels at the top:
1) Can you remove min/max/mean from these panels
2) change to list instead of table
3) change visualisation type to line with 90 deg points (see example)

image

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local cpuInputThroughputOneMinuteCountPanel = {
```
Dasomeone, line 674: From this panel and all the remaining below, we should again remove `{{presto_cluster}}` from the legend, and also make the legends lists that are bottom aligned, as there's no table values that we care to display, and the chance of multiple instances causing a multi-line break is quite low.

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+
+local dashboardUid = 'presto-coordinator-overview';
+
+local promDatasourceName = 'prometheus_datasource';
+
+local promDatasource = {
+  uid: '${%s}' % promDatasourceName,
+};
+
+local nonheapMemoryUsagePanel = {
```
Dasomeone, line 15: For all the gauges can you please verify the thresholds are set appropriately? Currently they're all using the panel defaults for yellow and red at 70% and 80% respectively

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
Dasomeone, line 1: Another big general comment that encompasses multiple panels:
we'd like to see the gauges increased in size, the error rate moved up and styled accordingly to this picture, and the bar charts changed to line charts as seen and described in another comment

image

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local memoryPoolPanel = {
```
Dasomeone, line 752: This panel is identical to the worker dashboard panel, so we'd prefer it's only shown there as there's nothing specific to it being on the coordinator overview

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+      sort: 'desc',
+    },
+  },
+};
+
+local cpuInputThroughputOneMinuteCountPanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'com_facebook_presto_execution_QueryManager_CpuInputByteRate_OneMinute_Total{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}',
```
Dasomeone, line 678: This is a rate metric and should as such probably use the throughput byte unit, rather than static byte unit

gaantunes:
Could you please spin up a cluster where the coordinator and worker nodes are running on different instances? This would simulated a more realistic environment.

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
```diff
@@ -0,0 +1,1343 @@
...
+      thresholds: {
+        mode: 'absolute',
+        steps: [
+          {
+            color: 'green',
+            value: null,
+          },
+        ],
+      },
+      unit: '/ sec',
```
Dasomeone, line 656: This panel, description, and units are really difficult to understand. Can you please update the description and use a more appropriate rate unit for this?

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,957 @@
...
+
+local nonheapMemoryUsagePanel = {
+  datasource: promDatasource,
+  targets: [
+    prometheus.target(
+      'java_lang_Memory_NonHeapMemoryUsage_used{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"} / clamp_min((java_lang_Memory_NonHeapMemoryUsage_used{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"} + java_lang_Memory_NonHeapMemoryUsage_committed{job=~"$job", presto_cluster=~"$presto_cluster", instance=~"$instance"}),1)',
+      datasource=promDatasource,
+    ),
+  ],
+  type: 'gauge',
```
Dasomeone, line 23: Same comments as on other dashboards, gauges take up too much space and should have adjusted non-default thresholds

presto-mixin/dashboards/presto-worker-overview.libsonnet
Dasomeone, line 1: Couple big changes suggested here, and we've saved the dashboard on the shared instance so you can view them.

Tl;dr:
- Combined failed/completed tasks in one panel with separate axis and specific colour coding
- Combined rows 1 and 2 into a single row and changed the height to match
- Removed `{{presto_cluster}}` from legends (and needs to be done throughout, only top two rows one as an example
- Adjust legends to be bottom lists, as this is a drilldown dashboard and we don't mind those running into multiple lines if people don't drill down

image

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,957 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local dataThroughputPanel = {
```
Dasomeone, line 609: Can you double check the metric and unit for this? it's using that `TotalCount` metric and it's not obvious if it's correct

presto-mixin/dashboards/presto-worker-overview.libsonnet
```diff
@@ -0,0 +1,957 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local jvmMetricsRow = {
```
gaantunes, line 692: Can you include GC panels for this one as well?

presto-mixin/dashboards/presto-logs-overview.libsonnet
```diff
@@ -0,0 +1,32 @@
+local g = import 'github.com/grafana/grafonnet/gen/grafonnet-latest/main.libsonnet';
+local logsDashboard = import 'github.com/grafana/jsonnet-libs/logs-lib/logs/main.libsonnet';
+{
+  grafanaDashboards+::
+    if $._config.enableLokiLogs then {
+      local prestoLogs =
+        logsDashboard.new(
+          'Presto logs overview',
+          datasourceName='loki_datasource',
+          datasourceRegex='',
```
Dasomeone, line 10: Missing the datasource regex needed. Currently defaults to `alert-state-history`

presto-mixin/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+## Presto coordinator overview
+The Presto coordinator overview dashboard provides details on various query counts and rates, query execution time, CPU time consumed, CPU input throughput, error failures, JVM metrics, and memory pool information.
+![First screenshot of the Presto coordinator overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/presto/screenshots/presto_coordinator_overview_1.png)
+![Second screenshot of the Presto coordinator overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/presto/screenshots/presto_coordinator_overview_2.png)
+
+## Presto worker overview
+The Presto worker overview dashboard provides details on various task rates, pool sizes, output positions, data throughput, JVM metrics, and memory pool information
+![First screenshot of the Presto coordinator overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/presto/screenshots/presto_worker_overview_1.png)
+![Second screenshot of the Presto coordinator overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/presto/screenshots/presto_worker_overview_2.png)
```
Dasomeone, line 35: Screenshot URLs 404 since these have not yet been uploaded. Let's remove that until we're 100% on the number of screenshots and naming, just to prevent accidents

presto-mixin/dashboards/presto-overview.libsonnet
Dasomeone, line 1: All legends should include k8s cluster label

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
Dasomeone, line 1: Can you rename this to `Presto coordinator`. No need to call it an overview

presto-mixin/dashboards/presto-overview.libsonnet
BominRahmani: Yea, it seems they are about the same. Using rate from Grafana combined with totalCount metric shows this
image
where the first two peaks are 2.22, Keeping in mind this uses rate_interval.
Using the precalculated rate shows this
image
excluding the first peak that started  at 14:25, the next two peaks are 6.43 and 3.65 respectively, the sum of these two would be 10.08, divide it by four since rate interval of the first is 4x the one minute, the answer is around 2.52. which is close to the peaks of the 2.22 each. I suspect the variance comes from the reference of when the calculations start. I also noticed Data dogs using the One_Minute version of the metrics, which i assume is to save on the calculations.

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
BominRahmani: Wanted to give it a single pass before I added all the k8's cluster stuff. Sorry should have communicated that better. (adding in the matchers and stuff is a bit of a pain especially if panels need to be taken out/added in!)
All of those have been added in as of now and the Cloud environment has the k8 option enabled even though there is no cluster currently, for observation purposes.)

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+  targets: [
+    prometheus.target(
+      'com_facebook_presto_metadata_DiscoveryNodeManager_ActiveResourceManagerCount{job=~"$job", presto_cluster=~"$presto_cluster"}',
+      datasource=promDatasource,
+      legendFormat='{{presto_cluster}}',
+      format='time_series',
+    ),
+  ],
+  type: 'stat',
+  title: 'Active resource managers',
```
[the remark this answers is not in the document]
BominRahmani: I'm not sure it would make sense in this context. The amount of resource managers is completely dependent on the type of presto environment, the thresholds of red/green wouldn't really hold too well. In the case of the coordinator, i can see having 0 as being red and anything over being green, but then that as a gauge seems out of place with the resource manager being a stat. The same can be done for workers with 0 for red and green for anything over, but then again, it seems overkill and out of place as a gauge. Feel free to push back on this, but how about this: i keep these as stat panels and add colors to number on the stat depending on what it is.

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local alertsPanel = {
```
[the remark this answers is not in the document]
BominRahmani: Yea I can definitely move this up! Alerts have been added and verified (I can change the parameters to purposely make it fire for the sake of this mixin review if you'd like), the thing is the alerts only show when firing when you add a instance label filter. I asked vitaly about this during the last integration since the alerts show up even when normal/pending if you choose a datasource folder and no filter selector, but he told me that adding in the that instance label filter makes it so these arent used anymore 
image

presto-mixin/dashboards/presto-overview.libsonnet
```diff
@@ -0,0 +1,962 @@
...
+      showLegend: true,
+    },
+    tooltip: {
+      mode: 'multi',
+      sort: 'desc',
+    },
+  },
+};
+
+local dataThroughputPanel = {
```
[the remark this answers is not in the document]
BominRahmani: 1. New title, what do you think: "Input/Output data size - one minute rate"
2. That is weird, I think this is a one off when metrics were reported maybe due to a very small query being run in that time range? When i look at this graph i see a y axis with the below values during the time period of 
image
image
3. It showed the total bytes, but i had applied the rate function to it, I found that there is actually an innate metric that will do this for me called `com_facebook_presto_execution_TaskManager_OutputDataSize_OneMinute_Rate`, So i have chosen to replace the totalcount with a rate function to this.

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
BominRahmani: Yea these should be good. Generally from what i've seen, anywhere from 0-70% could be considered safe on heap/non-heap, with 80-90% considered warning levels, and above being bad!

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
BominRahmani: I altered this panel a bit to make it more clear.
1. Clarified on the description
2. Changed the query from a OneMinute_Rate -> OneMinute_Count. Since the metric itself originally returns seconds, a rate on top of that would make it sec/sec or / sec which can be a bit confusing. By changing this into the OneMinute_Count i can just keep a flat sec as the unit while making the panel overall make a bit more sense.
Let me know what you think

presto-mixin/dashboards/presto-worker-overview.libsonnet
BominRahmani: @gaantunes Of course! Also this new design looks really cool! @Dasomeone. I'll get a similar look implemented for the Coordinator dashboard.

presto-mixin/dashboards/presto-worker-overview.libsonnet
BominRahmani: 👍 Changed the metric to be using OneMinute_Rate, units should be good. Also changed graph type to line with 90 degrees since.

presto-mixin/dashboards/presto-logs-overview.libsonnet
BominRahmani: image
I was going to add that, but I had noticed that Vitaly commented this on another integration thats in progress.

presto-mixin/dashboards/presto-overview.libsonnet
gaantunes: Cool that's good, let's move forward with the pre calculated metrics.

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
Dasomeone: Yep that makes sense, thanks for explaining!
Let's stick to coloured stat panels then, though I might still suggest dropping the graph background

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
Dasomeone: If you could set it to fire and add a screenshot here of it working that'd be grand! 😄

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
Dasomeone: 1) I think I'd phrase it more along the lines of "Data processing throughput" or similar? More descriptive of what's happening. No need to mention input/output either I think, as that's in the legend. Big fan of including the "one minute rate" description through!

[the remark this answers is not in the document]
Dasomeone: 2) That's still weird in your screenshot, because that unit is in milliBytes/s, as in 1000th of a Byte/s 😅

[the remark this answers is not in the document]
Dasomeone: 3. Perfect, thank you for checking! Rate metric is a better choice for this for sure :)

presto-mixin/dashboards/presto-coordinator.libsonnet
```diff
@@ -124,52 +125,46 @@ local heapMemoryUsagePanel = {
...
+      'presto_QueryManager_UserErrorFailures_OneMinute_Count{' + matcher + ', presto_cluster=~"$presto_cluster"}',
       datasource=promDatasource,
-      legendFormat='{{presto_cluster}} - {{instance}} - started',
+      legendFormat='' + legendMatcher + ' - user',
     ),
   ],
   type: 'timeseries',
-  title: 'Normal query - one minute count',
-  description: 'A count of completed, running, and started queries.',
+  title: 'Error failures - one minute count',
```
Dasomeone, line 144: Can you change this to avoid the double negative please? Just "Failures - one minute count" is fine

presto-mixin/dashboards/presto-coordinator-overview.libsonnet
Dasomeone: Yep that works, thank you for explaining :)

presto-mixin/dashboards/presto-worker-overview.libsonnet
Dasomeone: Same comment here re: naming of the panel, I think "data processing throughput" or similarly descriptive while dropping "input/output" will work.

Similarly, can you check what unit the metric is? Same reasoning as the other panel re: bytes & millibytes

Dasomeone (commented):
Couple comments left, but massive improvement so far! :)

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
BominRahmani: image

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
BominRahmani: 2. Oh... that is strange. Unfortunately there is little documentation surrounding these metrics. From the bit of digging i've done
image
Qubole and Datadogs both have this type of metric shown as bytes
image
and likewise metrics within Presto are measured in Bytes as well such as "com_facebook_presto_execution_QueryManager_CpuInputByteRate"

gaantunes (approved):
LGTM, thank you!

presto-mixin/dashboards/presto-overview.libsonnet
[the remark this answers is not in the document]
Dasomeone: Hmmm that is weird, maybe this is just an artefact of low amount of data given the lack of a high utilization sample app? 
Let's keep it as-is for now, but note it down as something to keep an eye on :)

Dasomeone (commented):
LGTM, we just need to get some updated screenshots taken and included, and we need to remember to keep an eye on the units where it's been acting weirdly, but I think it's otherwise great!

Dasomeone (approved):
Looks fantastic, thanks for all the iteration on this one :D

PR #1090, opened 2023-10-25
Improve disk space table

v-zhuravlev:
Can provide freeTarget instead of usageTarget as an option.

metrics for available bytes is actually present in windows/linux exporter.

```
      diskUsage:
        commonlib.panels.disk.table.usage.new(
          totalTarget=
          (
            t.diskTotal
            + g.query.prometheus.withFormat('table')
            + g.query.prometheus.withInstant(true)
          ),
          freeTarget=
          t.diskFree
          + g.query.prometheus.withFormat('table')
          + g.query.prometheus.withInstant(true),
          groupLabel='mountpoint'
          ,
          description='Disk utilisation in percent, by mountpoint. Some duplication can occur if the same filesystem is mounted in multiple locations.'
        ),
```

image


rgeyer (approved):
LGTM, nice improvement!

PR #1095, opened 2023-10-31
time: add `parse_rfc3339` function

colega:
Parsing RFC3339 timestamps can be useful sometimes.

I manually tested the wrong formats, I don't know how to keep tests for that (as the entire jsonnet fails).

### DIFF
--- a/time/date.jsonnet
+++ b/time/date.jsonnet
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
 // days_seconds returns the number of seconds in all days up to day-1.
 local days_seconds(day) = (day - 1) * 24 * 3600;
 
+// is_numeric returns a bool indicating whether the input string is numeric
+local is_numeric(input) = std.length(input) > 0 && std.length(std.prune([
+  if std.codepoint('0') > std.codepoint(char) || std.codepoint(char) > std.codepoint('9') then char
+  for char in std.stringChars(input)
+])) == 0;
+
+// parse_date_or_time parses input which has part_names separated by sep.
+local parse_date_or_time(input, sep, part_names) = (
+  local parts = std.split(input, sep);
+  assert std.length(parts) == std.length(part_names) : 'expected %(expected)d parts separated by %(sep)s in %(format)s formatted input "%(input)s", but got %(got)d' % {
+    expected: std.length(part_names),
+    sep: sep,
+    format: std.join(sep, part_names),
+    input: input,
+    got: std.length(parts),
+  };
+
+  {
+    [part_names[i]]:
+      // Fail with meaningful message if not numeric, otherwise it will be a hell to debug.
+      assert is_numeric(parts[i]) : '%(part_name)%s part "%(part)s" of %(format)s of input "%(input)s" is not numeric' % {
+        part_name: part_names[i],
+        part: parts[i],
+        format: std.join(sep, part_names),
+        input: input,
+      };
+      std.parseInt(parts[i])
+    for i in std.range(0, std.length(parts) - 1)
+  }
+);
+
 {
-  // date_to_unix_timestamp transforms a date into a unix timestamp.
+  // to_unix_timestamp transforms a date into a unix timestamp.
   to_unix_timestamp(year, month, day, hour, minute, second)::
     years_seconds(year) + months_seconds(year, month) + days_seconds(day) + hour * 3600 + minute * 60 + second,
+
+  // parse_rfc3339 parses an RFC3339 timestamp in UTC (ending in 'Z') with date and time separated by 'T', like '2006-01-02T15:04:05Z'
+  // The returned object has "year", "month", "day", "hour", "minute" and "second" keys, as well as a to_unix_timestamp method that returns the unix timestamp of the parsed date.
+  parse_rfc3339(input)::
+    assert std.endsWith(input, 'Z') : 'the provided RFC3339 "%s" should end with "Z"' % input;
+    local datetime = std.split(std.substr(input, 0, std.length(input) - 1), 'T');
+    assert std.length(datetime) == 2 : 'the provided RFC3339 timestamp "%s" does not have date and time parts separated by the character "T"' % input;
+
+    local date = parse_date_or_time(datetime[0], '-', ['year', 'month', 'day']);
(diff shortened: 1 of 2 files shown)


time/date.jsonnet
```diff
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
 // days_seconds returns the number of seconds in all days up to day-1.
 local days_seconds(day) = (day - 1) * 24 * 3600;
 
+// is_numeric returns a bool indicating whether the input string is numeric
+local is_numeric(input) = std.length(input) > 0 && std.length(std.prune([
+  if std.codepoint('0') > std.codepoint(char) || std.codepoint(char) > std.codepoint('9') then char
+  for char in std.stringChars(input)
+])) == 0;
```
jdbaldry, line 41: I don't think we tend to be particularly defensive with our functions but I would consider adding a type assertion for string input.

```jsonnet
local is_numeric(input) =
  local type = std.type(input);
  assert type == 'string' : 'is_numeric() only operates on strings, got %s' % type;
  EXPR;
```

For the per codepoint check, I think it would be more idiomatic in functional programming to use a fold function. The following `foldl` starts with an accumulator, `acc`, set to the boolean result of `std.length(input) > 0` and folds the result of the test for digits with that accumulator. I've inverted the test from your code because I personally find it easier to think in this way but you can flip it back if you like.

```jsonnet
std.foldl(
  function(acc, char) acc && std.codepoint('0') <= std.codepoint(char) && std.codepoint(char) <= std.codepoint('9'),
  std.stringChars(input),
  std.length(input) > 0,
)
```

time/date.jsonnet
```diff
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
...
 {
-  // date_to_unix_timestamp transforms a date into a unix timestamp.
+  // to_unix_timestamp transforms a date into a unix timestamp.
   to_unix_timestamp(year, month, day, hour, minute, second)::
     years_seconds(year) + months_seconds(year, month) + days_seconds(day) + hour * 3600 + minute * 60 + second,
+
+  // parse_rfc3339 parses an RFC3339 timestamp in UTC (ending in 'Z') with date and time separated by 'T', like '2006-01-02T15:04:05Z'
+  // The returned object has "year", "month", "day", "hour", "minute" and "second" keys, as well as a to_unix_timestamp method that returns the unix timestamp of the parsed date.
+  parse_rfc3339(input)::
+    assert std.endsWith(input, 'Z') : 'the provided RFC3339 "%s" should end with "Z"' % input;
```
jdbaldry, line 76: According to the RFC, this can also be `z`

> NOTE: Per [ABNF] and ISO8601, the "T" and "Z" characters in this syntax may alternatively be lower case "t" or "z" respectively.
      
https://www.ietf.org/rfc/rfc3339.txt

However, it also says that it is valid to restrict the set to just the uppercase so it's up to you if you want to support both.

time/date.jsonnet
```diff
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
...
+    local datetime = std.split(std.substr(input, 0, std.length(input) - 1), 'T');
+    assert std.length(datetime) == 2 : 'the provided RFC3339 timestamp "%s" does not have date and time parts separated by the character "T"' % input;
```
jdbaldry, line 78: Similarly, this can be a `t`.

time/date.jsonnet
```diff
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
...
+  }
+);
+
 {
-  // date_to_unix_timestamp transforms a date into a unix timestamp.
+  // to_unix_timestamp transforms a date into a unix timestamp.
   to_unix_timestamp(year, month, day, hour, minute, second)::
     years_seconds(year) + months_seconds(year, month) + days_seconds(day) + hour * 3600 + minute * 60 + second,
+
+  // parse_rfc3339 parses an RFC3339 timestamp in UTC (ending in 'Z') with date and time separated by 'T', like '2006-01-02T15:04:05Z'
```
jdbaldry, line 73: I appreciate that a full RFC parser is perhaps more work than is worth right now to achieve the goals you have.

I wonder if we should name this `parse_rfc3339_utc` to reflect that it's a partial implementation of the full RFC?

I think it would also be useful to recognize that we are not supporting fractional seconds.

time/date.jsonnet
```diff
@@ -34,8 +34,52 @@ local months_seconds(year, month) = std.foldl(
 // days_seconds returns the number of seconds in all days up to day-1.
 local days_seconds(day) = (day - 1) * 24 * 3600;
 
+// is_numeric returns a bool indicating whether the input string is numeric
+local is_numeric(input) = std.length(input) > 0 && std.length(std.prune([
+  if std.codepoint('0') > std.codepoint(char) || std.codepoint(char) > std.codepoint('9') then char
+  for char in std.stringChars(input)
+])) == 0;
+
+// parse_date_or_time parses input which has part_names separated by sep.
```
jdbaldry, line 43: I think it would be useful to explain that the output is an object with `part_names` members.

jdbaldry (commented):
I bet this is gonna be super handy!

I have some thoughts that you can choose to ignore because the tests seem to show this is working just fine :)

Thinking about this at a high level, for parsing the full RFC spec in a functional way, I would consider a parser combinator (https://en.wikipedia.org/wiki/Parser_combinator). There is a library I know of (https://github.com/sbarzowski/jsonnet-parser-combinators) but we could also think of implementing our own.

malcolmholmes:
This might make more sense in https://github.com/jsonnet-libs/xtd?

time/date.jsonnet
colega: Oh right... functional programming. I guess I have to switch my mindset as I'm still doing imperative in jsonnet.

time/date.jsonnet
colega: What if I state the basic things I support now (no timezones, no sub-second, no lowercase `t` and `z`, and no space separator) and also state in a comment that nobody should rely on the fact that if fails parsing those, so we can extend it in the future?

If someone _needs_ those, they can implement them later by extending the implementation.

colega:
I think it makes sense for xtd, superseded by https://github.com/jsonnet-libs/xtd/pull/23 where I applied some of your comments, PTAL.

PR #1097, opened 2023-11-03
Fix Apache HBase alerts

Caleb-Hurshman:
This PR fixes a few minor issues with the alerts for the Apache HBase mixin.
1. Alert panel regex is updated to not include the instance label, which was preventing the panel from showing alerts properly.
2. Alert descriptions are changed to include `$value` rather than `$labels.value`, which was resolving incorrectly.

### DIFF
--- a/apache-hbase-mixin/alerts/alerts.libsonnet
+++ b/apache-hbase-mixin/alerts/alerts.libsonnet
@@ -17,7 +17,7 @@
               summary: 'There is a limited amount of heap memory available to the JVM.',
               description:
                 (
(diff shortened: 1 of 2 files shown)


Caleb-Hurshman:
@v-zhuravlev @mshahzeb I noticed that after installing the integration on k3d, the alert descriptions were modified from `{{printf "%%.0f" $value}}` to `{{printf "%.0f" $value}}`, which caused some bad output when the alert was firing. It seems like one of the percent signs is getting escaped, so is there a fix to get around this in whatever parser is getting used?

v-zhuravlev (approved):

PR #1098, opened 2023-11-06
Allow dashboards to show gRPC codes as labels

duricanikolic:
This PR updates all dashboards to allow reporting gRPC codes as `status_code`s. In order to achieve this, the pattern for recognizing valid status codes has been changed from `[a-z]+` to `[a-zA-Z]+`, since gRPC status codes contain upper cases letters too, e.g., `OK`, `FailedPrecondition`, `Canceled`, etc.

**Note**: The biggest change introduced by this PR is that if a component starts producing responses with gRPC codes only, successful `status_code` labels of the corresponding server and gRPC client request duration metrics will change from respectively `success` and `2xx` to `OK`.

### DIFF
--- a/grafana-builder/grafana.libsonnet
+++ b/grafana-builder/grafana.libsonnet
@@ -430,6 +430,7 @@
       '3xx': '#6ED0E0',
       '4xx': '#EF843C',
       '5xx': '#E24D42',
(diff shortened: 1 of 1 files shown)


dimitarvdimitrov (approved):
LGTM

PR #1105, opened 2023-11-21
Add Windows Active Directory Mixin

BominRahmani:
Adds dashboards and alerts for the Windows Active Directory mixin.

# Notes to reviewers: 
This was done using the new grafonnet library for the first time. There may be some weird things being done, I am more than happy to refactor if this is the case.

The dashboards are available in the cloud environment for review, You must select the datasource from the selectors, since this defaults it to prometheus. Same with the logs.

# Dashboards:
- Windows Active Directory overview
- Windows logs
image
image

# Alerts:
- WindowsActiveDirectoryHighPendingReplicationOperations: There is a high number of pending replication operations in Active Directory. A high number of pending operations sustained over a period of time can indicate a problem with replication.
- WindowsActiveDirectoryHighReplicationSyncRequestFailures: There are a number of replication synchronization request failures. These can cause authentication failures, outdated information being propagated across domain controllers, and potentially data loss or inconsistencies.'
- WindowsActiveDirectoryHighPasswordChange: There is a high number of password changes. This may indicate unauthorized changes or attacks.
- WindowsActiveDirectoryMetricsDown: Windows Active Directory metrics are down.


windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+The Windows Active Directory mixin is a set of configurable Grafana dashboards and alerts.
+
+The Windows Active Directory mixin contains the following dashboards:
+
+- Windows Active Directory overview
+- Windows Active Directory logs
+
+and the following alerts:
+- alertsCriticalHighReplicationIssues
```
StefanKurek, line 11: I believe these should be the names of the alerts and not the thresholds

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+              level: levelText
+        - labels:
+            source:
+            level:
+            job: "integrations/windows"
+```
+
+## Alerts overview
+
+- alertsCriticalHighReplicationIssues: There is a high number of replication issues. Replication issues can cause authentication failures, outdated information across domain controllers, and potentially data loss.
```
StefanKurek, line 64: again, should have the names of alerts and not thresholds

windows-active-directory-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,62 @@
...
+            alert: 'WindowsActiveDirectoryHighBindOperationsWarning',
+            expr: |||
+              sum without (bind_method, agent_hostname) (increase(windows_ad_binds_total[5m])) / clamp_min(sum without (bind_method, agent_hostname) (increase(windows_ad_binds_total[10m])), 1) * 100 > %(alertsWarningHighBindOperations)s
+            ||| % $._config,
+            'for': '5m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'There is a high percentage of bind operations. This may indicate brute force attacks or configuration issues.',
```
StefanKurek, line 17: The description for this alert doesn't really match up with reality. The description sounds like we're talking about something like too many bind operations compared to total operations.

Also, does this math make sense? it seems like if there is a single bind operation every minute, (or even every 5 minutes), we're going to end up with 50% which hits the threshhold to trigger the alert?

windows-active-directory-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,62 @@
+{
+  prometheusAlerts+:: {
```
StefanKurek, line 2: looks like you're missing your high replication issues alert

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+          target_label: 'agent_hostname'
+      pipeline_stages:
+        - json:
+            expressions:
+              source: source
+              level: levelText
+        - labels:
+            source:
+            level:
+            job: "integrations/windows"
```
StefanKurek, line 59: shouldn't we relabel the job to something like "integration/active_directory" or something like that to tell the difference between different windows jobs (like IIS)?

StefanKurek (changes requested):
A few issues. Haven't looked at the dashboard yet. Waiting until spec approval.

windows-active-directory-mixin/README.md
dpaasman00: ^^^

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+The Windows Active Directory overview dashboard provides details on integration status/alerts, workers/coordinators, error failures, data throughput, blocked nodes, and distributed bytes.
+![Windows Active Directory overview dashboard (LDAP)](https://storage.googleapis.com/grafanalabs-integration-assets/windows-active-directory/screenshots/windows_active_directory_overview_1.png)
+![Windows Active Directory overview dashboard (database)](https://storage.googleapis.com/grafanalabs-integration-assets/windows-active-directory/screenshots/window_active_directory_overview_2.png)
+
+# Windows Active Directory logs
+
+The Windows Active Directory logs dashboard provides details on incoming system logs.
+![Windows Active Directory logs dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/windows-active-directory/screenshots/windows_active_directory_logs.png)
+
+Windows Active Directory system logs are enabled by default in the `config.libsonnet` and can be removed by setting `enableLokiLogs` to `false`. Then run `make` again to regenerate the dashboard:
```
dpaasman00, line 27:
```suggestion
Windows Active Directory system logs are enabled by default in the `config.libsonnet` and can be disabled by setting `enableLokiLogs` to `false`. Then run `make` again to regenerate the dashboard:
```

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+## Windows Active Directory overview
+
+The Windows Active Directory overview dashboard provides details on integration status/alerts, workers/coordinators, error failures, data throughput, blocked nodes, and distributed bytes.
+![Windows Active Directory overview dashboard (LDAP)](https://storage.googleapis.com/grafanalabs-integration-assets/windows-active-directory/screenshots/windows_active_directory_overview_1.png)
+![Windows Active Directory overview dashboard (database)](https://storage.googleapis.com/grafanalabs-integration-assets/windows-active-directory/screenshots/window_active_directory_overview_2.png)
+
+# Windows Active Directory logs
+
+The Windows Active Directory logs dashboard provides details on incoming system logs.
```
dpaasman00, line 24:
```suggestion
The Windows Active Directory logs dashboard provides details on incoming AD logs.
```
"system" doesn't seem right unless they are actual system logs, so maybe "AD" or something else

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,111 @@
...
+
+```
+{
+  _config+:: {
+    enableLokiLogs: false,
+  },
+}
+```
+
+In order for the selectors to properly work for system logs ingested into your logs datasource, please also include the matching `instance` and `job` labels onto the [scrape configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
```
dpaasman00, line 37:
```suggestion
For the selectors to properly work with AD logs ingested into your logs datasource, please also include `instance` and `job` labels in the [scrape configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) so they match the labels for ingested metrics.
```

dpaasman00 (commented):
Few comments for the moment

windows-active-directory-mixin/README.md
StefanKurek: that's interesting. because for IIS i thought we were changing the label to NOT be the generic "integrations/windows" so that we could filter by job and only have the IIS metrics. either way, if vitaly suggested that, then i'm ok with it as well

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,159 @@
+# Windows Active Directory mixin
+
+The Windows Active Directory mixin is a set of configurable Grafana dashboards and alerts.
+
+The Windows Active Directory mixin contains the following dashboards:
+
```
StefanKurek, line 6: nitpicky: but do we normally have a space here but not before the alerts?

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,159 @@
...
+  _config+:: {
+    enableLokiLogs: false,
+  },
+}
+```
+
+For the selectors to properly work with Windows Active Directory event logs ingested into your logs datasource, please also include the matching `instance` and `job` labels in the [scrape configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
+
+```yaml
+- job_name: integrations/windows-exporter-application
```
StefanKurek, line 40: is this spacing incorrect (comparing it to the others)?

windows-observ-lib/targets.libsonnet
```diff
@@ -335,5 +334,90 @@ local lokiQuery = g.query.loki;
         'irate(windows_net_packets_sent_total{%(queriesSelector)s}[$__rate_interval])' % variables
       )
       + prometheusQuery.withLegendFormat('{{ nic }} transmitted'),
+
+    //alertsPanel:
```
StefanKurek, line 338: should this be deleted?

windows-observ-lib/dashboards.libsonnet
```diff
@@ -28,6 +28,7 @@ local logslib = import 'github.com/grafana/jsonnet-libs/logs-lib/logs/main.libso
               panels.diskIOutilPercentTopK { gridPos+: { w: 12 } },
               panels.diskUsagePercentTopK { gridPos+: { w: 12 } },
               panels.networkErrorsAndDroppedPerSecTopK { gridPos+: { w: 24 } },
+              panels.replicationPendingOperations { gridPos+: { w: 6, h: 3 } },
```
StefanKurek, line 31: so what exactly is this and why is this panel special (why here for this one only?)

windows-observ-lib/panels.libsonnet
```diff
@@ -16,45 +16,45 @@ local utils = commonlib.utils;
             t.osInfo
             + g.query.prometheus.withFormat('table')
             + g.query.prometheus.withInstant(true)
-            + g.query.prometheus.withRefId('OS Info'),
+            + g.query.prometheus.withRefId('os info'),
```
StefanKurek, line 19: just curious why all these lowercase changes?

windows-observ-lib/panels.libsonnet
```diff
@@ -449,5 +449,331 @@ local utils = commonlib.utils;
                               targets=[t.networkInPacketsPerSec, t.networkOutPacketsPerSec]
                             )
                             + commonlib.panels.network.timeSeries.traffic.withNegateOutPackets(),
+
+      //	alertsPanel: g.panel.alertList.new(
```
StefanKurek, line 453: why are these commented out?

windows-observ-lib/panels.libsonnet
```diff
@@ -449,5 +449,331 @@ local utils = commonlib.utils;
...
+      //   Alerts that are actively firing in Windows Active Directory environment.
+      // |||
+      //),
+
+      replicationPendingOperations: commonlib.panels.generic.stat.info.new(
+        'Replication pending operations',
+        targets=[t.replicationPendingOperations],
+        description=|||
+          The number of replication operations that are pending in Active Directory.
+          These operations could include a variety of tasks, such as updating directory objects, processing changes made on other domain controllers, or applying new schema updates.
```
StefanKurek, line 466: any specific reason for these really detailed descriptions? i like them, but just want to check the reasoning

windows-observ-lib/dashboards.libsonnet
BominRahmani: This was when i was testing something out and forgot to get rid of it! Good catch!

windows-observ-lib/panels.libsonnet
BominRahmani: Not sure how this happened. It is unintended.

windows-observ-lib/panels.libsonnet
BominRahmani: This was my attempt at getting the alerts panel working previous to knowing how, since there is no alerts panel in the common lib. Had this commented out till i had more direction from Vitaly, Got it figured out now!

windows-active-directory-mixin/README.md
BominRahmani: I dont think it matters as the spacing changes depending on the mixin you are looking at, but i updated it to be more inline with the style you are thinking of i believe.

windows-observ-lib/panels.libsonnet
BominRahmani: I guess the reasoning is kinda circular, its more descriptive because i wanted them to be more descriptive?  On deeper thought its probably because these replication pending operations can include a variety of different things and I just wanted to include that into the description to more or less encapsulate that thought.

windows-active-directory-mixin/README.md
```diff
@@ -0,0 +1,154 @@
...
+{
+  _config+:: {
+    enableLokiLogs: false,
+  },
+}
+```
+
+For the selectors to properly work with the Windows logs ingested into your logs datasource, please also include the matching `instance` and `job` labels in the [scrape configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) to match the labels for ingested metrics.
+
+```yaml
```
dpaasman00, line 37: Is this yaml section new? I don't remember it being in the README before

windows-active-directory-mixin/alerts/alerts.libsonnet
```diff
@@ -0,0 +1,79 @@
...
+            alert: 'WindowsActiveDirectoryHighPendingReplicationOperations',
+            expr: |||
+              windows_ad_replication_pending_operations >= %(alertsHighPendingReplicationOperations)s 
+            ||| % $._config,
+            'for': '10m',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'There is a high number of pending replication operations in Active Directory. A high number of pending operations sustained over a period of time can indicate a problem with replication.',
```
dpaasman00, line 17: I feel the second sentence being here defeats the purpose of this summary section. As far as I know and how I've treated it, this is just a quick blurb about the alert, so I feel the first sentence is enough for the summary. I think the second sentence would fit better in the description. Ultimately your call though if you've gotten review in the past that says otherwise. 

This comment extends to HighReplicationSyncRequestFailures and HighPasswordChanges

windows-active-directory-mixin/README.md
StefanKurek: @BominRahmani looks like you didn't respond to this one. i think the job name should be moved over to be in line with `windows_events`?

windows-observ-lib/panels.libsonnet
```diff
@@ -449,5 +449,327 @@ local utils = commonlib.utils;
...
+                                      includeTimeField: false,
+                                      mode: 'reduceFields',
+                                      reducers: [
+                                        'last',
+                                      ],
+                                    },
+                                  },
+                                ]
+                              ),
+      intrasiteReplicationTraffic: commonlib.panels.generic.timeSeries.base.new(
```
dpaasman00, line 632: For this panel and the intersite version, do they need the min, max, mean legend calcs?

windows-observ-lib/panels.libsonnet
```diff
@@ -40,7 +40,7 @@ local utils = commonlib.utils;
             t.diskTotalC
             + g.query.prometheus.withFormat('table')
             + g.query.prometheus.withInstant(true)
-            + g.query.prometheus.withRefId('Disk C: total'),
+            + g.query.prometheus.withRefId('disk c: total'),
```
StefanKurek, line 43: @BominRahmani looks like you forgot one?

windows-observ-lib/panels.libsonnet
dpaasman00: Same for inboundReplicationUpdates

dpaasman00 (commented):
I think this looks good for the most part. I don't think I can give a proper review on the new library stuff, but what you have makes sense to me and the dashboard on the cloud seems fine to me. Just a few comments

windows-active-directory-mixin/README.md
BominRahmani: This yaml section is just to show what the log snippet would look like for the mixin. Its not new as far as I know and I have been asked to include it in the past.

windows-observ-lib/panels.libsonnet
BominRahmani: I added them in for now, I know previously I did not have them in, but I want to see what Grafana will say to this. The reason for the switch is I think having a calculated mean would be really helpful, and as far as i've seen whenever there is a calculated mean, there is always a min and max that follows it.

windows-active-directory-mixin/alerts/alerts.libsonnet
BominRahmani: Im not sure about this, I have always seen the description as being a flat out written form of where the thresholds have been crossed at what instances. The summary to me has always been why whatever alert has been triggered is an actual problem. I think I will keep it as is for now.

dpaasman00: That's fine, just something to keep in mind cus in the UI the summary doesn't have a lot of space, if you check out some of the pre-existing alerts on the cloud instance

windows-observ-lib/panels.libsonnet
dpaasman00: Fair enough to see what they say, but at that rate they may prefer a stat panel instead, based on feedback I got during spec review at one point when I had a similar situation ... just something to keep in mind

dpaasman00 (approved):

StefanKurek (approved):

gaantunes:
Both dashboards are loading wrong default datasources, causing errors when we open them.

The logs dashboard does not show any data, tried 30 days back. Overall the overview dashboard looks good, would just like to see more data flowing if that is possible.

Screenshot from 2023-12-15 18-39-28
image

BominRahmani:
> Both dashboards are loading wrong default datasources, causing errors when we open them.
> 
> The logs dashboard does not show any data, tried 30 days back. Overall the overview dashboard looks good, would just like to see more data flowing if that is possible.
> 
> Screenshot from 2023-12-15 18-39-28 image

Hey @gaantunes Not quite sure why the logs weren't showing up in that time range, i reset the VM and the logs are showing now so that should be good. As for the errors due to the default datasources, wouldn't this resolve itself on the integration itself similar to how the Log library resolves itself in the integration step? Maybe @v-zhuravlev you have some insights into this?

windows-observ-lib/panels.libsonnet
v-zhuravlev, line 632: Missing unit (bps?).

Tip: Since it is a network traffic of specific type, I think you can try to use `commonlib.panels.network.timeSeries.traffic` as a closer prototype here.

windows-observ-lib/panels.libsonnet
```diff
@@ -449,5 +449,327 @@ local utils = commonlib.utils;
...
+                                     |||
+                                   )
+                                   + g.panel.timeSeries.options.legend.withDisplayMode('table')
+                                   + g.panel.timeSeries.options.legend.withPlacement('bottom')
+                                   + g.panel.timeSeries.options.legend.withCalcs([
+                                     'min',
+                                     'max',
+                                     'mean',
+                                   ]),
+      intersiteReplicationTraffic: commonlib.panels.generic.timeSeries.base.new(
```
v-zhuravlev, line 646: missing unit.

windows-observ-lib/targets.libsonnet
```diff
@@ -335,5 +334,84 @@ local lokiQuery = g.query.loki;
...
+        '${' + variables.datasources.prometheus.name + '}',
+        'sum(windows_ad_replication_pending_synchronizations{%(queriesSelector)s})' % variables
+      )
+      + prometheusQuery.withLegendFormat('Operations'),
+    ldapBindRequests:
+      prometheusQuery.new(
+        '${' + variables.datasources.prometheus.name + '}',
+        'rate(windows_ad_binds_total{bind_method=~"ldap", %(queriesSelector)s}[$__rate_interval])' % variables
+      )
+      + prometheusQuery.withLegendFormat('{{ instance }}'),
```
v-zhuravlev, line 361: Perhaps we can be more generic here, so legend is transformed if something else, other than instance is used:

```suggestion
      + prometheusQuery.withLegendFormat(commonlib.utils.labelsToPanelLegend(this.config.instanceLabels)),
```

Please also add import to the top of the file:
`local commonlib = import 'common-lib/common/main.libsonnet';`

windows-observ-lib/targets.libsonnet
```diff
@@ -335,5 +334,84 @@ local lokiQuery = g.query.loki;
...
+
+    ldapOperations:
+      prometheusQuery.new(
+        '${' + variables.datasources.prometheus.name + '}',
+        'rate(windows_ad_directory_operations_total{origin=~"ldap", %(queriesSelector)s}[$__rate_interval])' % variables
+      )
+      + prometheusQuery.withLegendFormat('{{ instance }} - {{ operation }}'),
```
v-zhuravlev, line 368:
```suggestion
      + prometheusQuery.withLegendFormat('%s - {{ operation }}' % commonlib.utils.labelsToPanelLegend(this.config.instanceLabels)),
```
same as above

windows-observ-lib/targets.libsonnet
```diff
@@ -335,5 +334,84 @@ local lokiQuery = g.query.loki;
...
+      prometheusQuery.new(
+        '${' + variables.datasources.prometheus.name + '}',
+        'rate(windows_ad_binds_total{%(queriesSelector)s}[$__rate_interval])' % variables
+      )
+      + prometheusQuery.withLegendFormat('{{ instance }} - {{ operation }}'),
+
+    intrasiteReplicationTraffic:
+      prometheusQuery.new(
+        '${' + variables.datasources.prometheus.name + '}',
+        'rate(windows_ad_replication_data_intrasite_bytes_total{%(queriesSelector)s}[$__rate_interval])' % variables
```
v-zhuravlev, line 380: You can add * 8 here, or change units from bps to Bps on the panel level instead.

```suggestion
        'rate(windows_ad_replication_data_intrasite_bytes_total{%(queriesSelector)s}[$__rate_interval]) * 8' % variables
```

v-zhuravlev (commented):
Great work. Thanks for using modular libs approach.

Overall, I suggest to put not only panels/dashboards/targets into windows-observ-lib but also alerts group and config params.

That would leave folder `windows-active-directory-mixin` here as pure monitoring-mixin example, and nothing more.

When creating integration, we can import windows-observ-lib, and just set things like uid, filteringSelector, pick specific dashboards, etc... to generated required objects for AD integration.

windows-observ-lib/panels.libsonnet
```diff
@@ -449,5 +449,327 @@ local utils = commonlib.utils;
                               targets=[t.networkInPacketsPerSec, t.networkOutPacketsPerSec]
                             )
                             + commonlib.panels.network.timeSeries.traffic.withNegateOutPackets(),
+
+      alertsPanel: g.panel.alertList.new(
+                     'Windows Active Directory alerts'
+                   )
+                   + g.panel.alertList.options.UnifiedAlertListOptions.withAlertInstanceLabelFilter('job=~"${job:regex}", instance=~"${instance:regex}"'),
```
v-zhuravlev, line 456: Could you try to use `queriesSelector` or at least this.config.instanceLabels, this.config.groupLabels to make sure that filter is changed accordingly?

One idea I have is to go to file variables.libsonnet and generate new field there called `queriesSelectorAdvancedSyntax` that will be just like `queriesSelector` but with ${<label>:regex} . Then we can use it here.

BominRahmani: Due to the way labelsToPromQLSelector works in commonlib,
image
doing it this way would cause the output to look like this
`job=~"integrations/windows_exporter",job:regex=~"$job:regex",instance:regex=~"$instance:regex"`
Since it reuses the label for the label and variable it will cause both parts to receive the regex modification that I add into the instance labels. Also the labelsToPromQLSelector function doesn't account for the `{}` encapsulation of the variable, eg: advanced grafana syntax should look like this: `'${job:regex}'`
Because of this i extended the commonlibs util function and added an advanced version of the labelsToPromQLSelector. This should take care of this, but I think it would be worth adding this function into commonlibs itself since alert panels are a regular occurence.

windows-active-directory-mixin/mixin.libsonnet
```diff
@@ -0,0 +1,49 @@
...
+  {
+    grafana+: {
+      local link = g.dashboard.link,
+      links: {
+        backToLogs:
+          link.link.new('Windows logs', '/d/activedirectory-logs')
+          + link.link.options.withKeepTime(true),
+        backToActiveDirectoryOverview:
+          link.link.new('Active Directory overview', '/d/activedirectory')
+          + link.link.options.withKeepTime(true),
```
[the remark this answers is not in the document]
BominRahmani: Sorry I am not quite sure what you mean here, could you give an example?

BominRahmani:
> Great work. Thanks for using modular libs approach.
> 
> Overall, I suggest to put not only panels/dashboards/targets into windows-observ-lib but also alerts group and config params.
> 
> That would leave folder `windows-active-directory-mixin` here as pure monitoring-mixin example, and nothing more.
> 
> When creating integration, we can import windows-observ-lib, and just set things like uid, filteringSelector, pick specific dashboards, etc... to generated required objects for AD integration.

I moved the alerts and configs into windows observ library, but I had some questions, I left the config option of enablingADDashboard on the mixin just to enable it since its defaulted to false in the library, im guessing this is ok? Also since the alerts are sitting within the observ library, wouldnt the windows exporter integration automatically pull all these alerts even though they may not have AD enabled?

windows-active-directory-mixin/mixin.libsonnet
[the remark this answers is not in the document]
v-zhuravlev: I think that those exrtra dashobard links are redundant - they are useful in three tier view, but here we only have two dashboards.

BominRahmani:
> > > Great work. Thanks for using modular libs approach.
> > > Overall, I suggest to put not only panels/dashboards/targets into windows-observ-lib but also alerts group and config params.
> > > That would leave folder `windows-active-directory-mixin` here as pure monitoring-mixin example, and nothing more.
> > > When creating integration, we can import windows-observ-lib, and just set things like uid, filteringSelector, pick specific dashboards, etc... to generated required objects for AD integration.
> > 
> > 
> > I moved the alerts and configs into windows observ library, but I had some questions, I left the config option of enablingADDashboard on the mixin just to enable it since its defaulted to false in the library, im guessing this is ok? Also since the alerts are sitting within the observ library, wouldnt the windows exporter integration automatically pull all these alerts even though they may not have AD enabled?
> 
> yes, we can hide alerts there under AD feature flag as well.

Sounds good, I just added a check in the alerts section of windowsobservlib which checks whether that feature flag is enabled. I think this is what you had in mind yes? @v-zhuravlev

windows-mixin/config.libsonnet
```diff
@@ -19,6 +19,7 @@
     alertMemoryUsageThresholdCritical: '90',
     alertDiskUsageThresholdCritical: '90',
     // set to false to disable logs dashboard and logs annotations
+    enableADDashboard: false,
```
v-zhuravlev, line 22:
```suggestion
    enableADDashboard: false,
    // set to false to disable logs dashboard and logs annotations
```

v-zhuravlev (approved):

PR #1107, opened 2023-11-28
Update CPU modes panel

v-zhuravlev:
Add min=0,max=100 to see what mode is most used:
image
->
Screenshot 2023-11-28 at 22 49 28


gloriasc (approved):
Looks much better! Thanks for the update!

PR #1108, opened 2023-11-29
Update hello-world

v-zhuravlev:
Add more examples
Move config into separate file as a result of node-observ-lib feedback.

Screens from grr apply mixin.libsonnet:

image
image
image


Dasomeone (approved):
Looks great, thanks for making this change! 😄

PR #1121, opened 2024-01-09
wip: support native histograms along classic histograms in panels

krajorama:
This branch will have force pushes.

Related to https://github.com/grafana/mimir/issues/7154

Native histograms are
https://grafana.com/docs/mimir/latest/send/native-histograms/


mixin-utils/utils.libsonnet
```diff
@@ -1,6 +1,38 @@
 local g = import 'grafana-builder/grafana.libsonnet';
 
 {
+  // The classicNativeHistogramQuantile function is used to calculate histogram quantiles from native histograms or classic histograms.
+  // Metric name should be provided without _bucket suffix.
+  nativeClassicHistogramQuantile(percentile, metric, selector, sum_by=[], rate_interval='$__rate_interval')::
+    local classicSumBy = if std.length(sum_by) > 0 then ' by (%(lbls)s) ' % { lbls: std.join(',', ['le'] + sum_by) } else ' by (le) ';
+    local nativeSumBy = if std.length(sum_by) > 0 then ' by (%(lbls)s) ' % { lbls: std.join(',', sum_by) } else ' ';
+    'histogram_quantile(%(percentile)s, sum%(nativeSumBy)s(rate(%(metric)s{%(selector)s}[%(rateInterval)s]))) or histogram_quantile(%(percentile)s, sum%(classicSumBy)s(rate(%(metric)s_bucket{%(selector)s}[%(rateInterval)s])))' % {
```
krajorama, line 9: for reviewers: e.g.:
```
(histogram_quantile(0.99, sum (rate(cortex_request_duration_seconds{}[$__rate_interval]))) or
histogram_quantile(0.99, sum by (le) (rate(cortex_request_duration_seconds_bucket{}[$__rate_interval]))))
```

mixin-utils/utils.libsonnet
```diff
@@ -1,6 +1,38 @@
...
+      nativeSumBy: nativeSumBy,
+      percentile: percentile,
+      rateInterval: rate_interval,
+      selector: selector,
+    },
+
+  // The classicNativeHistogramSumRate function is used to calculate the histogram sum of rate from native histograms or classic histograms.
+  // Metric name should be provided without _sum suffix.
+  nativeClassicHistogramSumRate(metric, selector, rate_interval='$__rate_interval')::
+    'histogram_sum(rate(%(metric)s{%(selector)s}[%(rateInterval)s])) or rate(%(metric)s_sum{%(selector)s}[%(rateInterval)s])' % {
```
krajorama, line 21: for reviewers e.g., in context of a sum: 
```
sum(
 histogram_sum(rate(cortex_request_duration_seconds{}[$__rate_interval])) or
 rate(cortex_request_duration_seconds_sum{}[$__rate_interval])
)
```

mixin-utils/utils.libsonnet
beorn7: I think this can be improved.

Imagine you have a fairly long `rate_interval` (the problem exists for all intervals, but the longer, the more serious it gets), e.g. multiple days or so. While migrating to native histograms, you ingest classic and native histograms in parallel. Very soon, the first leg of the query above will yield a result, but it will be based just on a few samples of native histograms, or the last few minutes of the multi-day range.

So what you want is to use the leg of the query that has a complete coverage of the `rate_interval` (and in doubt prefer the native histogram). I'll try to come up with a way to do that in the next comment.

beorn7: Maybe this could work:

```PromQL
(
      histogram_quantile(
        0.99,
        sum(rate(cortex_request_duration_seconds[1w])) and group(cortex_request_duration_seconds offset 1w)
      )
    or
      histogram_quantile(
        0.99,
          sum by (le) (rate(cortex_request_duration_seconds_bucket[1w]))
        and
          group by (le) (cortex_request_duration_seconds_bucket)
      )
  or
    histogram_quantile(0.99, sum(rate(cortex_request_duration_seconds[1w])))
)
```

This assumes the migration is already going from classic to native histograms. First, the query checks if there has been a native histogram 1w ago. If so, it goes for native histograms. Second, it checks if there is a classic histogram now, and uses them. As the third and final option, it uses native histograms (which is relevant for a new metric that never had classic histograms).

Of course, the stupid PromQL engine will calculate all three legs, even if only one will ever be used.

krajorama:
Superseded by #1150

PR #1122, opened 2024-01-10
Add modules and auths key in snmp yml

mshahzeb:
Fixes 
https://github.com/grafana/support-escalations/issues/8785

The default yml is referenced in the Ubiquiti edge router's integration instructions.
But the yml is in incorrect format and causes the agent to crash and not startup.

This PR adds the required keys in the yml file to fix the issue.


rgeyer (approved):
This looks good to me.

I spent an unnecessarily long time looking at previous commits because.. I recall that we changed this already.

I know that the agent was updated to a [new version of the SNMP exporter](https://github.com/grafana/agent/pull/4265), which necessitated some changes to our config, which I was sure we (I?) did, but I can't find record of it.

Perhaps it was on the now defunct `ubnt` integration.

In any case, this looks good! Thanks for cleaning it up!

PR #1123, opened 2024-01-11
Remove deprecated angular table panel from SNMP mixin

mshahzeb:
Updates out-of-date angular table panel to new table panel.

Since the old grafonnet version does not have a new table panel - the dashboard has been updated in Grafana and saved as JSON.

image


snmp-mixin/dashboards/snmp-overview.json
```diff
@@ -0,0 +1,1078 @@
...
+      },
+      "id": 2,
+      "panels": [],
+      "showTitle": true,
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "grafanacloud-prom"
+          },
```
Dasomeone, line 58: Minor nitpick, but pretty sure these are all meant to be no type and `"uid": "$datasource"`

snmp-mixin/dashboards/snmp-overview.json
```diff
@@ -0,0 +1,1078 @@
...
+    "list": [
+      {
+        "current": {
+          "selected": false,
+          "text": "default",
+          "value": "default"
+        },
+        "hide": 0,
+        "includeAll": false,
+        "label": "Data Source",
```
Dasomeone, line 934: This won't pass linting, mixtool is getting quite particular on naming (e.g. should be `Data source`) Might be worth giving it a pass to see if there's anything else

Dasomeone (commented):
Couple minor comments but otherwise looks great

snmp-mixin/dashboards/snmp-overview.json
mshahzeb: Ran mixtool lint and it did not raise any issues.

Dasomeone: Interesting, then one of us has outdated mixtool :sweat_smile: 

Will update mine and test

snmp-mixin/dashboards/snmp-overview.json
mshahzeb: This is actually a row panel and I dont think it even needs targets - removed it. Thanks for the catch!

snmp-mixin/dashboards/snmp-overview.json
Dasomeone: Update: turns out it was my mixtool, so all good to go on this one :)

Dasomeone (approved):
LGTM, thanks Shahzeb :D

rgeyer:
> Is there a replacement for the deleted `snmp-mixin/lib/utils.libsonnet` in this PR? For now I've copied the function.

The goal of this PR was to move away from using deprecated panels, which includes moving from graph to timeseries. That util function was a bit of a hack to patch that change in.

As such, now that the dashboard uses timeseries natively, the util isn't needed anymore (for this mixin). Were you using the function in some other project?

zalegrala:
I was using the util for something else, but perhaps I need to migrate that as well.  In the meantime, the copy works fine for me while I figure out what direction to take the jsonnet dashbaords I've got.

PR #1129, opened 2024-01-18
Update Kafka mixin to use React based panels

Dasomeone:
This PR introduces a number of changes, though primarily targeted towards the replacement of Angular panels with React panels.
Specifically:
- All `graph` panels have been updated to `timeseries` panels
- All `table-old` panels have been updated to `table` panels

This has been done by flattening dashboards to JSON format, given the urgency of this change as Angular support will be dropped in Grafana soon.

The old libsonnet dashboards have been preserved in a new subfolder.


mshahzeb (approved):
This was a complex one! Thanks for tackling this.

LGTM

PR #1132, opened 2024-01-22
Add wildcard allValue option to mixin cluster label

v-zhuravlev:
To make sure mixin(and cloud integraiton) returns data even if 'cluster' label is not used.

Ref: 
https://community.grafana.com/t/memcached-integration-0-0-5-dashboards-appear-to-require-cluster-label/111479


gloriasc (approved):
Awesome! Thanks!

PR #1135, opened 2024-01-31
Commonlib signal

v-zhuravlev:
This part of the lib allows to define signals (i.e. metrics, logs). 

This shifts focus of dashboard building from panels to actual signals we want to observe.

Workflow to generate dashboards would be as follows:
- Define key indicators(signals) your want to observe.
- Use built-in functions to render each signal as a panel (i.e. signal.asTimeSeries())
- If you want to combine multiple signals inside a single panel, render signal not as [Grafana panel](https://grafana.github.io/grafonnet/API/panel/index.html), but as a [Target](https://grafana.github.io/grafonnet/API/panel/timeSeries/index.html#fn-queryoptionswithtargets), which is attachable to existing panel
- Bonus: use stylize() functions from commonlib/panels to apply common styles to signals.

See README for more.

Currently, only astimeSeries() for typical metric types as gauge, counter and histogram is implemented. Other panels rendering will be added later.


common-lib/common/signal/README.md
```diff
@@ -0,0 +1,69 @@
...
+        description='Bytes out through interface',
+        expr='node_network_transmit_bytes_total{%(queriesSelector)s}',
+    ),
+};
+
+
+
+#generate dashboard
+g.dashboard.new('Device')
++ g.dashboard.withVariables(<somevars>)
```
rgeyer, line 49: I'm not sure it's a good idea, so consider this a question, or a thought experiment.

Since signals end up defining the templating variables that they expect, could the base class/object emit a list of template variables, perhaps in a map with the label name as the key to reduce/eliminate duplication, that could be aggregated/combined into the list of variables for the dashboard?

Might not be tremendously useful ultimately, but I do worry that we're making it *very* easy to make fairly complex queries that depend upon template variables being set, and if they're absent it may be difficult to troubleshoot.

rgeyer (approved):
This looks great, and I am in full support of the idea of focusing on signals, and allowing them to define the rest of the visualizations.

I did offer one question/thought, but I'm not convinced any action is required, perhaps just a discussion.

common-lib/common/signal/README.md
v-zhuravlev: Good idea. I think it would fairly easy to render variables for dashboard that signals are expecting. I'll add it!

v-zhuravlev: added.

common-lib/common/signal/signal.libsonnet
```diff
@@ -0,0 +1,133 @@
...
+    datasource:: datasource,
+    aggLevel:: aggLevel,
+
+    // create vars
+    local grafanaVariables = variables.new(
+      filteringSelector[0],
+      groupLabels,
+      instanceLabels,
+      varMetric=varMetric,
+    ),
```
rgeyer, line 34: This is great! Thank you for implementing this idea!

rgeyer (approved):
LGTM! Great work!

PR #1139, opened 2024-02-05
modify withRunbookURL to allow internal annotation

tristanburgess:
see https://github.com/grafana/deployment_tools/pull/118284

this PR allows a user of the `withRunbookURL` function to indicate that the `internal_runbook_url` annotation should be used instead of the `runbook_url` annotation.

### DIFF
--- a/mixin-utils/utils.libsonnet
+++ b/mixin-utils/utils.libsonnet
@@ -117,12 +117,13 @@ local g = import 'grafana-builder/grafana.libsonnet';
   // withRunbookURL - Add/Override the runbook_url annotations for all alerts inside a list of rule groups.
   // - url_format: an URL format for the runbook, the alert name will be substituted in the URL.
   // - groups: the list of rule groups containing alerts.
-  withRunbookURL(url_format, groups, internal=false)::
+  // - annotation_key: the key to use for the annotation whose value will be the formatted runbook URL.
+  withRunbookURL(url_format, groups, annotation_key='runbook_url')::
     local update_rule(rule) =
       if std.objectHas(rule, 'alert')
       then rule {
         annotations+: {
-          [if !internal then 'runbook_url' else 'internal_runbook_url']: url_format % rule.alert,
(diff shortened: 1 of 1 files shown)


mixin-utils/utils.libsonnet
```diff
@@ -117,12 +117,12 @@ local g = import 'grafana-builder/grafana.libsonnet';
   // withRunbookURL - Add/Override the runbook_url annotations for all alerts inside a list of rule groups.
   // - url_format: an URL format for the runbook, the alert name will be substituted in the URL.
   // - groups: the list of rule groups containing alerts.
-  withRunbookURL(url_format, groups)::
+  withRunbookURL(url_format, groups, internal=false)::
```
Duologic, line 120: Adding conditional like this is a bit of anti pattern that leads to overloading a function signature to satisfy every user.

I'd suggest giving the original default like this, it'll allow the same override but needs less maintenance if someone follows a slightly alternative annotation approach.
```suggestion
  withRunbookURL(url_format, groups, annotation_key='runbook_url')::
```

Perhaps also add a line of documentation above this function to explain it.

Duologic (changes requested):

mixin-utils/utils.libsonnet
tristanburgess: updated and tested

```
tk diff ksonnet/environments/ruler-exporter/main.jsonnet --name environments/ruler-exporter/dev-us-central-0

...
     ---
-    alert: UnhealthyComponents
+    alert: "UnhealthyComponents"
     annotations:
-      message: Unhealthy Flow components detected.
+      custom_runbook_annotation_key: "https://github.com/grafana/cloud-onboarding/blob/main/runbooks/HostedExporters-Agent.md#AgentUnhealthyComponents"
+      message: "Unhealthy Flow components detected."
     expr: "sum(agent_component_controller_running_components{health_type!=\"healthy\"}) > 0"
-    for: 15m
+    for: "15m"
```

Duologic (approved):

PR #1144, opened 2024-02-15
Istio - Adds second version of mixin

StefanKurek:
# Overview
There is currently an existing mixin for Istio. I wasn't sure if it would be best to make an entirely new mixin, or to try and "overwrite" that mixin as it is so different. I ended up taking the first approach.

Here is an [overview](https://istio.io/latest/docs/examples/bookinfo/) of the bookinfo services shown in the dashboards. 
Also, the `sleep` service sends TCP requests to the `tcp-echo` service, and GRPC requests to the `echo` service.

There are two types of logs shown in the logs dashboard. Access logs ([which are not in a super user friendly form](https://istio.io/latest/docs/tasks/observability/logs/access-log/#default-access-log-format)) and system logs.  I did at one point try to parse all of the access log info into labels, but the pipeline component seemed to completely break (these logs stopped being exported to Grafana Cloud) after I reached a certain limit of regex capture groups and labels (somewhere around 8).

# Grafana Cloud Dashboards
[Istio overview dashboard](https://keithschmitty.grafana.net/d/istio-overview/istio-overview-mixin?orgId=1&from=1706115000000&to=1706119199000)

[Istio services overview dashboard](https://keithschmitty.grafana.net/d/istio-services-overview/istio-services-overview-mixin?from=1706115000000&to=1706119199000&orgId=1&var-datasource=grafanacloud-prom&var-job=All&var-cluster=All&var-namespace=All&var-service=details&var-service=istio-ingress&var-service=productpage&var-service=sleep&var-service=tcp-echo&var-service=echo&var-client_service=All&var-server_service=All)

[Istio workloads overview dashboard](https://keithschmitty.grafana.net/d/istio-workloads-overview/istio-workloads-overview-mixin?orgId=1&from=1706115000000&to=1706119199000&var-datasource=grafanacloud-prom&var-job=All&var-cluster=All&var-namespace=All&var-service=tcp-echo&var-service=sleep&var-service=echo&var-service=productpage&var-workload=All&var-client_workload=All&var-server_workload=All)

[Istio logs dashboard](https://keithschmitty.grafana.net/d/istio-logs/istio-logs?orgId=1&refresh=1m&var-loki_datasource=grafanacloud-logs&var-job=All&var-cluster=All&var-pod=All&var-log_type=All&var-protocol=All&var-request_method=All&var-response_code=All&var-level=All&var-regex_search=&var-datasource=grafanacloud-prom)


istio-2-mixin/links.libsonnet
```diff
@@ -0,0 +1,21 @@
+local g = import './g.libsonnet';
```
BominRahmani, line 1: Looks like the links default to "Back to ..." I think we should keep this as just the name of the dashboard. Traditionally its what been done. The windows exporter linking system is a bit wonky since its focused a lot more on a drill down aspect, which is why you see those back to verbiage.

istio-2-mixin/links.libsonnet
```diff
@@ -0,0 +1,21 @@
...
+      backToOverview:
+        link.link.new('Back to Istio overview', '/d/' + this.grafana.dashboards.overview.uid)
+        + link.link.options.withKeepTime(true),
+      backToServicesOverview:
+        link.link.new('Back to Istio services overview', '/d/' + this.grafana.dashboards.servicesOverview.uid)
+        + link.link.options.withKeepTime(true),
+      backToWorkloadsOverview:
+        link.link.new('Back to Istio workloads overview', '/d/' + this.grafana.dashboards.workloadsOverview.uid)
+        + link.link.options.withKeepTime(true),
+      otherDashboards:
```
BominRahmani, line 15: You probably don't need an all dashboards link, I would replace this with the logs link.

istio-2-mixin/config.libsonnet
```diff
@@ -0,0 +1,33 @@
+{
+  // Static selector to apply to ALL dashboard variables of type query, panel queries, alerts and recording rules.
+  filteringSelector: 'job=~"integrations/istio"',
+  // Used to identify 'group' of instances.
+  groupLabels: ['job', 'cluster'],
+  instanceLabels: ['instance'],
```
BominRahmani, line 6: I remember you mentioning Istio has no such thing as "instances" the way we use them. It looks like this label only gets used one more time in the logs? Can we get rid of it?

BominRahmani (changes requested):
This mixin looks awesome! Just some minor comments. Let me know whenever you get a chance to take a look at them!

istio-2-mixin/README.md
```diff
@@ -0,0 +1,251 @@
...
+![Istio overview dashboard (control plane)](https://storage.googleapis.com/grafanalabs-integration-assets/istio/screenshots/istio_overview_2.png)
+
+## Istio services overview
+The Istio services overview dashboard provides details on HTTP/GRPC throughput, HTTP/GRPC response time, TCP throughput, and workload info for services acting in both client and server roles.
+![Istio services overview dashboard (client)](https://storage.googleapis.com/grafanalabs-integration-assets/istio/screenshots/istio_services_overview_1.png)
+![Istio services overview dashboard (server)](https://storage.googleapis.com/grafanalabs-integration-assets/istio/screenshots/istio_services_overview_2.png)
+![Istio services overview dashboard (workloads)](https://storage.googleapis.com/grafanalabs-integration-assets/istio/screenshots/istio_services_overview_3.png)
+
+## Istio workloads overview
+The Istio overview dashboard provides details on HTTP/GRPC throughput, HTTP/GRPC response time, and TCP throughput for workloads acting in both client and server roles.
```
JonathanWamsley, line 35:
```suggestion
The Istio workloads dashboard provides details on HTTP/GRPC throughput, HTTP/GRPC response time, and TCP throughput for workloads acting in both client and server roles.
```

istio-2-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,164 @@
...
+    local uid = g.util.string.slugify(this.config.uid);
+    local vars = this.grafana.variables;
+    local annotations = this.grafana.annotations;
+    local refresh = this.config.dashboardRefresh;
+    local period = this.config.dashboardPeriod;
+    local timezone = this.config.dashboardTimezone;
+    local panels = this.grafana.panels;
+    local stat = g.panel.stat;
+    {
+      overview:
```
JonathanWamsley, line 18: Is there a way to give the output dashboard a name here? Maybe adding in a dashboardUid so that generated dashboards have the correct name and file type json as an output when running the `make` command.

JonathanWamsley (changes requested):
Made a couple of suggestions. Overall looks pretty good. 👍

istio-2-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,163 @@
...
+          summary: 'There are a high number of GRPC request errors in the Istio system.',
+          description: |||
+            GRPC requests from Istio service {{$labels.source_canonical_service}} to service {{$labels.destination_canonical_service}} on cluster {{$labels.cluster}} have an error rate above {{ printf "%%.0f" $value }}%%, which is above the threshold of %(alertsCriticalGRPCRequestErrorPercentage)s%%.
+          ||| % this.config,
+        },
+      },
+      {
+        alert: 'IstioMetricsDown',
+        expr: |||
+          up{%(filteringSelector)s} == 0
```
JonathanWamsley, line 147: Will this trigger if the prometheus exporter is running on the host itself and it goes down? Or is this an unlikely situation given Istio architecture and they are separate?

istio-2-mixin/variables.libsonnet
```diff
@@ -0,0 +1,204 @@
...
+          asc=true,
+          caseInsensitive=false
+        );
+        [variable],
+      overviewVariables:
+        [root.datasources.prometheus]
+        + groupVariablesFromLabels(groupLabels)
+        + createOverviewVariable(istiodLabel, 'Istiod', 'pilot_info', 'job=~"$job", cluster=~"$cluster"')
+        + createOverviewVariable(gatewayLabel, 'Gateway', 'istio_agent_process_cpu_seconds_total', 'job=~"$job", cluster=~"$cluster", pod=~"istio-egress.*|istio-ingress.*"')
+        + createOverviewVariable(proxyLabel, 'Proxy', 'istio_agent_process_cpu_seconds_total', 'job=~"$job", cluster=~"$cluster", pod!~"istio-egress.*|istio-ingress.*"'),
```
JonathanWamsley, line 123: Before the generated json had a sort variable set to 0 or 5(for "proxy" this may be a mistake from the old dashboard sent in review) for all the datasources. Now they appear to be 1

istio-2-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1061 @@
...
+            },
+            renameByName: {
+              'Value #A': 'HTTP/GRPC tx',
+              'Value #B': 'HTTP/GRPC rx',
+              'Value #C': 'HTTP/GRPC tx delay',
+              'Value #D': 'HTTP/GRPC rx delay',
+              'Value #E': 'HTTP tx success',
+              'Value #F': 'HTTP rx success',
+              'Value #G': 'TCP tx',
+              'Value #H': 'TCP rx',
```
JonathanWamsley, line 944: There appears to be in the old dashboard:
              "Value #I": "TCP client closes",
              "Value #J": "TCP client opens",
Did this get removed on purpose?

istio-2-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,164 @@
...
+            variables+: {
+              // add prometheus datasource for annotations processing
+              toArray+: [
+                this.grafana.variables.datasources.prometheus { hide: 2 },
+              ],
+            },
+          }.dashboards.logs,
+      }
+    else {},
+  //Apply common options(uids, tags, annotations etc..) to all dashboards above
```
JonathanWamsley, line 155: Are annotations supposed to be added here too? Such as:
  "annotations": {
    "list": [
      {
        "builtIn": 1,
        "datasource": {
          "type": "grafana",
          "uid": "-- Grafana --"
        },
        "enable": true,
        "hide": true,
        "iconColor": "rgba(0, 211, 255, 1)",
        "name": "Annotations & Alerts",
        "type": "dashboard"
      }
    ]
  },

istio-2-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1061 @@
...
+      proxies:
+        commonlib.panels.generic.stat.base.new(
+          'Proxies',
+          targets=[t.proxyCount],
+          description='Number of proxies in the Istio system.'
+        )
+        + stat.options.withGraphMode('none')
+        + stat.standardOptions.color.withMode('thresholds')
+        + stat.standardOptions.thresholds.withSteps([
+          stat.thresholdStep.withColor('super-light-red')
```
JonathanWamsley, line 29: In the old dashboard, there was an additional mode set to `absolute` instead of `threshold` like:
"thresholds": {
            "mode": "absolute",
            "steps": [
              {
                "color": "super-light-red",
                "value": null
              },
              {
                "color": "super-light-green",
                "value": 1
              }
            ]
          },
          
          Not sure if this needs to be updated.

istio-2-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1061 @@
...
+          + stat.thresholdStep.withValue(null),
+          stat.thresholdStep.withColor('super-light-green')
```
JonathanWamsley, line 31: There is also a  `"unitScale": true` value that may be missing. Not sure what the default is. This applies to other panels as well.

JonathanWamsley (commented):
little bit more feedback comparing the dashboards.

istio-2-mixin/panels.libsonnet
StefanKurek: @JonathanWamsley I don't think I care about this one?

istio-2-mixin/panels.libsonnet
StefanKurek: @JonathanWamsley yeah, those were old and not doing anything

istio-2-mixin/panels.libsonnet
StefanKurek: @JonathanWamsley It actually looks like the generated json has mode as "absolute" currently

istio-2-mixin/dashboards.libsonnet
JonathanWamsley: You can update this by using a string `'istio-service-overview.json'` or whatever the correct name for the dashboard you want in the dashboard.libsonnet in place of the `overview` variable.

istio-2-mixin/alerts.libsonnet
JonathanWamsley: So here is my concern on a more clear example.  Say you have a VM instance with MySQL and a prometheus exporter on the same instance. Then if something wrong occurs with the instance, the metrics won't be reporting at all and the value is nothing, not 0 right?  

Looking back at it, I suppose this is not saying istio down, but is saying istio metrics down. So it is saying something is wrong with the prometheus exporter or configuration to scrape and emit successfully. Maybe it is fine then.

BominRahmani (approved):
image
One small thing i noticed, is that the logs still has a "logs" legend, it parses out the majority of the response codes correctly, but if you check the ones that are still being labeled as "logs" you can see that they actually have a status code of 200 most of the time. I wonder if the regex might need to get tweaked or whats happening there.
Other than that this looks great to me!

istio-2-mixin/panels.libsonnet
JonathanWamsley: I just generated it and got it showing mode as thresholds while the old json showed it as absolute before. This is because of `+ stat.standardOptions.color.withMode('thresholds')` line. Just want to make sure this is correct:
```json
            "fieldConfig": {
               "defaults": {
                  "color": {
                     "mode": "thresholds"
                  },
                  "thresholds": {
                     "steps": [
                        {
                           "color": "super-light-red",
                           "value": null
                        },
                        {
                           "color": "super-light-green",
                           "value": 1
                        }
                     ]
                  }
               }
            },
```

istio-2-mixin/dashboards.libsonnet
JonathanWamsley: Actually, I am not sure if this is a solution if you have multiple dashboards that need to reference dashboard variables like in your case.

istio-2-mixin/panels.libsonnet
StefanKurek: @JonathanWamsley take a look at the JSON in the dashboards in the cloud. I'm seeing this being generated.
```
      "fieldConfig": {
        "defaults": {
          "color": {
            "mode": "thresholds"
          },
          "mappings": [],
          "thresholds": {
            "mode": "absolute",
            "steps": [
              {
                "color": "super-light-orange",
                "value": null
              },
              {
                "color": "super-light-green",
                "value": 1
              }
            ]
          }
        },
```

Not sure why you're seeing something different generated. Can you tell me which panel you're looking at specifically?

JonathanWamsley: ahh, I see. So the color mode is set to threshold which matches with the old dashboard. Then there is another mode under the `thresholds` that is left empty. But by default, when you generated the dashboard, it populates with absolute which matches with the old dashboard. 👍

JonathanWamsley (approved):
LGTM ✅

istio-2-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1061 @@
...
+          description='Available virtual memory compared to the resident memory for the various components of the Istio system.',
+        )
+        + timeSeries.options.legend.withPlacement('right')
+        + timeSeries.fieldConfig.defaults.custom.withScaleDistributionMixin({
+          log: 2,
+          type: 'log',
+        })
+        + timeSeries.standardOptions.withUnit('bytes'),
+      heapMemory:
+        commonlib.panels.generic.timeSeries.base.new(
```
v-zhuravlev, line 156:
```suggestion
        commonlib.panels.memory.timeSeries.usageBytes.new(
```
You could try memory specific panel instead:

istio-2-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1061 @@
...
+          ],
+          description='Number of open file descriptors for various components of the Istio system.',
+        )
+        + timeSeries.options.legend.withPlacement('right')
+        + timeSeries.fieldConfig.defaults.custom.withStackingMixin({
+          group: 'A',
+          mode: 'normal',
+        }),
+      virtualAndResidentMemory:
+        commonlib.panels.generic.timeSeries.base.new(
```
v-zhuravlev, line 137:
```suggestion
        commonlib.panels.memory.timeSeries.usageBytes.new(
```

v-zhuravlev (commented):
Great work! Some minor suggestions above.

istio-2-mixin/config.libsonnet
```diff
@@ -0,0 +1,32 @@
+{
+  // Static selector to apply to ALL dashboard variables of type query, panel queries, alerts and recording rules.
+  filteringSelector: 'job=~"integrations/istio"',
+  // Used to identify 'group' of instances.
+  groupLabels: ['job', 'cluster'],
+
```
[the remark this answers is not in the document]
StefanKurek: @v-zhuravlev are you thinking about having these labels be variables/selectors on the dashboards as well?

Istio is a bit of a different case, where you're interested in how the services/workloads are interacting with each other. And since each pod would correspond with an "instance" selector, I'm not sure how useful an instance and/or a pod variable/selector would be in the dashboards. We already have variables to filter by the things that we care about.

I'm actually not sure if it would even make sense with how the dashboard are structured (the services dashboard is supposed to get info across all pods for a service...so filtering out certain pods seems counter intuitive)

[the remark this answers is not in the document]
v-zhuravlev: Ok, good point! I guess you can skip this then.

istio-2-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,168 @@
...
+
+    groups: [
+      {
+        name: 'istio-alerts-' + this.config.uid,
+        rules: [
+          {
+            alert: 'IstioHighCPUUsageWarning',
+            expr: |||
+              100 * (sum without (instance, pod) (rate(process_cpu_seconds_total{%(filteringSelector)s, pod=~"istiod.*"}[5m])) + sum without (instance, pod) (rate(istio_agent_process_cpu_seconds_total{%(filteringSelector)s}[5m]))) > %(alertsWarningHighCPUUsage)s
+            ||| % this.config,
```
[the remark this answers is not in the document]
StefanKurek: Note: not doing this due to below discussion

istio-2-mixin/panels.libsonnet
StefanKurek: Changed

v-zhuravlev (approved):

rgeyer (approved):
LGTM

PR #1150, opened 2024-02-23
Support recording and switching between naive and classic latency histograms

krajorama:
Support migration of latency histograms by allowing the user to chose between which one to show: native or classic ?

Also add plumbing for recording related native histogram series.


beorn7 (commented):
Not sure what I am supposed to review here. The Grafana magic is way beyond me. The PromQL queries look OK at first glance, but I should probably see the resulting queries in action to really tell.

krajorama:
> Not sure what I am supposed to review here. The Grafana magic is way beyond me. The PromQL queries look OK at first glance, but I should probably see the resulting queries in action to really tell.

I'd say this is optional for you @beorn7 , not a lot to do with PromQL. Usage is in https://github.com/grafana/mimir/pull/7674 and I have a prototype running, I'll share the link in DM

johannaratliff (approved):
Nothing crazy sticks out to me. Approved as a utils/usage experience

PR #1154, opened 2024-02-28
Add K8 Support to PgBouncer Mixin

BominRahmani:
Adds K8 Support to the PgBouncer Mixin that can be enabled/disabled through the `enableMultiCluster` toggle in the config.


StefanKurek (approved):
I think this is all there is to it which is nice?

Caleb-Hurshman (approved):
Looks like an improvement over the old way ✅

PR #1156, opened 2024-03-04
utils: allow defining native histogram recording rule

krajorama:
For native histograms, we can define a one recording rule to rule them all.

Related to https://github.com/grafana/mimir/issues/7154


duricanikolic (approved):
LGTM

PR #1161, opened 2024-03-12
PgBouncer K8 Labels Fix

BominRahmani:
This PR does a couple things.

- it fixes the k8's labels being written over by the mixin.libsonnet
- uses different metric for variable discovery
- refactors the usage of some labels


gaantunes (approved):
LGTM.

PR #1163, opened 2024-03-13
Golang runtime observ lib

v-zhuravlev:
This lib can produce panels/rows and dashboards for both Golang instrumentation types:
1) prometheus 
2) otel

As input, this lib uses commonlib 'signals' that describe which KPI we want to track. Than we can quickly use commonlib signals package to transform signals to panels of different types (timeseries,stat...), panelMixins(targets+overrides) or targets.

Results of this observability library (panesl/rows) can be used together to compose other dashboards.

Prometheus based:
image

OTel based:
image


golang-observ-lib/README.md
```diff
@@ -0,0 +1,10 @@
+# Golang observ lib
```
Dasomeone, line 1: Could you include some screenshots like the ones used in the PR body? I think that'd be great as an idea for people thinking of adopting it

v-zhuravlev: Added screenshots

PR #1167, opened 2024-03-14
Fix: PgBouncer Database Selector

BominRahmani:
Allows PgBouncer overview dashboard to keep singleInstance while allowing multiple databases


gaantunes (approved):
LGTM

PR #1177, opened 2024-03-28
Feat: Add Velero Mixin

BominRahmani:
The following PR will serve as the base PR for which every other Velero mixin PR will merge into. The base PR contains only the ReadMe

### DIFF
--- a/velero-2-mixin/README.md
+++ b/velero-2-mixin/README.md
@@ -1,0 +1,172 @@
+# Velero Mixin
+
+The Velero mixin is a set of configurable Grafana dashboards and alerts.
+
+The Velero mixin contains the following dashboards:
+
+- Velero cluster overview
+- Velero overview
+- Velero logs overview
+
+and the following alerts:
+
+- VeleroBackupFailure
+- VeleroHighBackupDuration
+- VeleroHighRestoreFailureRate
+- VeleroUpStatus
+
+## Velero Cluster Overview
+
+The Velero cluster overview dashboard a top level overview that provides details on alerts, volume snapshots, backup and restore counts.
+
+![Velero Overview Dashboard]()
+
+## Velero Overview
+
+The Velero overview dashboard provides details on backup sizes, backup success rates, backup status, restore counts, restore success rates, and snapshot data.
+
+![Velero Overview Dashboard]()
+
+## Velero Logs Overview
+
+The Velero logs overview dashboard provides details on the Velero system. [Promtail and Loki needs to be installed](https://grafana.com/docs/loki/latest/installation/) and provisioned for logs with your Grafana instance.
+
+![Velero Logs Dashboard]()
+
+Velero logs are enabled by default in the `config.libsonnet` and can be removed by setting `enableLokiLogs` to `false`. Then run `make` again to regenerate the dashboard:
+
+```
+{
+  _config+:: {
+    enableLokiLogs: false,
+  },
+}
+```
+
+In order for the selectors to properly work for system logs ingested into your logs datasource, here is what the agent flow snippet will look like:
+
+```yaml
+  discovery.relabel "velero" {
+    targets = discovery.kubernetes.pods.targets
+    rule {
+      action        = "keep"
+      source_labels = ["__meta_kubernetes_pod_label_component"]
+      regex         = "velero"
+    }
+    rule {
+      source_labels = ["__meta_kubernetes_pod_container_port_number"]
+      regex = "8085"
+      action = "keep"
+   }
+    rule {
+      source_labels = ["__meta_kubernetes_pod_name"]
+      separator = "-"
+      target_label = "instance"
+    }
+    rule {
+      replacement = "integrations/velero"
+      target_label = "job"
+    }
+  }
+
+  prometheus.scrape "velero" {
+    targets      = discovery.relabel.velero.output
+    metrics_path = "/metrics"
+    forward_to   = [prometheus.relabel.metrics_service.receiver]
+  }
+  
+  discovery.relabel "logs_velero" {
+      targets = discovery.relabel.pod_logs.output
+
+      rule {
+        action        = "keep"
+        source_labels = ["__meta_kubernetes_pod_label_component"]
+        regex         = "velero"
+      }
+      rule {
+        target_label = "job"
+        replacement = "integrations/velero"
+      }
+      rule {
+        action = "replace"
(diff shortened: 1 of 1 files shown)


velero-2-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,77 @@
...
+              summary: 'Velero backup failures detected.',
+              description: |||
+                Backup failures detected on {{ $labels.instance }}. This could lead to data loss or inability to recover in case of a disaster.
+              ||| % this.config,
+            },
+          },
+          {
+            alert: 'VeleroHighBackupDuration',
+            expr: |||
+              histogram_quantile(0.5, sum(rate(velero_backup_duration_seconds_bucket{%(filteringSelector)s}[5m])) by (le)) > %(alertsHighBackupDuration)s * avg_over_time(histogram_quantile(0.5, sum(rate(velero_backup_duration_seconds_bucket{%(filteringSelector)s}[5m])) by (le))[48h:5m])
```
v-zhuravlev, line 28: Great alert overall! Have you got any chance to test its behavior? I think it might be prone to produce false positives:
1) aggregating different backup jobs together (by le only). Perhaps, this should be aggregated by backup jobs to have better accuracy?

velero-2-mixin/alerts.libsonnet
```diff
@@ -0,0 +1,77 @@
...
+              up{%(filteringSelector)s} != %(alertsVeleroUpStatus)s
+            ||| % this.config,
+            'for': '5m',
+            labels: {
+              severity: 'critical',
+            },
+            annotations: {
+              summary: 'Velero is down.',
+              description: |||
+                Cannot find any metrics related to Velero on {{ $labels.instance }}. This indicates that Velero is not running or the Grafana agent is not properly configured.
```
v-zhuravlev, line 69:
```suggestion
                Cannot find any metrics related to Velero on {{ $labels.instance }}. This indicates that Velero is not running or the Grafana agent is not properly configured.
```
Scraping agents could be different: Grafana Agent/ Alloy/ Prometheus etc...

v-zhuravlev: Overall, I would rephrase this alert to be more specific. It is not necessary indicates that Velero is down. It could be just that agent can't scrape metric for various reasons: like metrics endpoint is not configured or auth issues etc...

velero-2-mixin/variables.libsonnet
```diff
@@ -0,0 +1,150 @@
...
+// Generates chained variables to use on on all dashboards
+{
+  new(this, varMetric):
+    {
+      local filteringSelector = this.config.filteringSelector,
+      local groupLabels = this.config.groupLabels,
+      local instanceLabels = this.config.instanceLabels,
+      local scheduleLabel = 'schedule',
+      local groupVarMetric = 'velero_backup_attempt_total',
+      local querySelectorVar = ['job', 'cluster', 'instance', 'schedule'],
```
v-zhuravlev, line 15: shouldn't those be a combination of groupLabels, instanceLabels?

velero-2-mixin/variables.libsonnet
v-zhuravlev, line 1: You can drop this file altogether and import it inside main.libsonnet from commonlib now:
https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/variables/variables.libsonnet

velero-2-mixin/variables.libsonnet
BominRahmani: Yea, I've updated it now so its a combination of groupLabels, instanceLabels, and then the schedule label. Let me know what you think.

velero-2-mixin/variables.libsonnet
BominRahmani: Since i've had some further modifications done in this variables file, i am going to keep it.

velero-2-mixin/alerts.libsonnet
BominRahmani: I've set it up and have been keeping an eye on it. I havent seen it produce any false positives yet, but that maybe just luck. I can try doing aggregations by backup jobs as well though. I do not feel strongly about it either way. Let me know what you think or maybe @mshahzeb can weigh in on what he thinks as well

velero-2-mixin/mixin.libsonnet
```diff
@@ -0,0 +1,19 @@
+local velerolib = import './main.libsonnet';
+
+local velero =
+  velerolib.new()
+  + velerolib.withConfigMixin(
+    {
+      filteringSelector: 'job=~"integrations/velero"',
```
mshahzeb, line 7: Since this is a public mixin in jsonnet-lib. I am wondering if we should be hardcoding the selector `integrations/velerio` here?

velero-2-mixin/Makefile
```diff
@@ -0,0 +1,34 @@
...
+dashboards_out: mixin.libsonnet config.libsonnet dashboards.libsonnet
+	@mkdir -p dashboards_out
+	mixtool generate dashboards mixin.libsonnet -d dashboards_out
+
+prometheus_alerts.yaml: mixin.libsonnet alerts.libsonnet
+	mixtool generate alerts mixin.libsonnet -a prometheus_alerts.yaml
+
+.PHONY: clean
+clean:
+	rm -rf dashboards_out prometheus_alerts.yaml
```
mshahzeb, line 34: Missing newline

mshahzeb (changes requested):

velero-2-mixin/mixin.libsonnet
BominRahmani: Yea i think this is fine, it is how its been done historically.

v-zhuravlev: Yes, we didn't pay attention to this. Overall, I think putting less strict filter on the mixin level here would be better, then overriding in integrations.

velero-2-mixin/alerts.libsonnet
v-zhuravlev: If you tend to keep it I suggest two things:
- Split alerts by backup jobs separately so previous day stats are more accurate
- Try add `keep_firing_for: 24h` So alert don't get resolved after 5 minutes after alert is triggered. with keep_firing_for that long I expect it only gets resolved if next day backup is still having issues.

mshahzeb (approved):
LGTM! Thanks!

v-zhuravlev (approved):

PR #1197, opened 2024-04-22
Mixin updates: update loki logs regex filter

JonathanWamsley:
This PR updates the regex to filter out the usage-insights and alert-state-history from the selectors. A couple were missed beforehand that uses the newest mixin format include:
```+ var.datasource.withRegex('(?!grafanacloud.+usage-insights|grafanacloud.+alert-state-history).+')```
This follows the pattern like in [openstack](https://github.com/grafana/jsonnet-libs/blob/master/openstack-mixin/variables.libsonnet#L47).

### DIFF
--- a/helloworld-observ-lib/variables.libsonnet
+++ b/helloworld-observ-lib/variables.libsonnet
@@ -43,7 +43,7 @@ local utils = commonlib.utils;
         loki:
           var.datasource.new('loki_datasource', 'loki')
           + var.datasource.generalOptions.withLabel('Loki data source')
-          + var.datasource.withRegex('')
+          + var.datasource.withRegex('(?!grafanacloud.+usage-insights|grafanacloud.+alert-state-history).+')
           // hide by default (used for annotations)
           + var.datasource.generalOptions.showOnDashboard.withNothing(),
       },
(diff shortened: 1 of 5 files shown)


Dasomeone (approved):
@v-zhuravlev can you please have a look to double check if we want this in the hello-world and windows observ libs?

v-zhuravlev:
Hi! There is a `utils` function on the integration levels that patches such filter in for all integrations before they are loaded. So it should be already covered. Mixins themselves here, we should try to keep without any integrations specific .

JonathanWamsley:
> Hi! There is a `utils` function on the integration levels that patches such filter in for all integrations before they are loaded. So it should be already covered. Mixins themselves here, we should try to keep without any integrations specific .

If I am understanding this right, I should close this pr then? @v-zhuravlev @Dasomeone

Dasomeone:
> If I am understanding this right, I should close this pr then? @v-zhuravlev @Dasomeone

@JonathanWamsley I think that's safe to do so, Vitaly is certainly more up to date on this topic than I am. We'll just have to be careful when reviewing based on Grizzly installs then

v-zhuravlev:
here it is btw: https://github.com/grafana/jsonnet-libs/blob/master/grafana-cloud-integration-utils/util.libsonnet#L117

PR #1205, opened 2024-04-24
Commonlib signals update

v-zhuravlev:
Add detailed readme
Add marshallJsonMutli function: multiple metric sources can be defined for each signal  (otel, prometheus..., )
Add raw metric type(where expression can be defined manually)
Add additioanal functions:  asPanelMixin(), asPanelExpression()
Update helloworld example with signals


common-lib/common/signal/signal.libsonnet
```diff
@@ -5,52 +5,151 @@ local counter = import './counter.libsonnet';
...
+  //       type: 'info',
+  //       description: 'Golang version used.',
+  //       expr: 'go_info{%(queriesSelector)s}',
+  //       infoLabel: 'version',
+  //     },
+  //     signal2:...,
+  //     signal3:....
+  //   }
+  // }
+  unmarshallJson(signalsJson):
```
rgeyer, line 31: This is actually extraordinarily powerful and impactful.

This can allow us to COMPLETELY decouple the desired signals from the resulting dashboard, we could even store the signals in a separate file like `signals.json`, and allow it to be loaded by the mixin, or used as a basis for others dashboards.

When you added this, how did you envision it being used? Could you elaborate a bit on the use-cases in the README?

common-lib/common/signal/test_marshall_json.libsonnet
```diff
@@ -0,0 +1,135 @@
+local signal = import './signal.libsonnet';
+local test = import 'jsonnetunit/test.libsonnet';
+
+local jsonSignals =
+  {
+    aggLevel: 'group',
+    groupLabels: ['job'],
+    instanceLabels: ['instance'],
+    filteringSelector: 'job="integrations/agent"',
```
rgeyer, line 9: nit; We've moved all of our external facing assets (where possible) to use `alloy` in place of `agent`. Would be nice to update that here.

helloworld-observ-lib/.lint
```diff
@@ -7,3 +7,5 @@ exclusions:
     reason: "grafonnet created panels use --Mixed-- datasource by default."
   template-instance-rule:
     reason: "instanceLabels must be single-select for single instance dashboards."
+  target-job-rule:
+    reason: "using filtering selector lik job=\"<static value\" which is totally fine."
```
rgeyer, line 11:
```suggestion
    reason: "using filtering selector like job=\"<static value\" which is totally fine."
```

rgeyer (approved):
A few nits, and spelling fixes, and one question about the utility of two functions where the description is effectively the same.

These are all minor, so I'm going to approve, but would appreciate feedback on my question.

Awesome work!

common-lib/common/signal/signal.libsonnet
v-zhuravlev: Thanks. Indeed, decoupling is the key. I also wanted to keep signals as part of jsonnet config or event json, so they can be modified by users before generating final dashboards and so on. Also I keep grafana foundation sdk in mind, so perhaps one days those json files with signals can be used to generate dashboards using Go or TypeScript as well.

v-zhuravlev: For now, I think docker lib update can service as good use case example.

common-lib/common/signal/README.md
```diff
@@ -1,19 +1,81 @@
...
 
 Workflow to generate dashboards would be as follows:
+
 - Define key indicators(signals) your want to observe.
-- Use built-in functions to render each signal as a panel (i.e. signal.asTimeSeries())
-- If you want to combine multiple signals inside a single panel, render signal not as [Grafana panel](https://grafana.github.io/grafonnet/API/panel/index.html), but as a [Target](https://grafana.github.io/grafonnet/API/panel/timeSeries/index.html#fn-queryoptionswithtargets), which is attachable to existing panel
+- Use built-in functions to render each signal as a panel (i.e. `signal.asTimeSeries()`, `signal.asStat()`, `signal.asGauge()`, `signal.asStatusHistory()`)
+- If you want to combine multiple signals inside a single panel, render signal not as [Grafana panel](https://grafana.github.io/grafonnet/API/panel/index.html), but as a [Target](https://grafana.github.io/grafonnet/API/panel/timeSeries/index.html#fn-queryoptionswithtargets), which is attachable to existing panel. (`signal.asTarget()`)
+- If you need to put multiple signals inside single panel use `signal.asPanelMixin()` function. It would add `Target` And overrides relevant to it (units, value mappings...)
```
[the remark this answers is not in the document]
v-zhuravlev: asTarget() - renders just `target` part of a panel, in other words, a query with legendFormat etc..
asPanelMixin() - actually more useful. It patches any panel not only with `target` but also adds panel overrides to make sure units or value mappings defined in signals are applied to this signal as well!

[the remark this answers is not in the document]
v-zhuravlev: Use case for using just asTarget() could be that you need to modify a target: i.e. changing query type from timeSeries to table , from Range to Instance and so on.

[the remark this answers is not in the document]
rgeyer: Makes sense. Thanks for elaborating. Could we make this more clear in an example or README? I'll read through it again with this context and see if there are any improvements I can think of.

PR #1206, opened 2024-04-25
fix: removed pod=~"$pod" filter from CiliumAgentMapOperationFailures alert rule

balous:
This PR removes `pod=~"$pod"` filter from `CiliumAgentMapOperationFailures` alert rule in `cilium-enterprise-mixin`.  The filter causes the rule to not match anything as there is no `$pod` pod and I believe it was accidentally copied from similar query in dashboards.


Dasomeone (approved):
Nice catch, thanks @balous!

PR #1210, opened 2024-04-29
Add K8s Support to Apache Solr

Caleb-Hurshman:
This PR adds support for K8s deployments of Apache Solr.

### DIFF
--- a/apache-solr-mixin/config.libsonnet
+++ b/apache-solr-mixin/config.libsonnet
@@ -4,6 +4,8 @@
     solrSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
     multiclusterSelector: 'job=~"$job"',
     filterSelector: 'job=~"integrations/apache-solr"',
+    logLabels: if self.enableMultiCluster then ['job', 'cluster', 'solr_cluster', 'instance', 'level', 'filename']
+    else ['job', 'solr_cluster', 'instance', 'level', 'filename'],
 
     dashboardTags: ['apache-solr-mixin'],
     dashboardPeriod: 'now-30m',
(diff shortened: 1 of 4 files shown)


apache-solr-mixin/config.libsonnet
```diff
@@ -1,6 +1,6 @@
 {
   _config+:: {
-    enableMultiCluster: false,
+    enableMultiCluster: true,
```
JonathanWamsley, line 3: I believe this should be false by default still

JonathanWamsley (commented):
Overall looks pretty good. I think the logs overview needs a cluster selector too now right?

Caleb-Hurshman:
> Overall looks pretty good. I think the logs overview needs a cluster selector too now right?

I thought so too, but looking at the docs and previous examples, we've never changed the log dashboards. Maybe I'm looking in the wrong places, lmk if you know something I'm forgetting.

JonathanWamsley:
> > Overall looks pretty good. I think the logs overview needs a cluster selector too now right?
> 
> I thought so too, but looking at the docs and previous examples, we've never changed the log dashboards. Maybe I'm looking in the wrong places, lmk if you know something I'm forgetting.

I think they did not have the individual logs dashboard. It seems you updated the selector though for [couchbase](https://github.com/grafana/jsonnet-libs/blob/664ac8fda54706b3b0a089f3b47de40644dade5b/couchbase-mixin/dashboards/couchbase-node-overview.libsonnet#L1019) before. You can add an `cluster` label in the newer logs dashboard like with [clickhouse](https://github.com/grafana/jsonnet-libs/blob/6f568d7bc14260f10321f1344f53cfb2d6fc146f/clickhouse-mixin/dashboards/addlogs.libsonnet#L13) .

JonathanWamsley (approved):
Looks good 👍

Dasomeone (approved):
LGTM!

PR #1212, opened 2024-04-30
Aggregation labels for cilium mixin

balous:
I've implemented possibility to aggregate `cilium-enterprise-mixin` alerts by arbitrary set of labels. All alerts are aggregated by the same set, the default set is `[]` (which preserves previous behavior).

Are you interrested in such a change?


Dasomeone (approved):
Hi @balous

Yes, thank you, that sounds like it would be a good addition indeed! :+1:

PR #1216, opened 2024-05-02
OpenSearch `k8s` plugin support

algchoo:
### Description

This PR updates the OpenSearch mixin to support use with the `k8s` plugin. There is a note added to the documentation regarding the `cluster` label that is included with metrics from the exporter. 

**Note:** The logs panel on the node dashboard will not work for me. It says that it cannot find the datasource, even though when I visit the explorer, I can see a loki datasource and the logs from OpenSearch.

### Changes
* [wip: mixin update to include cluster and opensearch_cluster labels](https://github.com/grafana/jsonnet-libs/commit/6b171b0f1763a9fdfa65c9843724c24551901666)
* [note regarding labels on exporter metrics](https://github.com/grafana/jsonnet-libs/commit/a1efc705aa8497feffeb9faa30e64d26a9682856)
* [changed note indentation for consistency](https://github.com/grafana/jsonnet-libs/commit/dafc47261c35b5385eaa5798e3a2e37c562115db)

### DIFF
--- a/opensearch-mixin/config.libsonnet
+++ b/opensearch-mixin/config.libsonnet
@@ -1,12 +1,9 @@
 {
   _config+:: {
     enableMultiCluster: false,
-    multiclusterSelector: 'job=~"$job"',
-    opensearchSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
-
     // extra static selector to apply to all templated variables and alerts
-    filteringSelector: 'cluster!=""',
-    groupLabels: ['job', 'cluster'],
+    filteringSelector: if self.enableMultiCluster then 'opensearch_cluster!=""' else 'cluster!=""',
+    groupLabels: if self.enableMultiCluster then ['job', 'opensearch_cluster'] else ['job', 'cluster'],
     instanceLabels: ['node'],
     dashboardTags: ['opensearch-mixin'],
     dashboardPeriod: 'now-1h',
--- a/opensearch-mixin/config.libsonnet
+++ b/opensearch-mixin/config.libsonnet
@@ -2,8 +2,8 @@
   _config+:: {
     enableMultiCluster: false,
     // extra static selector to apply to all templated variables and alerts
-    filteringSelector: if self.enableMultiCluster then 'opensearch_cluster!="",cluster!=""' else 'opensearch_cluster!=""',
-    groupLabels: if self.enableMultiCluster then ['job', 'opensearch_cluster', 'cluster'] else ['job', 'opensearch_cluster'],
+    filteringSelector: if self.enableMultiCluster then 'cluster!="",opensearch_cluster!=""' else 'opensearch_cluster!=""',
+    groupLabels: if self.enableMultiCluster then ['job', 'cluster', 'opensearch_cluster'] else ['job', 'opensearch_cluster'],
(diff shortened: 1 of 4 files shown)


opensearch-mixin/config.libsonnet
```diff
@@ -1,8 +1,9 @@
 {
   _config+:: {
+    enableMultiCluster: false,
     // extra static selector to apply to all templated variables and alerts
-    filteringSelector: 'cluster!=""',
-    groupLabels: ['job', 'cluster'],
+    filteringSelector: if self.enableMultiCluster then 'opensearch_cluster!="",cluster!=""' else 'opensearch_cluster!=""',
+    groupLabels: if self.enableMultiCluster then ['job', 'opensearch_cluster', 'cluster'] else ['job', 'opensearch_cluster'],
```
BominRahmani, line 6: I believe you would want opensearch_cluster to come after cluster since the hierarchy goes from overarching down.

BominRahmani: so something like 
```suggestion
    groupLabels: if self.enableMultiCluster then ['job', 'cluster', 'opensearch_cluster'] else ['job', 'opensearch_cluster'],
```

opensearch-mixin/config.libsonnet
```diff
@@ -1,8 +1,9 @@
 {
   _config+:: {
+    enableMultiCluster: false,
     // extra static selector to apply to all templated variables and alerts
-    filteringSelector: 'cluster!=""',
-    groupLabels: ['job', 'cluster'],
+    filteringSelector: if self.enableMultiCluster then 'cluster!="",opensearch_cluster!=""' else 'opensearch_cluster!=""',
```
BominRahmani, line 5: Sorry not to familiar with recent k8 stuff, do we usually change the filtering selector with these k8 updates?

algchoo: @BominRahmani In this case, the metrics are exported with a `cluster` label by default and we've opted to replace that label with `opensearch_cluster`.

The reason for this is that the `cluster` label, in the context of k8s, is meant to represent the k8s cluster. Keeping the label as `cluster` when the deployment is not k8s based seemed inconsistent.

BominRahmani (approved):

Caleb-Hurshman (approved):

Dasomeone:
> Note: The logs panel on the node dashboard will not work for me. It says that it cannot find the datasource, even though when I visit the explorer, I can see a loki datasource and the logs from OpenSearch.

Indeed @algchoo, it looks like only the prometheus datasource is built into the node overview. Not sure when/how that was overlooked, but we'll need to add it 
image

Dasomeone (changes requested):
The K8s changes themselves look fine, but as you call out in the PR body, we'll need to add the missing loki datasource to the node dashboard before we merge this in :)

algchoo:
@v-zhuravlev Hope you're well! There's an issue with the Loki datasource on a dashboard, where we expect to see logs, but can't because this datasource doesn't exist. There was a [refactor](https://github.com/grafana/jsonnet-libs/pull/1045) and I'm unclear on how to modify the mixin to include this datasource, is this something you can help with?

v-zhuravlev:
this one should fix it

Dasomeone (approved):
LGTM, thanks for adding in the fix :)

PR #1222, opened 2024-05-06
Keep datasource in next level dashboard + "Filter by cluster/job" fix

clementduveau:
Hello team,

In the Fleet view for Windows and Linux (Linux is in another repo so I will fix it in another PR), we have 1 broken filter and 1 "sometimes" broken link.

Go to a Linux or Windows fleet overview and click on cluster or job in the table. It will break because the datasource is empty in the link.
Now select a non-default datasource, click on an instance. The next level dashboard is empty because the datasource is not part of the URL and so you end up on the default Prometheus which doesn't have your instance.

Simple fix: I added the datasource at the end of those URL.

### DIFF
--- a/windows-observ-lib/panels.libsonnet
+++ b/windows-observ-lib/panels.libsonnet
@@ -78,7 +78,7 @@ local utils = commonlib.utils;
             {
               targetBlank: false,
               title: 'Drill down to ${__field.name} ${__value.text}',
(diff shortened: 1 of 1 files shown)


clementduveau:
I saw that we are waiting on Prometheus to merge our refactor in node_exporter. So I am not going to touch that yet.

v-zhuravlev (approved):

v-zhuravlev:
Thanks!

PR #1226, opened 2024-05-08
Fix Apache Solr selector metric

Caleb-Hurshman:
This PR corrects a copy-paste error where the Apache Solr `cluster` selector used an incorrect metric on two dashboards.


nikosmeds (approved):

Dasomeone (approved):
Good catch, thanks Caleb!

PR #1228, opened 2024-05-10
fix: added threshold to cilium alerts with rate() function

balous:
Some cilium alerts use `rate()` without comparing the result with a threshold. `rate()` produces 0 value when the metric exists and is not increasing - i.e. when the problem had happened in the past but is not happening now. If the result of the function is not compared with a threshold, the alert fires all the time even when nothing is happening. So I've added comparison with 0 where `rate()` is used and no comparison is present (I've changed `sum(rate(...))` to `sum(rate(...)) > 0` in four alerts).

### DIFF
--- a/cilium-enterprise-mixin/alerts/ciliumAlerts.libsonnet
+++ b/cilium-enterprise-mixin/alerts/ciliumAlerts.libsonnet
@@ -20,7 +20,7 @@
           },
           {
             alert: 'CiliumAgentEndpointUpdateFailure',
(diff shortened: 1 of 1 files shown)


Dasomeone (approved):
Much appreciated @balous! Some great catches on the Cilium mixin :)

balous:
I am glad I can use you library, it's less work to fix it than to create it from scratch.

Just in a case - would you be interested if I'd implemented configurable per alert thresholds?

Dasomeone:
> I am glad I can use you library, it's less work to fix it than to create it from scratch.
> 
> Just in a case - would you be interested if I'd implemented configurable per alert thresholds?

Certainly! There actually is a number of examples of configurable per-alert thresholds already in this repository, one of which can be found with the [aerospike-mixin](https://github.com/grafana/jsonnet-libs/blob/master/aerospike-mixin/config.libsonnet), though it requires refactoring the mixin to libsonnet.

If you want to take that on, then by all means! :)

PR #1229, opened 2024-05-13
Feat: Add catchpoint mixin

BominRahmani:
The following PR adds a Catchpoint mixin.
The dashboards can be seen [here](https://keithschmitty.grafana.net/dashboards/f/bdlwsf2y9tudcc/integrations-catchpoint)

### Catchpoint overview
image
image

### Catchpoint web performance by test name
image
image

### Catchpoint web performance by node name
image
image


catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        + barGauge.panelOptions.withDescription('Number of elements loaded.')
+        + barGauge.options.withOrientation('horizontal')
+        + barGauge.standardOptions.thresholds.withSteps([
+          barGauge.thresholdStep.withColor('super-light-green'),
+        ]),
+
+      errors:
+        barGauge.new(title='Errors')
+        + barGauge.queryOptions.withTargets([t.objectLoadedError, t.DNSError, t.loadError, t.timeoutError, t.connectionError, t.transactionError])
+        + barGauge.panelOptions.withDescription('Indicates various potential errors that are occuring.')
```
mshahzeb, line 270: Instead of "potential errors" it should be just "errors" since this panel will only show errors that occur and are not bound to occur.

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        + barGauge.panelOptions.withDescription('Number of elements loaded.')
+        + barGauge.options.withOrientation('horizontal')
+        + barGauge.standardOptions.thresholds.withSteps([
+          barGauge.thresholdStep.withColor('super-light-green'),
+        ]),
+
+      errorsNodeName:
+        barGauge.new(title='Errors')
+        + barGauge.queryOptions.withTargets([t.objectLoadedErrorNodeName, t.DNSErrorNodeName, t.loadErrorNodeName, t.timeoutErrorNodeName, t.connectionErrorNodeName, t.transactionErrorNodeName])
+        + barGauge.panelOptions.withDescription('Indicates various potential errors that are occuring.')
```
mshahzeb, line 410: Same error as above

mshahzeb (changes requested):
Great effort!

Mentioned some observations below.

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+      bottomAvgRequestRatioTestName:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Bottom average success request ratio by tests',
+          targets=[t.bottomAvgRequestSuccessRatioTestName],
+          description='The lowest average success request ratio among all test names over the specified interval.'
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true')
+        + g.panel.timeSeries.standardOptions.withMax(1)
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withUnit('percentunit'),
```
Dasomeone, line 61: For this panel, I would suggest reducing the fill opacity to 5-10% at most

Additionally, please remove the minimum of 0 as it introduces a lot of empty space for the gradient to be distracting.

Example w. changes side by side with the node panel which has the same feedback
image
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      bottomAvgRequestSuccessRatioNodeName:
```
Dasomeone, line 63: Same comment as for `bottomAvgRequestRatioTestName`

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
+
+      topAvgRedirectsNodeName:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Top average redirects by nodes',
+          targets=[t.topAvgRedirectsNodeName],
+          description='The top average number of redirects among all node names over the specified interval.'
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 124: There's an issue with the legend leaving trailing dashes in this panel:
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        + g.panel.timeSeries.standardOptions.withUnit('ms'),
+
+      topAvgContentLoadingTimeNodeName:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Top average content loading time by nodes',
+          targets=[t.topAvgContentLoadingTimeNodeName],
+          description='The top average content loading time among all node names over the specified interval.'
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true')
+        + g.panel.timeSeries.standardOptions.withUnit('ms'),
```
Dasomeone, line 108: There's an issue with the legend leaving trailing dashes in this panel:
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        alertList.new('Catchpoint alerts')
+        + alertList.options.UnifiedAlertListOptions.withAlertInstanceLabelFilter(this.grafana.variables.queriesGroupSelectorAdvanced),
+
+      topErrorsByTestName:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Top errors by tests',
+          targets=[t.topErrorsByTestName],
+          description='The top number of errors encountered among all test names over the specified interval.'
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 136: I think this panel would benefit from having a right-aligned table legend with values so users don't have to hover to find theirs.
image

I would then make it full-width as well to make space for it

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+        + g.panel.timeSeries.standardOptions.withUnit('hosts')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
+
+      cacheAccess:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Cache access',
+          targets=[t.cacheAccess],
+          description='Number of cached elements accessed.'
+        )
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 226: I think this might benefit from including values in a table legend?

Also not sure if this panel was tested with the sample load, because there's no real data here 
image

catchpoint-mixin/panels.libsonnet
Dasomeone, line 1: General comments throughout:
- The lines are VERY bold with 2 size. I'd recommend reducing to 1
- Colours seem.. very bright too, but that's probably personal preference and not helped by the next point
- Please reduce fill opacity to no more than 5-10%, or 15% if you must. 30% is way too much as seen in this example once it starts stacking
image
- When you have percentage panels 0-100%, please don't add a minimum of 0, as it'll introduce empty space (see pic above as well)
- I'm personally not a fan of line smoothing, makes it looks like a creative drawing and I'd personally prefer unsmoothed, but I'll leave that up to @mshahzeb to chime in on. 
- Please have a look at legends throughout, there's a lot of really congested panels that might benefit from either pruning the info available, or making them right-aligned

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      DNSResolution:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Connection and DNS resolution',
+          targets=[t.DNSResolution, t.SSLTime, t.connectTime],
+          description='Time taken to establish an SSL handshake, DNS resolution, and connect.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('ms')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 155: this panel needs some work on the legend, as it's crowded and hard to read.

Possibly a case for full-width + right aligned table legend with values, as it's currently cropping the legend just with test data
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      additionalDelay:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Additional delays',
+          targets=[t.additionalDelay, t.waitTime],
+          description='Additional delays encountered due to redirects, as well as time from successful connection to receiving the first byte.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('ms')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 182: This panel has the same issues and comment as `DNSResolution`

catchpoint-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,114 @@
...
+              g.panel.row.new('Connectivity'),
+              panels.topAvgConnectionSetupTimeTestName,
+              panels.topAvgConnectionSetupTimeNodeName,
+              panels.topAvgContentLoadingTimeTestName,
+              panels.topAvgContentLoadingTimeNodeName,
+              panels.topAvgRedirectsTestName,
+              panels.topAvgRedirectsNodeName,
+              g.panel.row.new('Errors'),
+              panels.topErrorsByTestName,
+              panels.alertsPanel,
```
Dasomeone, line 39: I would move the errors row to the absolute top, as it seems like critical information, especially with the alerts panel

catchpoint-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,114 @@
...
+              panels.networkConnections { gridPos+: { w: 8 } },
+              panels.hostsContacted { gridPos+: { w: 8 } },
+              panels.cacheAccess { gridPos+: { w: 8 } },
+              g.panel.row.new('Request'),
+              panels.requestSucessRatio,
+              panels.redirections,
+              g.panel.row.new('Errors and content types'),
+              panels.errors { gridPos+: { w: 8 } },
+              panels.contentTypesLoadedBySize { gridPos+: { w: 8 } },
+              panels.contentLoadedByType { gridPos+: { w: 8 } },
```
Dasomeone, line 68: I would move the errors row to the absolute top, as it seems like critical information (and breaks up the flow of timeseries panels)

catchpoint-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,114 @@
...
+              panels.networkConnectionsNodeName { gridPos+: { w: 8 } },
+              panels.hostsContactedNodeName { gridPos+: { w: 8 } },
+              panels.cacheAccessNodeName { gridPos+: { w: 8 } },
+              g.panel.row.new('Request'),
+              panels.requestSucessRatioNodeName,
+              panels.redirectionsNodeName,
+              g.panel.row.new('Errors and content types'),
+              panels.errorsNodeName { gridPos+: { w: 8 } },
+              panels.contentTypesLoadedBySizeNodeName { gridPos+: { w: 8 } },
+              panels.contentLoadedByTypeNodeName { gridPos+: { w: 8 } },
```
Dasomeone, line 97: Once again, would move the errors row to the top

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      // Web Performance by Tests Dashboard Panels
+      pageCompletionTime:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Page completion time',
+          targets=[t.pageCompletionTime, t.pageTotalLoadTime],
+          description='Time taken for the browser to fully render the page after all resources are downloaded.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('ms')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 146: This panel also suffers from uh, legend congestion shall we say 
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      contentHandlingNodeName:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Content handling',
+          targets=[t.contentHandlingLoad, t.contentHandlingRender],
+          description='Time taken to load and render content on the webpage.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('ms')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 304: Another case of legend congestion 
image

catchpoint-mixin/panels.libsonnet
```diff
@@ -0,0 +1,418 @@
...
+
+      totalContentSize:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Total content size',
+          targets=[t.totalContentSize, t.totalHeaderSize],
+          description='Total size of the HTTP response content and headers.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('decbytes')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls('true'),
```
Dasomeone, line 200: I think this would be a decent panel to have stacked values, to track overall Totals implicitly as well.

@mshahzeb what do you think?
image

catchpoint-mixin/README.md
```diff
@@ -0,0 +1,94 @@
...
+
+![Catchpoint overview dashboard]()
+
+## Catchpoint web performance by test name
+
+The Catchpoint web performance by test name dashboard provides details on page completion times, response sizes, network connections, requests, content types, and errors. These metrics are all grouped by test name.
+
+![Catchpoint web performance by test name]()
+
+## Catchpoint web performance by node name
```
Dasomeone, line 31: To Shahzeb's point, I don't think we need to specify "test name" and "node name", just drop the "name" part. Applies throughout the readme

catchpoint-mixin/README.md
```diff
@@ -0,0 +1,94 @@
...
+```js
+{
+  _config+:: {
+  alertsHighServerResponseTime: 1000, //ms
+  alertsHighServerResponseTimePercent: 1.2, // 120%
+  alertsTotalTimeExceeded: 5000, //ms
+  alertsTotalTimeExceededPercent: 1.2, // 120%
+  alertsHighDNSResolutionTime: 500, //ms
+  alertsHighDNSResolutionTimePercent: 1.2, // 120%
+  alertsContentLoadingDelay: 3000, //ms
```
Dasomeone, line 58: a 3 second delay seems surprisingly long, is that really what we want to recommend?

catchpoint-mixin/targets.libsonnet
```diff
@@ -0,0 +1,683 @@
...
+    local vars = this.grafana.variables,
+    local config = this.config,
+    local testNameLabel = config.testNameLabel,
+    local nodeNameLabel = config.nodeNameLabel,
+    local testAndNodeLabel = config.nodeNameLabel + config.testNameLabel,
+
+    topAvgLoadTimeTestName:
+      prometheusQuery.new(
+        '${' + vars.datasources.prometheus.name + '}',
+        'topk(1, avg by (test_name) (avg_over_time(catchpoint_load_time{%(pureTestNameSelector)s}[$__interval:])))' % vars
```
Dasomeone, line 19: @BominRahmani, @mshahzeb  Do you think there would be any value in making this `topk(k, avg[...])` where `k` is changeable by the user as a selector variable on the dashboards? 

(that'd apply throughout, I guess)

catchpoint-mixin/targets.libsonnet
```diff
@@ -0,0 +1,683 @@
...
+      + prometheusQuery.withLegendFormat('%s' % utils.labelsToPanelLegend(testAndNodeLabel)),
+
+    topErrorsByTestName:
+      prometheusQuery.new(
+        '${' + vars.datasources.prometheus.name + '}',
+        'topk(1, sum by (test_name) (sum_over_time(catchpoint_any_error{%(pureTestNameSelector)s}[$__interval:])))' % vars
+      )
+      + prometheusQuery.withLegendFormat('%s' % utils.labelsToPanelLegend(testNameLabel)),
+
+    // Web performance by test name dashboard
```
Dasomeone, line 107: Minor nit, remember to update all the "test name" and "node name" references throughout to drop "name"

Dasomeone (changes requested):
Bunch of comments, but please don't be discouraged, this is an excellent start!!

catchpoint-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,114 @@
...
+              panels.topAvgRedirectsNodeName,
+              g.panel.row.new('Errors'),
+              panels.topErrorsByTestName,
+              panels.alertsPanel,
+            ], 12, 6
+          )
+        )
+        // hide link to self
+        + root.applyCommon(vars.overviewVariables, uid + '-overview', tags, links { catchpointOverview+:: {} }, annotations, timezone, refresh, period),
+      testNameOverview:
```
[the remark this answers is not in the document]
BominRahmani: Yea for sure! the Screenshots in the PR descriptions werent meant as the screenshots that will be used in the readme/integration. Just wanted to give a rough idea to anyone glancing at the PR what to expect.

catchpoint-mixin/panels.libsonnet
BominRahmani: I'll make it 2/3 of the width and reduce the alerts to 1/3. I like the idea of having the table on the right.

catchpoint-mixin/panels.libsonnet
BominRahmani: Im not sure about the values, I could go for it if you think it would look nice. I am going to go ahead and setup the load gen for this metric real quick and i guess we can make a decision then.

catchpoint-mixin/panels.libsonnet
BominRahmani: Yea, the color, line boldness, and opacity are all the defaults that come with common.lib. I can change it for each of these panels, but i think it would probably be better if we change it on common lib.  @v-zhuravlev What do you think?

BominRahmani: As for the Max and Mins on the percentage graphs, I'll go ahead and swap those out.

catchpoint-mixin/targets.libsonnet
BominRahmani: I originally had this goal in mind, but @JonathanWamsley and I quickly found out that due to how the data is presented, introducing TopK with k selectors clutters up each of the panels to the point its unreadable. Each of the tests or nodes gets recognized as the top value at a specific point in time and if they are the only value that comes in at that period, then they are automatically recognized as a top value. This leads to basically every test or node being recognized as a top value which defeats the purpose of it. Its kind of hard to explain, you can see if you make the query you were talking about above with topk(k, avg[...]) and insert any number above 2 for k

catchpoint-mixin/README.md
BominRahmani: Yea i think this is a bit long, It was on the spec, but I personally think pulling it down to 1500 ms would be better. What do you think @Dasomeone?

catchpoint-mixin/dashboards.libsonnet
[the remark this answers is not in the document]
Dasomeone: Ah perfect, wasn't sure :D

catchpoint-mixin/panels.libsonnet
Dasomeone: Yeah I like this, that's a lot cleaner :smile:

catchpoint-mixin/panels.libsonnet
Dasomeone: That's a good point about those being defaults. I think it works on less populated panels, but for these where we have a lot of values it gets a bit much. Curious to see what you think in this case @v-zhuravlev, should we maybe have a set of presets for low/high density panels?

catchpoint-mixin/panels.libsonnet
Dasomeone: Having seen it with load gen, it's definitely better, but I still worry a little about it when it's scaled up with multiple labels.
Less so on this variant 
image


And more so on the "by test" variant where it's labelling the sites 
image


Honestly I think this is the sort of thing where we just have to go with one and adjust as we get feedback. I think I'm ok with it as is, but @mshahzeb what do you think?

catchpoint-mixin/panels.libsonnet
Dasomeone: Much better, thanks @BominRahmani!

catchpoint-mixin/targets.libsonnet
Dasomeone: I think I see what you mean yeah, with anything other than topk(1) being equivalent to dropping the topk entirely. 
Thanks for explaining!

Dasomeone (commented):
Lots of great improvements. Loving the new layout, so it mostly LGTM, just holding on approval until the discussion threads are wrapped up :)

mshahzeb (approved):

catchpoint-mixin/README.md
Dasomeone: Yes, I think that makes sense! Let's go with that and revisit in the future if a longer period seems useful

catchpoint-mixin/panels.libsonnet
mshahzeb: Yes I agree with having them stacked.

catchpoint-mixin/panels.libsonnet
mshahzeb: Yes I agree lets go with this one as is for now.

catchpoint-mixin/panels.libsonnet
BominRahmani: Sounds good, I'll get that implemented!

Dasomeone:
Went back over the PR as promised @BominRahmani.

I agree that changing the defaults re: line width and opacity is a discussion for elsewhere, but I still feel we need to override the opacity to a lower fill value on some of the more populated panels. Default doesn't mean we shouldn't change it when necessary after all :) 

Everything else, however, I think we can revisit later (including line width as that was more a minor nit)

PR #1236, opened 2024-05-24
Add support for renaming prometheus datasource and datasourceLabel in commonlib

v-zhuravlev:
Can be used to rename datasource name and label in signals:

for example:
```
local dockerlib = import './main.libsonnet';


local docker =
  dockerlib.new() +
  dockerlib.withConfigMixin(
    (import './config.libsonnet')._config
    {
      signals+: {
        machine+: {
          datasource: 'prometheus custom ds',
          datasourceLabel: 'custom_prom',
        },
        container+: {
          datasource: 'prometheus custom ds',
          datasourceLabel: 'custom_prom',
        },
      },
    },
  );

docker.asMonitoringMixin()
```

### DIFF
--- a/common-lib/common/signal/README.md
+++ b/common-lib/common/signal/README.md
@@ -36,6 +36,8 @@ Init level:
 
 |Name|Description|Possible values|Example value|Default value|
 |-----|---|---|---|---|
+|datasource| Prometheus Datasource name. |*|`custom_datasource`|`datasource`|
+|datasourceLabel| Prometheus Datasource label. |*|`Custom data source`|`Data source`|
 |filteringSelector|Used to filter metric scopes. Added to all dashboards'(as part of `queriesSelector`) and alerts' queries. |*|`[job="blablablah]"`,|`['job!=""']`|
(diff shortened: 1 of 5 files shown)


common-lib/common/signal/README.md
```diff
@@ -36,6 +36,8 @@ Init level:
 
 |Name|Description|Possible values|Example value|Default value|
 |-----|---|---|---|---|
+|datasource| Prometheus Datasource name. |*|`custom_datasource`|`datasource`|
+|datasourceLabel| Prometheus Datasource label. |*|`Custom data source`|`Data source`|
```
Dasomeone, line 40: Should `lokiDatasourceName` and `lokiDatasourceLabel` not also be mentioned here? :)

Dasomeone (commented):
Love the concept! Just one minor comment, but otherwise looks great

common-lib/common/signal/README.md
v-zhuravlev: Loki can be renamed in variables, a more generic function. And this is a readme for signals, where only prometheus could be renamed.

Dasomeone (approved):
LGTM, thanks for explaining :)

PR #1241, opened 2024-05-30
Fix windows fleet table

v-zhuravlev:
This improves main table regex. Ensures to drop duplicated columnts with unuseful names like `job 10`, `job 11`, etc if number of metrics in table reaches that number.


BominRahmani (approved):
LGTM!

PR #1244, opened 2024-05-31
Add + timeSeries.queriesSelector.withMaxDataPoints(100) to logslib

v-zhuravlev:
This controls max number of bars on the logs lib dashboard. Otherwise, they can become too thin on large intervals, putting too heavy load on backend.

### DIFF
--- a/logs-lib/logs/panels.libsonnet
+++ b/logs-lib/logs/panels.libsonnet
@@ -27,8 +27,6 @@ function(
       + custom.withDrawStyle('bars')
       + custom.stacking.withMode('normal')
       + custom.withFillOpacity(50)
-      // should be set, otherwise interval is around 1s by default
-      + timeSeries.queryOptions.withInterval('30s')
       + options.tooltip.withMode('multi')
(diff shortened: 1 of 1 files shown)


logs-lib/logs/panels.libsonnet
```diff
@@ -23,11 +23,10 @@ function(
       + timeSeries.queryOptions.withDatasource(
         logsVolumeTarget.datasource.type, logsVolumeTarget.datasource.uid
       )
+      + timeSeries.queriesSelector.withMaxDataPoints(100)
       + custom.withDrawStyle('bars')
       + custom.stacking.withMode('normal')
       + custom.withFillOpacity(50)
-      // should be set, otherwise interval is around 1s by default
-      + timeSeries.queryOptions.withInterval('30s')
```
Dasomeone, line 30: Is this no longer required?

v-zhuravlev: yes, interval is calculated based on maxdatapoints now, which works better.

mshahzeb (approved):

PR #1246, opened 2024-05-31
Add reasonable 'for' into Harpoxy alerts.

v-zhuravlev:
Generally, all alerts should have some grace period.  Current alerts with small for could event fire on system boot.


nikosmeds (commented):
I suspect due to the `rate(metric[5m])` expressions this will not work as you're expecting. If there's a single increase, then `rate(metric[5m])` will show a non-zero value for 5 minutes since we're measuring the last 5 minutes of data.

Therefore there's really no different between a pending duration `5s` or `5m`, besides how fast the alert will move from pending to firing. Though let me know if I'm mistaken.

To alert if the values are increasing multiple times within the time window, I believe we need to increase the duration to more than `5m`.

v-zhuravlev:
I think you are right. Increaseing to 10m then.

mshahzeb (approved):

Dasomeone (approved):

PR #1249, opened 2024-06-03
Fix typo in Dashboards

carkod:
I'm a customer using Grafana Cloud. Found this typo using the MongoDB Dashboards, so here is a quick fix.


Dasomeone (approved):
Thanks for catching this @carkod!

Issue #985, opened 2023-05-14
Apache HTTP codes histogram is not displayed correctly


filips123:
I'm using Grafana Agent v0.33.2 with Apache HTTP integration v0.0.1 which sends metrics and logs to the Grafana Cloud instance.

My problem is that the "logs by HTTP codes" histogram does not correctly display the number of HTTP codes per bucket. It appears that almost all HTTP code buckets have roughly the same number of responses.

![slika](https://github.com/grafana/jsonnet-libs/assets/16626308/d61726fe-c8ef-422d-97b1-94a97e9e2d38)

I think this happens because each histogram bucket also includes all codes below it. For example, in this case, there were actually 541 200-299 responses, three 300-399 responses, and three 400-499 responses. However, on the displayed histogram, all these numbers are summed together.

![slika](https://github.com/grafana/jsonnet-libs/assets/16626308/350a0c25-cb53-41f2-a6d3-49fc3f5a68c0)

I'm not sure, but I think this might also affect the "errors rate" panel in the main dashboard and the "ApacheErrorsRateTooHigh" alert rule. *Edit:* After some very likely false positive "ApacheErrorsRateTooHigh" alerts, I'm pretty sure it also affects that.

![slika](https://github.com/grafana/jsonnet-libs/assets/16626308/85860013-87d3-44d3-a8ca-59fc0cd3cea5)

<details><summary>Grafana Agent Config</summary>
<p>

```yaml
integrations:
  agent:
    enabled: true
    relabel_configs:
      - action: replace
        source_labels: [agent_hostname]
        target_label: instance
    metric_relabel_configs:
      - source_labels: [exported_job]
        target_label: job
      - source_labels: [exported_instance]
        target_label: instance
      - action: labeldrop
        regex: (exported_instance|exported_job)

  node_exporter:
    enabled: true
    relabel_configs:
      - target_label: instance
        replacement: my-hostname

  mysqld_exporter:
    enabled: true
    instance: my-hostname
    data_source_name: "{{grafana_mysql_username}}:{{grafana_mysql_password}}@(localhost:3306)/"
    relabel_configs:
      - source_labels: [__address__]
        target_label: job
        replacement: integrations/mysql

  apache_http:
    enabled: true
    instance: my-hostname
    scrape_uri: "https://my-hostname/server-status?auto"

  prometheus_remote_write:
    - basic_auth:
        password: {{grafana_api_key}}
        username: {{grafana_metrics_id}}
[...]

PR #1253, opened 2024-06-03
Fix apache http mixin

gaantunes:
This is fixing https://github.com/grafana/jsonnet-libs/issues/985.

I identified that for some reason using `label_replace` is causing the histograms to accummulate, even though it is being queried using histogram format, as expected. This is basically moving the relabel to a transformation within the dashboard.

### DIFF
--- a/apache-http-mixin/dashboards/apache-exporter-full.libsonnet
+++ b/apache-http-mixin/dashboards/apache-exporter-full.libsonnet
@@ -882,14 +882,14 @@ local errorsPanel =
       },
     ],
     transformations: [
-        {
-          id: "renameByRegex",
-          options: {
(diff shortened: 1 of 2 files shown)


v-zhuravlev:
Are changes applied to jvm related?

v-zhuravlev:
Hm, are you sure that moving label_replace to transform fixes it?

gaantunes:
> Hm, are you sure that moving label_replace to transform fixes it?

Yep it does.

image

image


> Hm, are you sure that moving label_replace to transform fixes it?

v-zhuravlev (approved):

Dasomeone (approved):
Nice catch :)

PR #1255, opened 2024-06-05
[jaeger-agent-mixin] Add a container name resource attribute

ptodev:
The container resource attribute is necessary so that [otelcol.processor.k8sattributes](https://grafana.com/docs/alloy/latest/reference/components/otelcol.processor.k8sattributes/) can enrich the telemetry with container attributes. Without that attribute, it will not know which container within the pod the spans are coming from.

Notice that we are adding the container attribute with OTel semantics, regardless whether `with_otel_resource_attrs` is set to `true`. I'm doing this because I'd like to deprecate `with_otel_resource_attrs` in favour of always using OTel semantics.

This is a public repository, and I hope no one in the general public depends on the fact that this mixin has been using non-OTel semantics.


tpaschalis (approved):
🙈

mattdurham (approved):
LGTM

PR #1258, opened 2024-06-06
Feat: Add vSphere Mixin

StefanKurek:
The following PR adds a vSphere mixin.
The dashboards can be seen [here](https://keithschmitty.grafana.net/dashboards/f/ddmbphlezuku8d/final-vsphere)


vsphere-mixin/variables.libsonnet
```diff
@@ -0,0 +1,232 @@
...
...t_name=~"$vcenter_host_name",vcenter_resource_pool_inventory_path=~"$vcenter_resource_pool_inventory_path",vcenter_resource_pool_inventory_path!=""})),"vm_path","/","vcenter_resource_pool_inventory_path","vcenter_vm_name")) by (vm_path,vcenter_vm_name)) or (sum(label_join(sgn(sum by(vcenter_virtual_app_inventory_path,vcenter_vm_name)(vcenter_vm_memory_usage_mebibytes{job=~"$job",vcenter_datacenter_name=~"$vcenter_datacenter_name",vcenter_cluster_name=~"$vcenter_cluster_name|",vcenter_host_name=~"$vcenter_host_name",vcenter_virtual_app_inventory_path=~"$vcenter_virtual_app_inventory_path",vcenter_virtual_app_inventory_path!=""})),"vm_path","/","vcenter_virtual_app_inventory_path","vcenter_vm_name")) by (vm_path,vcenter_vm_name)) or (sum(label_replace(sgn(sum by(vcenter_resource_pool_inventory_path,vcenter_virtual_app_inventory_path,vcenter_vm_name)(vcenter_vm_memory_usage_mebibytes{job=~"$job",vcenter_datacenter_name=~"$vcenter_datacenter_name",vcenter_cluster_name=~"$vcenter_cluster_name|",vcenter_host_name=~"$vcenter_host_name",vcenter_resource_pool_inventory_path="",vcenter_virtual_app_inventory_path=""})),"vm_path","$1","vcenter_vm_name","(.*)")) by (vm_path,vcenter_vm_name)))',
```
v-zhuravlev, line 60: This looks really complex . Could explain a bit why such query is needed? Also, could you maybe try to define it using multiline syntax so it is a bit easier to read? thanks.

v-zhuravlev: Perhaps, even leaving a comment in code would be great.

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1428 @@
...
+        )
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.standardOptions.withUnit('percent'),
+
+      clustersTable:
+        commonlib.panels.generic.table.base.new(
+          'Clusters table',
+          targets=[
+            t.effectiveCPUClusters + g.query.prometheus.withFormat('table')
+            + g.query.prometheus.withRange(true)
```
v-zhuravlev, line 117: nit: you could use withRefId() here to name each query with meaningful name instead of Value #A-Z. Not required, but easy to follow in latter transforms what is being changed.

vsphere-mixin/variables.libsonnet
StefanKurek: @v-zhuravlev Added a comment here for this

StefanKurek: I tried to split this up into a multiline, but I was having trouble with the linter because of the `query_result` I think

vsphere-mixin/panels.libsonnet
StefanKurek: I changed all of these as well even though I probably didn't need to 😄

vsphere-mixin/variables.libsonnet
```diff
@@ -0,0 +1,236 @@
...
+        std.mapWithIndex(chainVarProto, utils.chainLabels(groupLabels, [filteringSelector])),
+      datasources: {
+        prometheus:
+          var.datasource.new('prometheus_datasource', 'prometheus')
+          + var.datasource.generalOptions.withLabel('Data source')
+          + var.datasource.withRegex(''),
+        loki:
+          var.datasource.new('loki_datasource', 'loki')
+          + var.datasource.generalOptions.withLabel('Loki data source')
+          + var.datasource.withRegex('(?!grafanacloud.+usage-insights|grafanacloud.+alert-state-history).+')
```
v-zhuravlev, line 110: I think this is added automatically on the integration level.

vsphere-mixin/README.md
```diff
@@ -0,0 +1,125 @@
...
+- VSphereDatastoreWarningDiskUtilization
+- VSphereDatastoreCriticalDiskUtilization
+- VSphereHostWarningHighPacketErrors
+
+## vSphere overview
+
+The vSphere overview dashboard provides an overview on CPU, memory, and network throughput details for clusters, resource pools, ESXi hosts, and datastores.
+
+![vSphere overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/vsphere/screenshots/vsphere_overview_1.png)
+![vSphere overview dashboard (hosts)](https://storage.googleapis.com/grafanalabs-integration-assets/vsphere/screenshots/vsphere_overview_2.png)
```
v-zhuravlev, line 26:
```suggestion
![vSphere overview dashboard](https://storage.googleapis.com/grafanalabs-integration-assets/vsphere/screenshots/vsphere_overview_1.png)  
![vSphere overview dashboard (hosts)](https://storage.googleapis.com/grafanalabs-integration-assets/vsphere/screenshots/vsphere_overview_2.png)
```

Looks like line return missing when rendered. Here and for below screenshots.

vsphere-mixin/README.md
```diff
@@ -0,0 +1,125 @@
+# vSphere Mixin
```
v-zhuravlev, line 1:
```suggestion
# vSphere mixin
```

vsphere-mixin/README.md
```diff
@@ -0,0 +1,125 @@
...
+}
+```
+
+In order for the selectors to properly work for system logs ingested into your logs datasource, please also include the matching `job` label onto the [scrape_configs](https://grafana.com/docs/loki/latest/clients/promtail/configuration/#scrape_configs) as to match the labels for ingested metrics.
+
+```yaml
+
+```
+
+## Alerts Overview
```
v-zhuravlev, line 72:
```suggestion
## Alerts overview
```

vsphere-mixin/README.md
```diff
@@ -0,0 +1,125 @@
...
+  alertsHighCPUUtilization: 90,
+  alertsHighMemoryUtilization: 90,
+  alertsWarningDiskUtilization: 75,
+  alertsCriticalDiskUtilization: 90,
+  alertsHighPacketErrors: 20,
+  },
+}
+```
+
+## Install Tools
```
v-zhuravlev, line 96:
```suggestion
## Install tools
```

vsphere-mixin/README.md
```diff
@@ -0,0 +1,125 @@
...
+```bash
+go install github.com/google/go-jsonnet/cmd/jsonnetfmt@latest
+```
+
+The files in `dashboards_out` need to be imported
+into your Grafana server. The exact details will be depending on your environment.
+
+`prometheus_alerts.yaml` needs to be imported into Prometheus.
+
+## Generate Dashboards And Alerts
```
v-zhuravlev, line 116:
```suggestion
## Generate dashboards and alerts
```

v-zhuravlev (approved):

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+        barGauge.new(title='Cluster memory limit')
+        + barGauge.queryOptions.withTargets([
+          t.clusterMemoryLimit,
+        ])
+        + barGauge.panelOptions.withDescription('The available memory capacity of the cluster.')
+        + barGauge.options.withOrientation('horizontal')
+        + barGauge.standardOptions.thresholds.withSteps([
+          barGauge.thresholdStep.withColor('super-light-green'),
+        ])
+        + barGauge.standardOptions.withUnit('bytes'),
```
Dasomeone, line 1081: Minor potential issue on these two panels, with the legend effectively only supporting a single entry before you have to scroll.
They might benefit from having their height increased a bit, or changing the legend to a list style 
image

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+
+      hostModifiedMemory:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Modified memory',
+          targets=[t.hostModifiedMemoryBallooned, t.hostModifiedMemorySwapped],
+          description='The amount of memory that has been swapped or ballooned on the ESXi host.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('mbytes')
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.options.legend.withPlacement('right'),
```
Dasomeone, line 505: There's an issue with the legend on this panel, where some entries begin with `-`, which I reckon is due to a missing cluster label?
image

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+        + g.panel.timeSeries.options.legend.withDisplayMode('table'),
+
+      hostMemoryUtilization:
+        commonlib.panels.memory.timeSeries.usagePercent.new(
+          'Memory utilization',
+          targets=[t.hostMemoryUtilization],
+          description='The memory utilization percentage of the ESXi host.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('percent')
+        + g.panel.timeSeries.options.legend.withDisplayMode('table'),
```
Dasomeone, line 495: These four panels have the same squished legend issue, might benefit from list style or increased height:
image

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+
+      hostPacketErrorRate:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Avg packet errors',
+          targets=[t.hostPacketReceivedErrorRate, t.hostPacketTransmittedErrorRate],
+          description='The 20s average of received or transmitted packets dropped over the ESXi hosts network compared to the overall packets received or transmitted.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('percent')
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.options.legend.withPlacement('bottom'),
```
Dasomeone, line 525: Same issue re: squished legends here

vsphere-mixin/panels.libsonnet
Dasomeone, line 505: Personal take for this panel, I'd only want to know if the values are non-zero. Currently as-is it'd include ballooned and swapped memory stats for each hosts, even if 0

vsphere-mixin/panels.libsonnet
Dasomeone, line 525: Similarly here, I'm not sure I'd be interested in anything but non-zero values showing up. @v-zhuravlev how have we addressed things like this in the past?

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+        + g.panel.timeSeries.standardOptions.withUnit('percent'),
+
+      topMemoryUtilizationClusters:
+        commonlib.panels.memory.timeSeries.usagePercent.new(
+          'Top memory utilization by cluster',
+          targets=[t.topMemoryUtilizationClusters],
+          description='The clusters with the highest memory utilization percentage in the datacenter.'
+        )
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.standardOptions.withUnit('percent'),
```
Dasomeone, line 110: Another case of list instead of table for the legend, I think

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+        + g.panel.timeSeries.standardOptions.withUnit('shares'),
+
+      topMemoryShareResourcePools:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Top memory shares by resource pools',
+          targets=[t.topMemoryShareResourcePools],
+          description='The resource pools with the highest amount of memory shares allocated in the datacenter.'
+        )
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.standardOptions.withUnit('shares'),
```
Dasomeone, line 423: These four I'd like to see increased heights for, as table-based legend makes sense. 
It **might** fit as right-aligned, but if not then I think 25% taller would fix it.

e.g. either of the lefthand options
image

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+        + g.panel.timeSeries.standardOptions.withUnit('ms'),
+
+      topPacketErrorRateHosts:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Top packet errors by ESXi hosts',
+          targets=[t.topPacketErrorRateHosts],
+          description='The ESXi hosts with the highest percentage of packet errors in the datacenter.'
+        )
+        + g.panel.timeSeries.options.legend.withDisplayMode('table')
+        + g.panel.timeSeries.standardOptions.withUnit('percent'),
```
Dasomeone, line 459: Same issue for these four.

Any one of these options would suffice:
- List based
- Table (current) with more height
- Table (current) right-aligned

vsphere-mixin/panels.libsonnet
Dasomeone, line 1: Pretty much all the panels throughout the dashboards need to have their legends adjusted as they're a tad too squished.

Have left a few comments already, but to recap:

They're currently using table based, bottom-aligned.

I think they'd benefit from either:
- List based, bottom aligned
- Table based, bottom aligned, but with being taller panels (at least 25-50%)
- Table based, right aligned (depends on the length of the legend text)

It likely will be a case by case basis, but please have a look :)

vsphere-mixin/panels.libsonnet
```diff
@@ -0,0 +1,1472 @@
...
+          + fieldOverride.byName.withProperty('custom.displayMode', 'gradient-gauge')
+          + fieldOverride.byName.withProperty('custom.align', 'left')
+          + table.fieldOverride.byName.withProperty('custom.width', 157)
+          + fieldOverride.byName.withPropertiesFromOptions(
+            table.standardOptions.withUnit('percent')
+            + table.standardOptions.color.withMode('continuous-BlPu')
+            + table.standardOptions.withMin(0)
+            + table.standardOptions.withMax(100)
+            + table.standardOptions.withDecimals(1)
+          ),
```
Dasomeone, line 1396: The gradient for the disk utilization panel feels a little overly aggressive, and personally I'm not sure about the colour scheme.

I'd expect something more akin to green -> yellow -> red, rather than blue with interrupted yellows, but I think it's just a case of playing about a little

image

Dasomeone (commented):
Reviewed the dashboards

vsphere-mixin/panels.libsonnet
StefanKurek: @Dasomeone these are actually standalone hosts (won't have a cluster). I don't love this legend either....but I couldn't figure out if there was some "conditional" that you could do here.

Dasomeone: It's not possible based on metric inspection, no, but in this case I think it might be appropriate if we simply move the cluster part to be appended rather than prefixed?

Not perfect, but would be a little cleaner.

Alternatively, we could drop the cluster part from the legend entirely and rely on the selector dropdowns

vsphere-mixin/panels.libsonnet
StefanKurek: @Dasomeone I've definitely had some panels where the queries had a != 0 in them. If an object is not consistently meeting criteria (in this case dropping packets) it can look spotty (you might just have a little positive blip on the graph) if we go this route though. Lines would only be consistent and smooth for hosts consistently dropping packets (maybe ok?)

vsphere-mixin/panels.libsonnet
StefanKurek: @Dasomeone same as what I mentioned for dropped packets (although for this graph we would probably be safer in the lines staying consistent. I feel pretty comfortable changing this here).

StefanKurek: I'll change this for the VM panel as well.

vsphere-mixin/panels.libsonnet
StefanKurek: @Dasomeone the green to red actually seems more aggressive iMO haha? 
image
I think we've used blue to red on the last few integrations. I believe it's actually mentioned in the style guide as well for gradients.

vsphere-mixin/panels.libsonnet
Dasomeone: That's a good point for the error rate for sure. It probably wouldn't be quite so jarring if the panel type was swapped to a bar-chart, however?

vsphere-mixin/panels.libsonnet
Dasomeone: In that case please ignore my message! Clearly I need to double check our own docs :sweat_smile:

vsphere-mixin/panels.libsonnet
Dasomeone: Sounds good, thank you Stefan! As mentioned in the other comment, don't be afraid to swapping to a bar chart or keep it you think the interrupted lines would be too frequent.

vsphere-mixin/variables.libsonnet
Dasomeone: Just a note here, it is, but it then breaks when using Grizzly, so something we all need to be aware of when reviewing dashboards :)

Dasomeone (approved):
Tentative approval from my side, but would prefer @v-zhuravlev and @rgeyer give the final sign-off

vsphere-mixin/panels.libsonnet
StefanKurek: It's true...but it's also nice to see some history of errors 😄

PR #1267, opened 2024-06-14
[jaeger-agent-mixin] Align with OTel semantics, add additional Jaeger params.

ptodev:
Updating this public repo so that the resource attributes which `jaeger-agent-mixin` uses for enrichment of traces align better with OTel semantics. This PR will also add additional `JAEGER_SAMPLER_TYPE` and `JAEGER_SAMPLER_PARAM` env vars if the mixin is configured to use them.

### DIFF
--- a/jaeger-agent-mixin/jaeger.libsonnet
+++ b/jaeger-agent-mixin/jaeger.libsonnet
@@ -5,6 +5,8 @@ local k = import 'ksonnet-util/kausal.libsonnet';
     cluster: error 'Must define a cluster',
     namespace: error 'Must define a namespace',
     jaeger_agent_host: null,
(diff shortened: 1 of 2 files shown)


ptodev:
Hi, @csmarchbanks - I hope that after this update you will no longer need a custom fork?

ptodev:
@csmarchbanks I'm going to merge this, and we could discuss this further when I raise a PR to update the machine learning cluster :)

csmarchbanks (commented):
Sorry for the delay, I think this is probably fine, we will still have a second configuration that we require due to needing `OTEL_RESOURCE_ATTRIBUTES` instead of `JAEGER_TAGS` on some services.

PR #1220, opened 2024-05-04
Add jvm observ lib

v-zhuravlev:
A. Add JVM observability lib  (and process-observ-lib):

1) with the following signal categories:

- JVM memory (heap,nonheap)
- JVM garbage collection
- JVM buffers 
- JVM threads
- JVM classes
- Hikari pools
- Logback stats
- Also, general process stats (uptime,cpu, memory...) (as separate library)

2) with basic JVM alerts
3) with supporting following metric sources:
- OpenTelemetry https://github.com/open-telemetry/opentelemetry-java-contrib/blob/main/jmx-metrics/docs/target-systems/jvm.md transformed to Prometheus via otelcollector exporter(https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusexporter/README.md)
- Prometheus https://prometheus.github.io/client_java/instrumentation/jvm/
- Springboot(micrometer) metrics collected through Prometheus actuator https://docs.micrometer.io/micrometer/reference/implementations/prometheus.html .

B. Convert jvm-mixin to use jvm-observ-lib.
C. Convert spring-boot-mixin to use jvm-observ-lib.

Requires #1268 to be merged first.

image

PR #1268, opened 2024-06-14
Add support for aggKeepLabels and aggFunction on signal's level

v-zhuravlev:
1) aggKeepLabels - can keep some very specific labels when automatically aggregating on group/instance level
For example, 

...
aggKeepLabels: ['pool],
aggFunction: 'avg),
aggLevel: 'group'
groupLabels: ['job'],

Would wrap gauge type signals into:

```
avg by (job, pool)  (
   <metric>
)
```

instead of just 
```
avg by (job)  (
   <metric>
)
```

This makes it possible to use that label in panel legend, as well as to how different types of aggregations (not only by instance, but also by pool, by disk, by interface, etc...).

example:
https://github.com/grafana/jsonnet-libs/pull/1220/files#diff-e5ed90f7366df479c5f5e38e8059c86a8f026584279804fa25097789214124b6R25

2) aggFunction can now be redefined on signal's level.


rgeyer (approved):
This is a great addition. Thank you for quickly implementing this feedback!

I'll be able to use this in the CSP dashboards I'm working on.

PR #1271, opened 2024-06-26
Fix: vSphere Mixin - Fixes packet alert

StefanKurek:
Fixes the packet error alert for the vSphere mixin. Was accidentally referring to old metric names.


v-zhuravlev (approved):

BominRahmani (approved):
Lgtm!

PR #1272, opened 2024-06-29
Fix: vSphere Mixin - Fixes Multiple Queries

StefanKurek:
Alert fixed to not aggregate both `object` label and already pre-aggregated non `object` label metric values together.

Fixes a number of aggregated queries where aggregation wasn't needed (stopped aggregating `object` label metric values when there was already a pre-aggregated non `object` label metric value).

Also changes a few utilization calculations to use a better metric ("limit" vs "effective") in the denominator (more appropriate and no longer get > 100% values).

Some other minor naming tweaks to panels.

New dashboards should be uploaded [here](https://keithschmitty.grafana.net/dashboards/f/ddmbphlezuku8d/final-vsphere)

### DIFF
--- a/vsphere-mixin/alerts.libsonnet
+++ b/vsphere-mixin/alerts.libsonnet
@@ -72,7 +72,7 @@
           {
             alert: 'VSphereHostWarningHighPacketErrors',
             expr: |||
-              100 * sum without (direction, object) (vcenter_host_network_packet_error_rate{%(filteringSelector)s}) / clamp_min(sum without (direction, object) (vcenter_host_network_packet_rate{%(filteringSelector)s}), 1) > %(alertsHighPacketErrors)s
+              100 * sum without (direction, object) (vcenter_host_network_packet_error_rate{object="",%(filteringSelector)s}) / clamp_min(sum without (direction, object) (vcenter_host_network_packet_rate{object="",%(filteringSelector)s}), 1) > %(alertsHighPacketErrors)s
             ||| % this.config,
             'for': '5m',
             labels: {
--- a/vsphere-mixin/panels.libsonnet
+++ b/vsphere-mixin/panels.libsonnet
@@ -113,17 +113,17 @@ local utils = commonlib.utils;
         commonlib.panels.generic.table.base.new(
           'Clusters table',
           targets=[
-            t.effectiveCPUClusters + g.query.prometheus.withFormat('table')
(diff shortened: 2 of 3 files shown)


Dasomeone (changes requested):
Couple comments, and apologies for the wait given I was out on PTO :)

1) Minor nit, the updated dashboards on the instance have changed UIDs, which broke links.

Not necessarily something to fix, but something to beware for screenshots/reviewers in case it appears broken

2) More importantly, three of the dashboards show no data at all for the past 7 days, while only the overview (with the old UID at that) and cluster dashboards having data:
Specifically, the ones without any data for the past 7 days are:
- vSphere Logs (even when manually selecting the correct datasource)
- vSphere Hosts
- vSphere virtual machines

Unfortunately I can't review it properly in its current state

StefanKurek:
@Dasomeone Data should be good now and links weren't actually broken. There were just duplicate dashboards uploaded so I was modifying the UIDs. All should be good now.

StefanKurek:
@Dasomeone so the only panels which have a table legend ATM are the top ones in the overview dashboard because they have a height of 8 vs. the standard 6 that is everywhere else. All other panels are actually using the list legend, although it may not be noticeable as there aren't enough items to make it obvious.

Initially I believe most of the legends were table legends, but they had a chopped off look for all the 6 height panels. You suggested either increasing the heights to 8, or to switching to list legends. The latter ended up making more sense to me.  Let me know if you want to revisit this across the board.

Dasomeone:
> Initially I believe most of the legends were table legends, but they had a chopped off look for all the 6 height panels. You suggested either increasing the heights to 8, or to switching to list legends. The latter ended up making more sense to me. Let me know if you want to revisit this across the board.

Ahh, thank you for reminding me!

I think the ideal solution sort of differs a little based on the expected entry length, like if it's a few words long then a list is fine, but the example above contain quite complex path-like legends that are difficult to see the ends of.

Specifically for the VM dashboard I'd like to see either 8 height and table based, because you're right it doesn't look good cut off at 6 height unfortunately. :( 

@v-zhuravlev and @rgeyer what do you think in this particular case?

v-zhuravlev (approved):
LGTM

v-zhuravlev:
As of table vs list - I lean towards list as the default choice for all panels where you don't show aggregated values in legend.

Dasomeone (approved):

PR #1274, opened 2024-07-04
Add helper function for adding asserts specific variables

mshahzeb:
Adds a helper function to add asserts specific variables at the integration level.

### DIFF
--- a/grafana-cloud-integration-utils/util.libsonnet
+++ b/grafana-cloud-integration-utils/util.libsonnet
@@ -211,6 +211,7 @@ local integration_version_panel(version, statusPanelDataSource, height, width, x
   // Adds asserts specific variables to the dashboards
   add_asserts_variables(dashboard, config)::
     if std.member(config.statusPanelsDashboardsMetrics, dashboard.title) then
+      dashboard
       {
         templating: {
           list: [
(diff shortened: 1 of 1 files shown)


v-zhuravlev (approved):

grafana-cloud-integration-utils/util.libsonnet
```diff
@@ -207,6 +208,71 @@ local integration_version_panel(version, statusPanelDataSource, height, width, x
...
+            var.query.new('env')
+            + var.query.withDatasource('prometheus', config.statusPanelsDatasourceNameMetrics)
+            + var.query.queryTypes.withLabelValues(
+              'asserts_env',
+              'asserts:mixin_workload_job',
+            )
+            + var.query.generalOptions.withLabel('Asserts environment')
+            + var.query.selectionOptions.withIncludeAll(
+              value=true,
+              customAllValue='.+'
```
v-zhuravlev, line 226: One question, how would it handle situations when there is no asserts_env label at all? (for cases where integration ared used without asserts?) 
Would `.*` be better?

mshahzeb: Yes as suggested by you on the call today, `.*` would be better. Updating.

v-zhuravlev: LGTM

PR #1278, opened 2024-07-11
Extend csplib to add Azure SQL Database

rgeyer:
Adds an azure sql database dashboard to csp-mixin.

image

image


csp-mixin/panels.libsonnet
```diff
@@ -268,5 +268,120 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+          this.signals.azuresqldb.dtuPercent.asTarget()
+          + g.query.prometheus.withFormat('table')
+          + g.query.prometheus.withInstant(true),
+
+          this.signals.azuresqldb.dtuLimit.asTarget()
+          + g.query.prometheus.withFormat('table')
+          + g.query.prometheus.withInstant(true),
+        ],
+        'DTU utilization and limits by database'
+      ) + self._asql_tableCommon(),
```
v-zhuravlev, line 385: nit: you could also try to define tables with using asTable() and asTableColumn() now:
https://github.com/grafana/jsonnet-libs/pull/1290

csp-mixin/signals/azuresqldb.libsonnet
```diff
@@ -0,0 +1,139 @@
...
+        description: 'Percent of CPU utilization by database. If database uses vCores, the vCore percent utilization is shown',
+        type: 'gauge',
+        unit: 'percent',
+        sources: {
+          azuremonitor: {
+            expr: '100 * (label_replace(azure_microsoft_sql_servers_databases_cpu_used_average_count{%(queriesSelector)s}, "database", "$1", "resourceID", ".+/(.*)") / label_replace(azure_microsoft_sql_servers_databases_cpu_limit_average_count{%(queriesSelector)s}, "database", "$1", "resourceID", ".+/(.*)"))',
+            legendCustomTemplate: '{{database}} vCore',
+          },
+        },
+      },
```
v-zhuravlev, line 74: type=gauge  and aggLevel=none is equivalent of type=raw. Right now it would just fine, however, changing aggLevel to instance/group later on would break this signal, as 100 * would be applied in the wrong place.
Best to change type to raw, or move `100 *` to `exprWrappers`:
https://github.com/grafana/jsonnet-libs/tree/master/common-lib/common/signal#configuration-options

v-zhuravlev (approved):

csp-mixin/signals/azuresqldb.libsonnet
rgeyer: Ahh excellent. Thank you for this, I'll add it to the exprWrappers.

PR #1279, opened 2024-07-11
Add SNMPRestarted alert

mshahzeb:
Adds an alert to detect SNMP restart.

### DIFF
--- a/snmp-mixin/alerts/alerts.libsonnet
+++ b/snmp-mixin/alerts/alerts.libsonnet
@@ -28,6 +28,18 @@
             },
             'for': '5m',
           },
+          {
+            alert: 'SNMPRestarted',
+            expr: 'sysUpTime < 60',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'SNMP Target Restarted',
+              description: 'The SNMP target on instance {{$labels.instance}} from job {{$labels.job}} has restarted (sysUpTime < 60 seconds).',
+            },
+            'for': '1m',
+          },
         ],
       },
     ],


v-zhuravlev:
I would increase it to 300 as you might miss that small uptime, due to scrape frequency (once in a minute?)

snmp-mixin/alerts/alerts.libsonnet
```diff
@@ -28,6 +28,18 @@
             },
             'for': '5m',
           },
+          {
+            alert: 'SNMPRestarted',
+            expr: 'sysUpTime < 60',
+            labels: {
+              severity: 'warning',
```
v-zhuravlev, line 35:
```suggestion
              severity: 'warning',
```
I would set it to info, as you don't know if restart is planned or not. Also, boot from power off state would also fire it, I would set it to info.

snmp-mixin/alerts/alerts.libsonnet
```diff
@@ -28,6 +28,18 @@
...
+            },
+            annotations: {
+              summary: 'SNMP Target Restarted',
+              description: 'The SNMP target on instance {{$labels.instance}} from job {{$labels.job}} has restarted (sysUpTime < 60 seconds).',
+            },
+            'for': '1m',
```
v-zhuravlev, line 41: I'm not sure if there is a point waiting here.

snmp-mixin/alerts/alerts.libsonnet
```diff
@@ -28,6 +28,18 @@
...
             'for': '5m',
           },
+          {
+            alert: 'SNMPRestarted',
+            expr: 'sysUpTime < 60',
+            labels: {
+              severity: 'warning',
+            },
+            annotations: {
+              summary: 'SNMP Target Restarted',
```
v-zhuravlev, line 38:
```suggestion
              summary: 'SNMP target restarted',
```
according to our style guide.

snmp-mixin/alerts/alerts.libsonnet
mshahzeb: Makes sense

snmp-mixin/alerts/alerts.libsonnet
mshahzeb: Corrected

snmp-mixin/alerts/alerts.libsonnet
mshahzeb: Havent completely removed it - added a 5s wait instead

v-zhuravlev (approved):

PR #1282, opened 2024-07-15
Varnish `k8s` plugin support

algchoo:
### Description
This PR updates the Varnish mixin to support the `k8s` plugin

Logs for this integration required sidecar forwarding and having `alloy` read from an actual file on the container's filesystem, rather than reading logs from the container's `stdout`. With this, even though it's `k8s` we're still going to have a `filename` label and it makes sense here.

### Changes
* [add selectors to config and updated dashboard](https://github.com/grafana/jsonnet-libs/commit/e3b9af484f79ad5e128de4faa25d72e194fb7d18)
* [updated overview dashboard filter labels](https://github.com/grafana/jsonnet-libs/commit/0da2899ff6d1743788cdbeb620ce0b134d547c55)
* [update allValues to .+ for job and instance labels](https://github.com/grafana/jsonnet-libs/commit/ddb6429f53bc12af7f0c9d64160f3588f88bce29)
* [removed log_type because filename is present in a k8s context due to the nature of log collection](https://github.com/grafana/jsonnet-libs/commit/68f1375eac90119f083e1572f9dcd1be154c8487)

### DIFF
--- a/varnish-mixin/dashboards/varnish-overview.libsonnet
+++ b/varnish-mixin/dashboards/varnish-overview.libsonnet
@@ -1174,7 +1174,7 @@ local frontendLogsPanel(matcher) = {
     {
       datasource: lokiDatasource,
       editorMode: 'code',
-      expr: '{' + matcher + '} |= `` | (' + frontendLogFilter + ' or log_type="frontend")',
+      expr: '{' + matcher + ', ' + frontendLogFilter + '}',
       queryType: 'range',
       refId: 'A',
     },
@@ -1200,7 +1200,7 @@ local backendLogsPanel(matcher) = {
     {
       datasource: lokiDatasource,
       editorMode: 'code',
-      expr: '{' + matcher + '} |= `` | (' + backendLogFilter + ' or log_type="backend")',
+      expr: '{' + matcher + ', ' + backendLogFilter + '}',
       queryType: 'range',
       refId: 'A',
     },
--- a/varnish-mixin/dashboards/varnish-overview.libsonnet
+++ b/varnish-mixin/dashboards/varnish-overview.libsonnet
@@ -1262,7 +1262,7 @@ local backendLogsPanel(matcher) = {
               refresh=2,
               includeAll=true,
               multi=true,
-              allValues='',
+              allValues='.+',
               sort=0
             ),
             template.new(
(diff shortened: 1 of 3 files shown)


varnish-mixin/dashboards/varnish-overview.libsonnet
```diff
@@ -1258,18 +1262,30 @@ local backendLogsPanel = {
...
+              multi=true,
+              allValues='.*',
+              hide=if $._config.enableMultiCluster then '' else 'variable',
               sort=0
             ),
             template.new(
               'instance',
               promDatasource,
-              'label_values(varnish_main_sessions,instance)',
+              'label_values(varnish_main_sessions{%(multiclusterSelector)s},instance)' % $._config,
```
Caleb-Hurshman, line 1283: This should use the `varnishSelector` rather than `multiclusterSelector`

Caleb-Hurshman (changes requested):
One minor change before I approve

BominRahmani (approved):
Lgtm! (fix that problem caleb pointed out) ✅

varnish-mixin/dashboards/varnish-overview.libsonnet
```diff
@@ -9,6 +9,10 @@ local dashboardUid = 'varnish-overview';
 local promDatasourceName = 'prometheus_datasource';
 local lokiDatasourceName = 'loki_datasource';
 
+local getMatcher(cfg) = '%(varnishSelector)s, instance=~"$instance"' % cfg;
+local frontendLogFilter = 'filename=~"/var/log/varnish/varnishncsa-frontend.*.log|/opt/varnish/log/varnishncsa-frontend.*.log"';
+local backendLogFilter = 'filename=~"/var/log/varnish/varnishncsa-backend.*.log|/opt/varnish/log/varnishncsa-backend.*.log"';
```
Dasomeone, line 14: Just a thought, but do we want to call out the two potential log paths in the readme with a section for k8s setup, along with an explanation for users of the mixin to check that it matches their deployments paths?

If there are no other variations in path than these two, then I think we're good to omit that

Dasomeone (commented):
Mostly LGTM, just a documentation question from my end.

varnish-mixin/dashboards/varnish-overview.libsonnet
algchoo: Where the file(s) are is completely customizable, that's my understanding. As long as the file/directory permissions are set correctly and whatnot, so I think we need to call out these log paths as something to be matched. 👍

Dasomeone (approved):
Perfect, thanks Austin! Forgot to approve in the morning so doing so now

PR #1284, opened 2024-07-17
Update opensearch alerts

v-zhuravlev:
It is common practice to use alertmanager inhibit rules to match rules by alertname, so that warning threshold is supressed when alert with same name and severity critical is present.  Renaming alert names so it easier to use that pattern.  For example, same pattern is used in node_exporter's alerts.


Dasomeone (approved):
Nice catch Vitaly!

PR #1285, opened 2024-07-17
grafana-builder: add support for native/classic stat panel query

krajorama:
Seamlessly allow using statPanel with classic/native queries.


duricanikolic (approved):
LGTM

PR #1291, opened 2024-07-23
Kafka observ lib

v-zhuravlev:
1) Add kafka-observ-lib that currently contains two signal groups:
- topics
- consumerGroups of topics
Together signals' panels can compose the single dashboard called 'Kafka topics overview' that supersedes 'Kafka topics' and 'Kafka lag' dashboards, since it contains all keys metrics of those two.
2) as an aditional feature, this library works with two different versions of JMX config:
    - metricsSource: `grafanacloud`: JMX configs from kafka-mixin/jmx dir
    - metricsSource: `prometheus`: JMX configs from [prometheus/jmx_exporter](https://github.com/prometheus/jmx_exporter/blob/main/example_configs/kafka-2_0_0.yml)

3) Drop json 'Kafka topics' and 'Kafka lag' dashboards from kafka-mixin
4) Add new 'Kafka topics overview' dashboard to kafka-mixin instead

Result:
image


gaantunes (approved):
LGTM, thanks for the changes!

PR #1294, opened 2024-07-29
Add zookeeper-observ-lib

v-zhuravlev:
- Add zookeeper-observ-lib and uses it inside kafka-observ-lib/kafka-mixin
- Drop old zookeeper dashboard from kafka-mixin
- Add support for metricsSource: prometheus_old in jvm-observ-lib. This is metric style used in prometheus instrumentation prior to 1.0.0 version (in zookeeper metrics for example)
- Add support for collapsible rows, thanks to https://github.com/grafana/grafonnet/pull/214

Zookeeper dashboard in Kafka now looks like this:
image

With all JVM related rows imported from jvm-observ-lib but collapsed, to let users focus on zookeeper metrics by default: 
https://github.com/grafana/jsonnet-libs/pull/1220


gaantunes (approved):
LGTM. :rocket:

jvm-observ-lib/signals/hikari.libsonnet
```diff
@@ -9,7 +9,8 @@ function(this)
     aggFunction: 'avg',
     discoveryMetric: {
       java_micrometer: 'hikaricp_connections',  // https://github.com/brettwooldridge/HikariCP/blob/dev/src/main/java/com/zaxxer/hikari/metrics/micrometer/MicrometerMetricsTracker.java
-      prometheus: '?',
+      // prometheus: '?',
+      // prometheus_old: '?',
```
Dasomeone, line 13: Just a question here, if we're not using these discovery metrics, why have they been added to and then commented out?

kafka-observ-lib/config.libsonnet
```diff
@@ -1,16 +1,17 @@
 {
   local this = self,
   filteringSelector: 'job!=""',
-  groupLabels: ['kafka_cluster'],
```
Dasomeone, line 4: I'm not 100% on this change. `cluster` is a reserved label for use with kubernetes clusters, which is why we changed this to `kafka_cluster` in the past to support multi-cluster kafka separately from kubernetes clusters.

kafka-observ-lib/config.libsonnet
```diff
@@ -1,16 +1,17 @@
...
   uid: 'kafka',
-  dashboardNamePrefix: 'Kafka',
+  dashboardNamePrefix: '',
   dashboardTags: ['kafka'],
   alertKafkaLagTooHighThreshold: '100',
   alertKafkaLagTooHighSeverity: 'critical',
   metricsSource: 'prometheus',  //or grafanacloud
   //Can be regex:
   topicsIgnoreSelector: '__consumer_offsets',
+  zookeeperEnabled: false,
```
Dasomeone, line 14: Do we not want to enable Zookeeper by default?

Dasomeone (commented):
Couple comments, but nothing major :)

kafka-observ-lib/config.libsonnet
v-zhuravlev: you are right, changing.

jvm-observ-lib/signals/hikari.libsonnet
v-zhuravlev: wanted to highlight that haven't found them. Removing for cleanliness.

PR #1295, opened 2024-08-03
Fix the wrong utils.ncHistogramLeRate implementation

duricanikolic:
This PR fixes the wrong `utils.ncHistogramLeRate()` implementation. Namely, as stated in the native histogram [documentation](https://grafana.com/docs/mimir/next/visualize/native-histograms/#find-rate-of-observations), it is necessary to sum inside the `histogram_fraction` and not outside it. More precisely, this PR replaces 
```
sum by (cluster, namespace) (
  histogram_fraction(0, 0.1, rate(request_duration_seconds{cluster="cluster1", job="job1"}[$__rate_interval]))
 *
  histogram_count(rate(request_duration_seconds{cluster="cluster1", job="job1"}[$__rate_interval]))
)
```
with
```
histogram_fraction(0, 0.1, sum by (cluster, namespace) (
  rate(request_duration_seconds{cluster="cluster1", job="job1"}[$__rate_interval]))
)
*
histogram_count(sum by (cluster, namespace) (
  rate(request_duration_seconds{cluster="cluster1", job="job1"}[$__rate_interval]))
)
```

Moreover, this PR fixes the behaviour of `utils.ncHistogramLeRate` in case of `le=+Inf`. 
In case of the classic histogram, it used to generate the following regexp:
```
le=~"+Inf"
```
instead of 
```
le=~"\\+Inf"
```

PR #1297, opened 2024-08-05
Fix utils.ncHistogramLeRate formatting issues

duricanikolic:
In #1295  and #1296 we fixed some problems related to `utils.ncHistogramLeRate()`, but we also changed the query formatting, which doesn't get along well in combination with other queries.
This PR reverts the previous formatting by keeping the semantical changes introduced by the previous 2 PRs.


krajorama (approved):
LGTM, I see only \n removed, no other change

PR #1301, opened 2024-08-13
feat: Tomcat log expression update

algchoo:
### Description

The main purpose of this PR is to update the log expression for Apache Tomcat to accept either the `filename` or `log_type` label. The intention is to allow users running tomcat on k8s to not have to input an arbitrary value for `filename` that may lead to confusion, instead the `log_type` label has a static value users can just copy/paste.

I noticed that we accidentally created two separate mixins. One called `apache-tomcat-mixin` and the other named `tomcat-mixin`. The k8s plugin support changes were actually only applied to `tomcat-mixin` which is not what the integration points to by default. I've removed the extra mixin and made updates accordingly.

### Changes
* [update the logsPanel expression to support log_type label](https://github.com/grafana/jsonnet-libs/commit/c5193cf3be7fe5d310c62bcbd5c1f6c98b77e401)
* [k8s support for tomcat](https://github.com/grafana/jsonnet-libs/commit/4ecc44f6d0094de98e5c88e2f7f6a580912c6be0)
* [deleted extra tomcat mixin](https://github.com/grafana/jsonnet-libs/commit/96ca4e63c7eef2da4e544af56142f1c56081a117)
* [updated hosts dashboard for k8s support](https://github.com/grafana/jsonnet-libs/commit/acc935f156b87dd9f652dc4e7509d33e598e54ff)
* [revert removal of additional tomcat mixin](https://github.com/grafana/jsonnet-libs/commit/b8f3bd37204891146224f1021f3dda2dc172e253)
* [updated the tomcat mixins to match](https://github.com/grafana/jsonnet-libs/commit/b30acc10503e37cc13b266cf97f84dabe4a3ec84)
* [confirmed that extra tomcat mixin can be removed](https://github.com/grafana/jsonnet-libs/commit/571fe5bb5d1819e4b0f344711dc8614a983dedf2)


Caleb-Hurshman (approved):
Nice job ✅

apache-tomcat-mixin/dashboards/apache-tomcat-overview.libsonnet
```diff
@@ -160,7 +164,7 @@ local cpuUsagePanel = {
           },
         ],
       },
-      unit: 'percentunit',
+      unit: 'percent',
```
StefanKurek, line 167: was this change necessary? just checking

algchoo: @StefanKurek When I was comparing the two mixins side-by-side, I saw that this was `percent` in the other mixin and `percentunit` in the main one, I wasn't certain whether or not this change was necessary, but it felt right. Do you know what `percentunit` is and is it different from `percent`? I was basically just guessing that it was a typo or something.

StefanKurek: @algchoo i "think" one is for a raw percent and one is for a number 0-1 which will then show as a percent. so next i would ask if the original caused percentages that seemed too large (bigger than 100). that would be a sign that you need to switch.

algchoo: @StefanKurek Just tested this out and it probably definitely needs to be `percentunit`, thanks for calling this out 🤠

StefanKurek (approved):

apache-tomcat-mixin/dashboards/apache-tomcat-overview.libsonnet
```diff
@@ -670,7 +674,7 @@ local logsPanel = {
   grafanaDashboards+:: {
     'apache-tomcat-overview.json':
       dashboard.new(
-        'Apache Tomcat overview',
+        'Apache Tomcat Overview',
```
Dasomeone, line 677: Please revert this change, our style guide says to only capitalise the first word and/or product names :)

algchoo: Will do, sorry about that!

Dasomeone (approved):

PR #1312, opened 2024-08-20
nginx-directory: configurable log_format

colega:
Sometimes the log format of the nginx directory doesn't contain enough information, and sometimes it contains useless information.

For example, $remote_addr is often the gateway that forwards the request, and $remote_user is empty when running behind an oauth proxy.

This makes log_format configurable, so each deployment can adjust it to their needs (for example, to read the forwarded oauth user header).

I'm replacing the multiline definition of this with a single line, which cause a non-zero diff even with the default config.


jvrplmlmn (approved):
👍

PR #1313, opened 2024-08-20
Fix consumerGroup metrics in case of grafana exporter fork

v-zhuravlev:
consumergroup lag name is different.


gaantunes (approved):
LGTM.

PR #1316, opened 2024-08-22
Docker mixin: Adds root cgroup filter to filter out root cgroup stats

mshahzeb:
Adds a root cgroup filter - otherwise metrics get inflated when metrics are counted which do not have a `name` label.


v-zhuravlev (approved):
Good call, thanks. 
I think you can also use: 
`id=~"/system.slice/docker.+"` to make sure metrics are related to docker, not cgroups?

hairyhenderson (approved):
Thank you!

PR #1317, opened 2024-08-23
commonlib: Support combined metricsSource

v-zhuravlev:
Add support for using multiple metricsSource at once by providing list instead of string in `unmarshallJsonMulti`:

```
local jsonSignals =  {
    aggLevel: 'group',
    groupLabels: ['job'],
    instanceLabels: ['instance'],
    filteringSelector: 'job="integrations/agent"',
    discoveryMetric: {
      otel: 'up2',
      prometheus: 'up3',
    },
    signals: {
      abc: {
        name: 'ABC',
        type: 'gauge',
        description: 'ABC',
        unit: 's',
        sources: {
          otel: {
            expr: 'abc2{%(queriesSelector)s}',
          },
          prometheus: {
            expr: 'abc{%(queriesSelector)s}',
          },
        },
      },
      bar: {
        name: 'BAR',
        type: 'counter',
        description: 'BAR',
        unit: 'ns',
        sources: {
          otel: {
            expr: 'bar{%(queriesSelector)s}',
          },
          prometheus: {
            expr: 'bar{%(queriesSelector)s}',
          },
        },
      },
    },
  };

local signals = signal.unmarshallJsonMulti(jsonSignals, type=['otel', 'prometheus']);
```

1) PromQL expressions from different sources are combined using `or` (duplicates dropped)
1) PromQL expressions used in variables discovery are combined as well using `__name__=~"metric1|metric2"` (duplicates dropped)
1) ValueMappings are combined
1) LegendCustomTemplate: the first used
1) InfoLabels are combined (duplicates dropped)

To implement this change, some refactoring was required(thus large files diff), but according to tests it hasn't broke anything. 

Note that there are breaking changes of internal addSignal() function, which was never used directly AFAIK (some args moved to sourceMaps arg. UnmarshallJsonMulti() is used instead, which is kept the same


rgeyer (approved):
This looks great! Thanks for the quick implementation.

PR #1320, opened 2024-08-29
Fix kafka alerts

v-zhuravlev:
to support combined signals


rgeyer (approved):
Looks good. Is this meant to still be a draft?

PR #1322, opened 2024-08-30
define new gcp and azure load balancer dashboard in csp-mixin

yduartep:
define new gcp load balancer dashboard in csp-mixin

Check generated dashboards here:

- [GCP Load Balancing](https://awsintegrationrevamp.grafana-dev.net/d/gcp-loadbalancer-test/gcp-load-balancing)
- [Azure Load Balancing ](https://awsintegrationrevamp.grafana-dev.net/d/azure-loadbalancer-test/azure-load-balancing)


csp-mixin/signals/loadbalancer.libsonnet
```diff
@@ -0,0 +1,61 @@
+local commonlib = import 'common-lib/common/main.libsonnet';
+function(this)
+  {
+    local s = self,
+    filteringSelector: this.filteringSelector,
+    groupLabels: this.groupLabels,
+    instanceLabels: this.instanceLabels,
```
rgeyer, line 7: This is where you can override the labels which are used. Take a look at [this documentation](https://github.com/grafana/jsonnet-libs/tree/master/common-lib/common/signal#configuration-options) for details, and I'm happy to answer any questions that come up when you start working with it.

csp-mixin/dashboards.libsonnet
```diff
@@ -27,6 +27,28 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+          for v in variables
+        ])
+        + g.dashboard.withPanels(
+          g.util.grid.wrapPanels(
+            csplib.grafana.rows.glb_requests
+          )
+        ),
+    }
+    +
     if csplib.config.uid == 'azure' then
```
rgeyer, line 52: Since the GCP loadbalancer dashboard is GCP only (for now) we'll need to add a conditional like this for dashboards which are GCP only.

The first dashboard, for blob storage, is Azure and GCP. So when you append your new dashboard to that list, it expects that there will be signals for both sources.

I can add a commit which makes this modification.

rgeyer (commented):
Couple comments which should hopefully answer the questions you had.

These are exactly the reason(s) why I need to do some organization, and possibly break this into a few different repositories/mixins. At the moment there are some odd workarounds for things that I did not initially plan for.

csp-mixin/signals/loadbalancer.libsonnet
yduartep: @rgeyer do you know how can I define a different list of `groupLabels` depending of the metric being used? For example the following metrics not contains the **client_country** label:

- stackdriver_https_lb_rule_loadbalancing_googleapis_com_https_backend_response_bytes_count
- stackdriver_https_lb_rule_loadbalancing_googleapis_com_https_backend_request_bytes_count

I would like to define a different array depending of the metric but no idea how to access that value.

v-zhuravlev: @yduartep , in signal itself you can add `aggKeepLabels` to keep extra labels for specific signals. This affects aggregation if you use `%(agg)s`  template for example or type != raw. See docs above for details.

In general, If you want different groupLabels then you can split signals file into two and define groupLabels separately.

yduartep: Hi @v-zhuravlev and @rgeyer I could define multiple signals but then I want to have just one dashboard composed of multiple signals because the filters are different on a couple of panels defined on the same dashboard.

Do you know if in the `main.libsonnet` file I can define my dashboard concatenating or merging multiple signals like:

```
 loadbalancer: commonlib.signals.unmarshallJsonMulti(this.config.signals.loadbalancer, type=this.config.metricsSource) + commonlib.signals.unmarshallJsonMulti(this.config.signals.loadbalancerOther, type=this.config.metricsSource),
```

Currently, I am defining the filters manually just for the panel that have different list of filters and is working like on the next example. I don't know if is a good practice:

```
errorRate: {
        name: 'Error rate visualization',
        description: 'Error rate visualization',
        type: 'raw',
        sources: {
          stackdriver: {
            expr: 'sum(stackdriver_https_lb_rule_loadbalancing_googleapis_com_https_request_count{job="integrations/gcp",job=~"$job",project_id=~"$project_id",client_country=~"$client_country",backend_target_name=~"$backend_target_name", response_code_class!="200", response_code_class!="0"}) / sum(stackdriver_https_lb_rule_loadbalancing_googleapis_com_https_request_count{%(queriesSelector)s})',
            legendCustomTemplate: 'Error Rate visualization',
          },
        },
      },
```

v-zhuravlev: Try kafka as the reference.

There most of signal groups have instance labels equal to broker level.
However, for topic and consumer groups signals instanceLabel is different:

https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/signals/consumerGroup.libsonnet#L7
instanceLabels: ['topic', 'consumerGroup']
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/signals/topic.libsonnet#L7
instanceLabels: ['topic']


Resulting dashboard uses union of these two for variables:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/dashboards.libsonnet#L7-L12
So you can choose different topics, but can't select consumerGroups.

csp-mixin/azureconfig.libsonnet
```diff
@@ -1,7 +1,7 @@
 {
   _config+:: {
     local this = self,
-    dashboardTags: ['azure'],
+    dashboardTags: ['azure-cloud-provider'],
```
rgeyer, line 4: This will change the tags used by the other azure dashboards that are generated by this lib as well. We'll probably want/need to increment versions appropriately when we integrate it downstream.

csp-mixin/main.libsonnet
```diff
@@ -6,9 +6,8 @@ local commonlib = import 'common-lib/common/main.libsonnet';
     config: config,
     signals:
       {
-        blobstore: commonlib.signals.unmarshallJsonMulti(this.config.signals.blobstore, type=this.config.metricsSource),
-        azureelasticpool: commonlib.signals.unmarshallJsonMulti(this.config.signals.azureelasticpool, type=this.config.metricsSource),
-        azuresqldb: commonlib.signals.unmarshallJsonMulti(this.config.signals.azuresqldb, type=this.config.metricsSource),
+        [sig]: commonlib.signals.unmarshallJsonMulti(this.config.signals[sig], type=this.config.metricsSource)
+        for sig in std.objectFields(this.config.signals)
```
rgeyer, line 10: Nice optimization!

rgeyer (approved):
Looks good!

csp-mixin/azureconfig.libsonnet
yduartep: yeah, As you can see on next images, GCP and Azure use the same tag name.

Screenshot 2024-09-10 at 09 22 55

Screenshot 2024-09-10 at 09 22 48

PR #1328, opened 2024-09-04
Add support for gauge metric in static-exporter

gauravgpta93:
Add support for gauge metric in static-exporter
Keep default as `counter` to avoid any changes for other places

### DIFF
--- a/static-exporter/main.libsonnet
+++ b/static-exporter/main.libsonnet
@@ -35,7 +35,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
             function(acc, metric)
               acc + [
                 '# HELP %(name)s %(description)s' % metric,
-                '# TYPE %(name)s counter' % metric,
+                '# TYPE %(name)s %(metricType)s' % metric,
               ] + [
                 metric.name + value
                 for value in metric.values
@@ -47,13 +47,16 @@ local k = import 'ksonnet-util/kausal.libsonnet';
     }),
 
   metric:: {
-    new(name, description)::
+    new(name, description, metricType="counter")::
       self.withName(name)
-      + self.withDescription(description),
+      + self.withDescription(description)
+      + self.withMetricType(metricType),
 
     withName(name): { name: name },
 
     withDescription(description): { description: description },
+
+    withMetricType(metricType): { type: metricType },
 
     local generateValues(labelMap, value=1) =
       local labels = [


static-exporter/main.libsonnet
```diff
@@ -35,7 +35,7 @@ local k = import 'ksonnet-util/kausal.libsonnet';
             function(acc, metric)
               acc + [
                 '# HELP %(name)s %(description)s' % metric,
-                '# TYPE %(name)s counter' % metric,
+                '# TYPE %(name)s %(metricType)s' % metric,
```
jdbaldry, line 38: This is string formatting with the `metric` object. 

The `self.with<FUNCTION>` indirection can make it a little hard to think about but the `metric` object has three fields, `name`, `description`, and `type`:

```bash
jsonnet -e "(import './main.libsonnet').metric.new('NAME', 'DESCRIPTION', 'TYPE')"

{
   "description": "DESCRIPTION",
   "name": "NAME",
   "type": "TYPE"
}
```

So I think you want:

```suggestion
                '# TYPE %(name)s %(type)s' % metric,
```

jdbaldry (commented):
nit: for consistency with the other `withFUNCTION` methods, perhaps it should be just `withType` and not `withMetricType`?

static-exporter/main.libsonnet
```diff
@@ -47,14 +47,17 @@ local k = import 'ksonnet-util/kausal.libsonnet';
     }),
 
   metric:: {
-    new(name, description)::
+    new(name, description, metricType='counter')::
```
Duologic, line 50: You could set this value below and require `withMetricType` instead of overloading the constructor.

jdbaldry: Is there a reason to prefer that over having all the arguments in the constructor?
I'm not sure I see how `metricType` is different from `description` or `name` except that the library originally only supported `counter` types.

In other words, why have:

```jsonnet
metric.new('NAME', 'DESCRIPTION') + metric.withMetricType('TYPE')
```

when you can just have:

```jsonnet
metric.new('NAME', 'DESCRIPTION, 'TYPE')
```

jdbaldry: Also, to help me understand our patterns better. Why does the constructor do:

```jsonnet
self.withName(name)
+ self.withDescription(description)
+ self.withMetricType(metricType)
```

instead of:

```jsonnet
{
  name: name,
  description: description,
  type: metricType,
}
```

I am thinking it's to use the public API presented by the object so that if `withName` ever does validation, the constructor also does the validation without refactor?
In that case, should we have some convention of indicating that the `type`, `description`, and `type` fields shouldn't be modified except through that public API. Perhaps they should be fields within a hidden `_internal` field or something?

Duologic: > Is there a reason to prefer that over having all the arguments in the constructor?

I'll answer with a question: At what point do we stop adding things to the constructor?

My rule of thumb is to only add arguments to the constructor that are strictly required to create the object. Then create functions for changing the default values of the returned object.

> Why does the constructor do <functions>?

For small objects it doesn't matter that much I guess, for bigger objects these functions may encapsulate larger chunks of code. Opting to ~always use the functions increases consistency. I'm not too nitty-gritty on that, especially for manually written libraries.

There is a benefit on composability. Let's say the user wants to prefix the name for all instances from a library. 

In the object approach, they have to override the constructor:

```jsonnet
local prefixLib =
  lib + {
    new(name, description):
      lib.new(name, description) 
      + { name: 'prefix-' + name },
  };
```

With the function approach:

```jsonnet
local prefixLib =
  lib + {
    withName(name):
      { name: 'prefix-' + name },
  };
```

I prefer to do the latter.

PR #1331, opened 2024-09-13
Add support for otel_with_suffixes metricSource

v-zhuravlev:
It is expected that otel metrics are transformted to Prometheus format via otel-collector/grafana alloy prometheusremotewrite exporter.

This change adds support of popular `add_metric_suffixes` option set to true in addition to false to observ libs.
https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/exporter/prometheusremotewriteexporter/README.md#getting-started


Dasomeone (approved):
All looks reasonable to me, great work Vitaly :D

PR #1332, opened 2024-09-17
Add azure virtual network ddos dashboard to cspmixin

rgeyer:
Adds an azure virtual network dashboard that will visualize the metrics available in that azure monitor namespace.

Unfortunately, those metrics are almost exclusively for DDoS mitigation stats. So we'll likely never see them. Thus, the included screenshot is.. Pretty barren.

It *should* work, in the case of an actual DDoS event...

image

### DIFF
--- a/csp-mixin/dashboards.libsonnet
+++ b/csp-mixin/dashboards.libsonnet
@@ -116,9 +116,9 @@ local commonlib = import 'common-lib/common/main.libsonnet';
                    + g.dashboard.withVariables(variables)
                    + g.dashboard.withPanels(
                      g.util.grid.wrapPanels(
-                      csplib.grafana.rows.vn_overview +
-                      csplib.grafana.rows.vn_bytes +
-                      csplib.grafana.rows.vn_packets
+                       csplib.grafana.rows.vn_overview +
+                       csplib.grafana.rows.vn_bytes +
+                       csplib.grafana.rows.vn_packets
                      )
                    ),
                } else {},
--- a/csp-mixin/dashboards.libsonnet
+++ b/csp-mixin/dashboards.libsonnet
@@ -107,7 +107,7 @@ local commonlib = import 'common-lib/common/main.libsonnet';
 
                  [csplib.config.uid + '-virtualnetwork.json']:
                    local variables = csplib.signals.azureloadbalancer.getVariablesMultiChoice();
-                   g.dashboard.new(csplib.config.dashboardNamePrefix + 'Virtual Network')
+                   g.dashboard.new(csplib.config.dashboardNamePrefix + 'Virtual network')
                    + g.dashboard.withUid(csplib.config.uid + '-virtualnetwork')
                    + g.dashboard.withTags(csplib.config.dashboardTags)
                    + g.dashboard.withTimezone(csplib.config.dashboardTimezone)
--- a/csp-mixin/dashboards.libsonnet
+++ b/csp-mixin/dashboards.libsonnet
@@ -104,5 +104,22 @@ local commonlib = import 'common-lib/common/main.libsonnet';
                        csplib.grafana.rows.alb_loadbalancers
                      )
                    ),
+
+                 [csplib.config.uid + '-virtualnetwork.json']:
+                   local variables = csplib.signals.azureloadbalancer.getVariablesMultiChoice();
+                   g.dashboard.new(csplib.config.dashboardNamePrefix + 'Virtual Network')
+                   + g.dashboard.withUid(csplib.config.uid + '-virtualnetwork')
+                   + g.dashboard.withTags(csplib.config.dashboardTags)
+                   + g.dashboard.withTimezone(csplib.config.dashboardTimezone)
+                   + g.dashboard.withRefresh(csplib.config.dashboardRefresh)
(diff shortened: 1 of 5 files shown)


csp-mixin/dashboards.libsonnet
```diff
@@ -104,5 +104,22 @@ local commonlib = import 'common-lib/common/main.libsonnet';
                        csplib.grafana.rows.alb_loadbalancers
                      )
                    ),
+
+                 [csplib.config.uid + '-virtualnetwork.json']:
+                   local variables = csplib.signals.azureloadbalancer.getVariablesMultiChoice();
+                   g.dashboard.new(csplib.config.dashboardNamePrefix + 'Virtual Network')
```
v-zhuravlev, line 110: Sentence case?

csp-mixin/panels.libsonnet
```diff
@@ -842,5 +842,71 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+              index: 0,
+            },
+            '1': {
+              text: 'Under Attack',
+              color: 'dark-red',
+              index: 1,
+            },
+          },
+        }
+      ),
```
v-zhuravlev, line 864: btw, you can add mappings to signal itself, so it is added to any panel where this signal is attached.

csp-mixin/signals/azurevirtualnetwork.libsonnet
```diff
@@ -0,0 +1,308 @@
...
+        azuremonitor: {
+          expr: 'azure_microsoft_network_virtualnetworks_udpbytesdroppedddos_bytespersecond_maximum{%(queriesSelector)s}',
+          legendCustomTemplate: 'UDP Dropped',
+        },
+      },
+    },
+    
+    udpBytesForwarded: {
+      name: 'UDP bytes Forwarded DDoS',
+      description: 'Inbound udp bytes per second forwarded by DDoS protection.',
```
v-zhuravlev, line 106:
```suggestion
      description: 'Inbound UDP bytes per second forwarded by DDoS protection.',
```

csp-mixin/signals/azurevirtualnetwork.libsonnet
```diff
@@ -0,0 +1,308 @@
...
+    },
+    
+    underDdos: {
+      name: 'Currently DDoS Status',
+      description: 'Is the system currently under DDoS attack or not.',
+      type: 'gauge',
+      sources: {
+        azuremonitor: {
+          expr: 'azure_microsoft_network_virtualnetworks_ifunderddosattack_count_maximum{%(queriesSelector)s}',
+          exprWrappers: [['','OR on() vector(0)']]
```
v-zhuravlev, line 303: nice!

csp-mixin/rows.libsonnet
```diff
@@ -216,5 +216,51 @@ local g = import './g.libsonnet';
...
+        + g.panel.timeSeries.gridPos.withH(6),
+
+        this.grafana.panels.vn_pingmesh_avg
+        + g.panel.timeSeries.gridPos.withW(6)
+        + g.panel.timeSeries.gridPos.withH(6),
+
+        this.grafana.panels.vn_packet_trigger
+        + g.panel.timeSeries.gridPos.withW(12)
+        + g.panel.timeSeries.gridPos.withH(6),
+      ],
```
v-zhuravlev, line 234: btw, you can now define all panels INSIDE row, like so:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/rows.libsonnet#L6-L24

this was, you can decide if you want specific row to be collapsed or not collapsed and render less important rows as collapsed by by using g.util.panel.resolveCollapsedFlagOnRows() func:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/dashboards.libsonnet#L17-L26

csp-mixin/signals/azurevirtualnetwork.libsonnet
```diff
@@ -0,0 +1,308 @@
...
+      sources: {
+        azuremonitor: {
+          expr: 'azure_microsoft_network_virtualnetworks_udppacketsdroppedddos_countpersecond_maximum{%(queriesSelector)s}',
+          legendCustomTemplate: 'UDP Dropped',
+        },
+      },
+    },
+
+    udpPacketsForwarded: {
+      name: 'UDP packets Forwarded DDoS',
```
v-zhuravlev, line 215: TItle or sentence case here?

rgeyer: Yeah, good call. I'll clean these all up.. I'm the worst offender for this rule. Pretty sure I'm the reason it exists ;)

csp-mixin/panels.libsonnet
rgeyer: Nice! I added this, thanks for pointing it out.

csp-mixin/rows.libsonnet
rgeyer: Ahh nice! I don't really need any of these rows collapsed tho. Is there any other obvious reason I should consider the new format?

v-zhuravlev (approved):

PR #1334, opened 2024-09-18
virtual machine dashboard for azure

yduartep:
virtual machine dashboard for azure and gcp


csp-mixin/panels.libsonnet
```diff
@@ -842,5 +842,312 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + this.signals.virtualMachinesOverview.outboundFlows.asPanelMixin()
+      + g.panel.timeSeries.standardOptions.withOverrides([])
+      + g.panel.timeSeries.standardOptions.withUnit('cps')
+      + self._avm_timeSeriesCommon(),
+
+    avm_network_in_by_instance:
+      this.signals.virtualMachines.networkInByVM.asTimeSeries()
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + g.panel.timeSeries.standardOptions.withUnit('decbytes')
```
v-zhuravlev, line 961: nit: you can just set units in signal itself. It will be 'propogated' to any panel where you would render that signal with functions like asTimeSeries(), asStat(), asPanelMixin(), asTableColumn() and so on.

csp-mixin/signals/virtualMachines.libsonnet
```diff
@@ -0,0 +1,141 @@
...
+        description: 'Amount of physical memory, in bytes, immediately available for allocation to a process or for system use in the Virtual Machine',
+        type: 'raw',
+        unit: 'decbytes',
+        sources: {
+          azuremonitor: {
+            expr: 'sum by (resourceName, job, resourceGroup, subscriptionName) (azure_microsoft_compute_virtualmachines_available_memory_bytes_average_bytes{%(queriesSelector)s})',
+            legendCustomTemplate: '{{resourceName}}',
+          },
+        },
+      },
```
v-zhuravlev, line 37: you can avoid using raw everywhere and keep expr clean.
For example:

```suggestion
      availableMemory: {
        name: 'Available memory',
        description: 'Amount of physical memory, in bytes, immediately available for allocation to a process or for system use in the Virtual Machine',
        type: 'gauge
        unit: 'decbytes',
        aggFunction: 'sum',
        sources: {
          azuremonitor: {
            expr: azure_microsoft_compute_virtualmachines_available_memory_bytes_average_bytes{%(queriesSelector)s}',
            legendCustomTemplate: '{{resourceName}}',
            aggKeepLabels: ['resourceName', 'resourceGroup', 'subscriptionName'],
          },
        },
      },
```

this will sum by aggKeepLabels and by dynamically set `groupLabels`, `instanceLabels` like job, instance in the default case. (due to signals group aggLevel=instance)

csp-mixin/signals/virtualMachinesOverview.libsonnet
```diff
@@ -0,0 +1,181 @@
...
+        description: 'The number of bytes sent/received over the network.',
+        type: 'raw',
+        unit: 'decbytes',
+        sources: {
+          azuremonitor: {
+            expr: 'sum(rate(azure_microsoft_compute_virtualmachines_network_in_total_total_bytes{%(queriesSelector)s}[$__rate_interval]))',
+            legendCustomTemplate: 'Received',
+          },
+        },
+      },
```
v-zhuravlev, line 140: just another example how to turn this to type=counter form:
```suggestion
      networkInTotal: {
        name: 'Network throughput send/received',
        description: 'The number of bytes sent/received over the network.',
        type: 'counter',
        unit: 'decbytes',
        sources: {
          azuremonitor: {
            expr: 'azure_microsoft_compute_virtualmachines_network_in_total_total_bytes{%(queriesSelector)s},
            exprWrappers: ['sum(',')'], 
            legendCustomTemplate: 'Received',
          },
        },
      },
```

// by default it will aggregate by aggLevel=instance, so we can add additional wrapper applied at the end, if we want to aggregate even further than instance or group level.

Keeping the expr form simplier would allow also use those expressions inside alerts, by using signal.x.asRuleExpression() function, (where [5m] aggregation would be applied instead of grafana-only $__rate_interval.

csp-mixin/panels.libsonnet
```diff
@@ -842,5 +842,312 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + g.panel.timeSeries.standardOptions.withUnit('short')
+      + self._avm_timeSeriesCommon(),
+
+    avm_disk_total_bytes:
+      this.signals.virtualMachinesOverview.diskReadBytes.asTimeSeries()
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + this.signals.virtualMachinesOverview.diskWriteBytes.asPanelMixin()
+      + g.panel.timeSeries.standardOptions.withOverrides([])
+      + g.panel.timeSeries.standardOptions.withUnit('decbytes')
```
v-zhuravlev, line 931: You don't need this, as units are automatically applied from signals when asTimeseries() or asPanelMixin() are used.

csp-mixin/signals/virtualMachines.libsonnet
```diff
@@ -0,0 +1,141 @@
...
+        unit: 'short',
+        sources: {
+          azuremonitor: {
+            expr: 'sum by (resourceName, job, resourceGroup, subscriptionName) (azure_microsoft_compute_virtualmachines_cpu_credits_consumed_average_count{%(queriesSelector)s})',
+            legendCustomTemplate: '{{resourceName}}',
+          },
+        },
+      },
+      cpuCreditsRemaining: {
+        name: 'CPU Credits Remaining',
```
v-zhuravlev, line 51: Best if we keep our titles consistent. Here is the Title Case, while next signal has Sentence case name. According to grafana writers guide, Sentence case should be used everywhere.

csp-mixin/signals/virtualMachines.libsonnet
```diff
@@ -0,0 +1,141 @@
+local commonlib = import 'common-lib/common/main.libsonnet';
+function(this)
+  {
+    local s = self,
+    filteringSelector: this.filteringSelector,
+    groupLabels: this.groupLabels,
+    instanceLabels: this.instanceLabels,
+    aggLevel: 'instance',
```
v-zhuravlev, line 8:
```suggestion
    filteringSelector: this.filteringSelector,
    groupLabels: this.groupLabels,
    instanceLabels: this.instanceLabels,
    aggLevel: 'instance',
    aggKeepLabels: ['resourceName', 'resourceGroup', 'subscriptionName'],
```

nit: if you change all types in this group from 'raw' to 'gauge','counter', 'histogram' etc, then you can actually set aggKeepLabels here instead of defining it in every signal. Any automatic aggregations would include those common labels then.

v-zhuravlev: more on this https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/README.md

csp-mixin/panels.libsonnet
```diff
@@ -842,5 +842,312 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+              job: '',
+              resourceGroup: 'Group',
+              resourceName: 'Instance',
+            },
+          },
+        },
+      ]),
+
+    avm_top5_disk_read_write:
+      this.signals.virtualMachinesOverview.top5DiskRead.common
```
v-zhuravlev, line 1078: Just FYI, here is another example how you can compose tables with signals:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/panels/topic.libsonnet#L15-L41

Extra transformation was very specific to topics table and generally not needed. You can just add multiple asTableColumn() and that should work.

csp-mixin/panels.libsonnet
```diff
@@ -842,5 +842,312 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+                    '1': {
+                      color: 'green',
+                      index: 0,
+                      text: 'Available',
+                    },
+                  },
+                  type: 'value',
+                },
+              ],
+            },
```
v-zhuravlev, line 892: nit:you can define valueMapping in the signal itself. Check signals lib Readme.

csp-mixin/signals/virtualMachines.libsonnet
yduartep: Love it. I never use the other signals **type** because I don't know which information they add to the expression. I will change it in the rest of the queries. thanks

csp-mixin/panels.libsonnet
yduartep: I tried to config **valueMappings** on the signal but this is what I see:

Screenshot 2024-09-20 at 14 44 33

This is the expected behaviour:
Screenshot 2024-09-20 at 14 45 48

No idea what is doing behind.

v-zhuravlev: got it. You can always check json of the generated panel to see if everything is rendered as expected.

csp-mixin/signals/azurevmOverview.libsonnet
```diff
@@ -0,0 +1,213 @@
...
+        unit: 'short',
+        sources: {
+          azuremonitor: {
+            expr: 'avg by(resourceName) (azure_microsoft_compute_virtualmachines_vmavailabilitymetric_average_count{%(queriesSelector)s})',
+            legendCustomTemplate: '',
+          },
+        },
+      },
+      top5CpuUtilization: {
+        name: 'Top 5 Instances - CPU Utilitization',
```
anaivanov, line 39: Can you update these titles please? 
```suggestion
        name: 'Top 5 Instances by CPU Utilitization',
```

csp-mixin/signals/azurevmOverview.libsonnet
```diff
@@ -0,0 +1,213 @@
...
+            expr: 'azure_microsoft_compute_virtualmachines_percentage_cpu_average_percent{%(queriesSelector)s}',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['resourceName'],
+            exprWrappers: [['topk(5,', ')']],
+          },
+        },
+      },
+
+      bottom5MemoryAvailable: {
+        name: 'Top 5 Instances - Less available memory',
```
anaivanov, line 54: For consistency
```suggestion
        name: 'Top 5 Instances by Less available memory',
```

csp-mixin/signals/azurevmOverview.libsonnet
```diff
@@ -0,0 +1,213 @@
...
+            expr: 'azure_microsoft_compute_virtualmachines_available_memory_bytes_average_bytes{%(queriesSelector)s}',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['resourceName'],
+            exprWrappers: [['bottomk(5,', ')']],
+          },
+        },
+      },
+
+      top5DiskRead: {
+        name: 'Top 5 Instances - Disk read bytes',
```
anaivanov, line 70:
```suggestion
        name: 'Top 5 Instances by Disk read bytes',
```

csp-mixin/signals/azurevmOverview.libsonnet
```diff
@@ -0,0 +1,213 @@
...
+            expr: 'azure_microsoft_compute_virtualmachines_disk_read_bytes_total_bytes{%(queriesSelector)s}',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['resourceName'],
+            exprWrappers: [['topk(5,', ')']],
+          },
+        },
+      },
+
+      top5DiskWrite: {
+        name: 'Top 5 Instances - Disk write bytes',
```
anaivanov, line 86:
```suggestion
        name: 'Top 5 Instances by Disk write bytes',
```

csp-mixin/rows.libsonnet
```diff
@@ -171,96 +171,194 @@ local g = import './g.libsonnet';
...
+        + g.panel.timeSeries.gridPos.withH(8),
+      ],
+
+      avm_instance: [
+        g.panel.row.new('Instance')
+        + g.panel.row.withCollapsed(true)
+        + g.panel.row.withPanels(
+          [
+            this.grafana.panels.azurevm.avm_cpu_utilization
+            + g.panel.timeSeries.gridPos.withW(12)
```
anaivanov, line 322: What do you think of adding a text panel with some short guidance, similar to the GCP dashboard?
Commit from GCP https://github.com/grafana/jsonnet-libs/pull/1336/commits/f1aacad278d399a8c0f6ec8ecc0430b81421d63d

anaivanov (commented):
Looks good! I added some suggestions

csp-mixin/rows.libsonnet
yduartep: sure

csp-mixin/azureconfig.libsonnet
```diff
@@ -14,5 +14,7 @@
     groupLabels: ['job', 'resourceGroup', 'subscriptionName'],
     instanceLabels: ['resourceName'],
     metricsSource: 'azuremonitor',
+    alertAzureVMAvailableMemoryLowThreshold: '367001600',
```
rgeyer, line 17: Is it possible to determine the percentage of used/remaining memory? Usually we prefer to use a percentage, so that it scales with the VM.

Unfortunately, memory thresholds are difficult at times. For certain nodes (work stations, web servers, application nodes) high memory utilization percentage may not be desirable, but for others (caching servers, database servers) using as much memory as possible is desirable.

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,551 @@ local commonlib = import 'common-lib/common/main.libsonnet';
       + g.panel.timeSeries.fieldConfig.defaults.custom.stacking.withMode('normal')
       + this.signals.azurevirtualnetwork.tcpPacketsForwarded.asPanelMixin()
       + this.signals.azurevirtualnetwork.udpPacketsForwarded.asPanelMixin(),
+
+    // Azure Virtual Machines
```
rgeyer, line 903: It looks like all of panels are still defined in this file?

The panels only need to be defined in one place, so this file can just become a single import statement.

Thank you for this re-organization. I realized after the first couple dashboards that we should have done this from the start 😂

rgeyer (approved):
Looks good! Thank you for the re-organization of panels!

csp-mixin/azureconfig.libsonnet
yduartep: I removed this alert and added another one for cpu utilization under 80% and unavailability of the VM.

csp-mixin/panels.libsonnet
yduartep: removed!

PR #1335, opened 2024-09-19
CSP Mixin linting fixes.

rgeyer:
Tweaking a few things on the csp mixin to allow the dashboard linter to pass.

Does the dashboard/mixtool linter not run on mixins in this repo any longer? 🤔


v-zhuravlev (approved):
Perhaps it was shut off when CI moved from drone to github?

PR #1336, opened 2024-09-23
Add GCP Compute Engine dashboard

anaivanov:
Add GCP Compute Engine dashboard


csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,129 @@
...
+    discoveryMetric: {
+      stackdriver: 'stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization',
+    },
+    signals: {
+      // instancesText: {
+      //   name: "Instances",
+      //   type: 'text',
+      //   content: 'Use this section to look at one instance at a time by picking a value in the *Instance* picker at the top of the dashboard.'
+      //   // To Do: check if this works???
+      // },
```
anaivanov, line 18: Is there any text panel type?

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,487 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + this.signals.gcpceOverview.networkReceived.asPanelMixin(),
+    gce_bytes_read_write:
+      this.signals.gcpceOverview.diskBytesRead.asTimeSeries()
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + this.signals.gcpceOverview.diskBytesWrite.asPanelMixin(),
+
+    gce_text_instances:
+      this.signals.gcpce.instancesText.as()
+      + commonlib.panels.generic.text.base.stylize(),
```
anaivanov, line 1355: Same, I couldn't find if there any asText type?

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+        },
+      },
+      tableReadBytes: {
+        name: 'tableReadBytes',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: 'sum by (instance_name, job, project_id)(rate(stackdriver_gce_instance_compute_googleapis_com_instance_disk_read_bytes_count{job!="",job=~".+",project_id=~".+"}[$__rate_interval]))',
+            legendCustomTemplate: '',
+            step: '1m',
```
anaivanov, line 161: In the table that I manually created [here](https://awsintegrationrevamp.grafana-dev.net/d/ddxod2rhnn474c/gcp-google-compute-engine?from=now-6h&to=now&timezone=browser&var-datasource=grafanacloud-prom&var-job=$__all&var-project_id=$__all&var-instance=gke-dev-eu-west-3-main-n2s8-1-b79cd1-bee11d00-54sh&showCategory=Override%207) I use `step: 1`, but I couldn't make this line work the same. Any idea if the prop is named differently?

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+            legendCustomTemplate: 'Total Memory Used',
+          },
+        },
+      },
+
+      packetsSent: {
+        name: 'Total packets count sent/received',
+        description: 'Count of packets sent/received over the network. Sampled every 60 seconds',
+        type: 'counter',
+        unit: 'short',
```
anaivanov, line 208: For some reason, this panel is the only one where the unit is not applied. any idea?

csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,129 @@
...
+      },
+      cpuUsageTime: {
+        name: 'Cpu usage time',
+        description: 'CPU usage, in seconds. For Container-Optimized OS, or Ubuntu running GKE.',
+        type: 'counter',
+        unit: 'seconds',
+        aggFunction: 'avg',
+        sources: {
+          stackdriver: {
+            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_cpu_usage_time{instance_name=~"$instance"}',
```
yduartep, line 53: Why not use all the filters here with `%(queriesSelector)s`? Now it says no data

csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,129 @@
...
+      },
+      networkSent: {
+        name: 'Network sent',
+        description: 'Count of bytes sent from the network. Sampled every 60 seconds.',
+        type: 'raw',
+        unit: 'bytes',
+        aggFunction: 'avg',
+        sources: {
+          stackdriver: {
+            expr: 'rate(stackdriver_gce_instance_compute_googleapis_com_instance_network_sent_bytes_count{instance_name=~"$instance"}[$__rate_interval])',
```
yduartep, line 66: should this be type='counter'and apply `%(queriesSelector)s` filters?

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+      top5CpuUtilization: {
+        name: 'Top 5 Instances by CPU Utilitization',
+        description: 'Fractional utilization of allocated CPU on an instance',
+        type: 'gauge',
+        aggFunction: 'sum',
+        sources: {
+          stackdriver: {
+            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{%(queriesSelector)s}',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['instance_name', 'job', 'project_id'],
```
yduartep, line 46: you don't need to add **job** and **project_id** because they are already added as groupLabel.

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+        },
+      },
+
+      top5DiskWrite: {
+        name: 'Top 5 Instances by Disk write bytes',
+        description: 'List of top 5 instances by disk write bytes',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: 'topk(5, sum by (instance_name, job, project_id) (increase(stackdriver_gce_instance_compute_googleapis_com_instance_disk_write_bytes_count{%(queriesSelector)s}[$__range]))\n)',
```
yduartep, line 70: can you remove \n here?

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+        },
+      },
+
+      top5DiskRead: {
+        name: 'Top 5 Instances by Disk read bytes',
+        description: 'List of top 5 instances by disk read bytes',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: 'topk(5, sum by (instance_name, job, project_id) (increase(stackdriver_gce_instance_compute_googleapis_com_instance_disk_read_bytes_count{%(queriesSelector)s}[$__range]))\n)',
```
yduartep, line 82: also here remove \n

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+        },
+      },
+
+      // Table
+      tableCpuUtilization: {
+        name: 'tableCpuUtilization',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: '100 * sum by (instance_name, job, project_id)(stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{job!="",job=~".+",project_id=~".+"})',
```
yduartep, line 94: don't you want to filter by job or project_id?

yduartep (commented):
I added a fe comments...

Also for consistency with the azure dashboard, can you rename the instance_name, project_id and job column of **Instances** table to be: Instance, Project ID, Job for example.

yduartep:
I don't know if there is something wrong with the table Instances by the values are not being rendered:
Screenshot 2024-09-23 at 18 52 01

yduartep:
On this panel, you have defined `unit=short` but then on the override cell, you have defined. `unit=/s`.
 
Screenshot 2024-09-23 at 18 56 25

yduartep:
On the second query of this panel, you are using a wrong query. It should be **stackdriver_gce_instance_compute_googleapis_com_instance_network_received_bytes_count**?

Screenshot 2024-09-23 at 18 58 58

yduartep:
If I use all the filters on this query, the graph is correctly shown. Now says No data.

Screenshot 2024-09-23 at 19 02 06

yduartep:
Why don't you show Network Received and Network sent on the same line and show CPU Usage time and Cpu utilization on the same line?
Screenshot 2024-09-23 at 19 03 41

csp-mixin/signals/gcpceOverview.libsonnet
rgeyer: Apparently, this is getting overwritten by the [generateUnits](https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/utils.libsonnet#L49-L61) function of the signals library, specifically [here](https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/utils.libsonnet#L56).

Candidly... I'm not sure why.. @v-zhuravlev may be able to shed some light on that 🤔

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,487 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+
+    gce_instance_count:
+      this.signals.gcpceOverview.instanceCount.asStat()
+      + commonlib.panels.generic.stat.base.stylize(),
+    gce_system_problem_count:
+      this.signals.gcpceOverview.systemProblemCount.asStat()
+      + commonlib.panels.generic.stat.base.stylize(),
+    gce_top5_cpu_utilization:
+      this.signals.gcpceOverview.top5CpuUtilization.asTable(format='table')
+      + commonlib.panels.generic.table.base.stylize()
```
rgeyer, line 913: Whenever you use the `as<panel type>` functions on signals, the appropriate styling is already applied, so these lines are superfluous.

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+      },
+
+      top5SystemProblem: {
+        name: 'Top 5 Instances by System problem',
+        description: 'List of top 5 instances with system problems.',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: 'topk(5, sum by (instance_name, job, project_id) (increase(stackdriver_gce_instance_compute_googleapis_com_guest_system_problem_count{%(queriesSelector)s}[$__range])))',
+            legendCustomTemplate: '',
```
rgeyer, line 59: You've used `exprWrappers` on some other signals. Is the default aggregation not suitable for this signal? Could you use gauge, and a `rangeFunction` of `increase` for this?

Same feedback/question for the next few signals as well.

csp-mixin/signals/gcpceOverview.libsonnet
rgeyer: In this file, you're using the "Signals" interface, which is an abstraction on-top of the default Grafana schema. [Signals ](https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/README.md)does not have any support for `step`, so you'll have to add it as an override in `panels.libsonnet`. I don't know that there is a good example of this particular override anywhere.

rgeyer (commented):
Few observations, suggestions, and questions. :)

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: Nice catch!

csp-mixin/signals/gcpce.libsonnet
v-zhuravlev: 'type' is for metric types here: gauge, counter, histogram, info, or raw.

csp-mixin/panels.libsonnet
v-zhuravlev: there is no text panel in the commonlib, and no asText() in singals. I think you can just create a content inside text panel here directly, and you don't need a signal: 
```
    gce_text_instances:
       g.panel.text.new("Instances")
       + g.panel.text.options.withContent("Use this section to look at one instance at a time by picking a value in the *Instance* picker at the top of the dashboard.")
```
https://grafana.github.io/grafonnet/API/panel/text/index.html

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: @rgeyer If I configure it like this 
```
type: 'gauge',
        rangeFunction: 'increase',
        sources: {
          stackdriver: {
            expr: 'stackdriver_gce_instance_compute_googleapis_com_guest_system_problem_count{%(queriesSelector)s}',
            legendCustomTemplate: '',
            aggKeepLabels: ['instance_name'],
            exprWrappers: [['topk(5, ', ')']],
          },
        },
```
it results in: `topk(5, 
  avg by (job,project_id,instance_name) (  stackdriver_gce_instance_compute_googleapis_com_guest_system_problem_count{job!="",job=~"$job",project_id=~"$project_id"}))`
which lacks the increase. I also tried adding the `increase` in the stackdriver object, but it's the same.

v-zhuravlev: you have type: gauge here. For gauge range functions are not applied as doesn't make sense. 

increase can be added by setting signal type to `counter` and also setting `rateFunction:increase` instead of default's `rate`.

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: Thank you both, this was really helpful!

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: This metric(**stackdriver_gce_instance_compute_googleapis_com_guest_system_problem_count**) is a count, so I think we should use type: 'counter' with a rate inside, like this:
```
topk(5, 
  avg by (job,project_id,instance_name) (
  rate(stackdriver_gce_instance_compute_googleapis_com_guest_system_problem_count{job!="",job=~"$job",project_id=~"$project_id"}[$__rate_interval])
)
)
```
but the results aren't looking right.

csp-mixin/signals/gcpceOverview.libsonnet
v-zhuravlev: `short` means no units basicaly... no?

anaivanov: If I use short in the UI, I get the same result as `sishort`, but here it's not the same.
I changed the unit to `sishort` and now I get the result I wanted. Thanks!
https://github.com/grafana/jsonnet-libs/pull/1336/commits/7b2e3a357f88beb10c2e9ec581226f63331ba80b

csp-mixin/panels.libsonnet
anaivanov: Thanks! I added it https://github.com/grafana/jsonnet-libs/pull/1336/commits/f1aacad278d399a8c0f6ec8ecc0430b81421d63d

yduartep:
Hi Ana, I am not sure if using **stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization** metric is the correct way to measure the **number of instances**. I am afraid, if for any case that metric is not sent to Grafana, you will have 0 instances. I wonder if is better to use **stackdriver_gce_instance_compute_googleapis_com_instance_uptime**. Obviously can happen the same but I would check with somebody from the backend team which metric is better to use. You could do something like:

```
count(sum by (instance_name) (stackdriver_gce_instance_compute_googleapis_com_instance_uptime{job!="", job=~".+",project_id=~".+"}))
```

yduartep:
I am not sure if the **System problem count** panel has the correct query. If I breakdown the number of problems by instance in a table in the last moment, the sum of the errors of all of them is more than **200**. I think you are displaying the number of instances that have errors not the number of errors. Exactly what do you want to measure here?
Screenshot 2024-09-24 at 13 59 39

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,279 @@
...
+        name: 'Instance count',
+        description: 'Number of VM instances',
+        type: 'raw',
+        unit: 'short',
+        sources: {
+          stackdriver: {
+            expr: 'count(sum by (instance_name) (stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{%(queriesSelector)s}))',
+            legendCustomTemplate: '',
+          },
+        },
```
yduartep, line 23: Here you could use `type='gauge'`,  `exprWrappers: [['count(', ')']],` `expr='stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{%(queriesSelector)s}'`.

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,279 @@
...
+      },
+      tableUptime: {
+        name: 'tableUptime',
+        type: 'raw',
+        sources: {
+          stackdriver: {
+            expr: 'sum by (instance_name, job, project_id)(stackdriver_gce_instance_compute_googleapis_com_instance_uptime_total{%(queriesSelector)s})',
+            legendCustomTemplate: '',
+          },
+        },
```
yduartep, line 109: You can use `type=gauge` here and `expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_uptime_total{%(queriesSelector)s}'`

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,279 @@
...
+      },
+      tableSentPackets: {
```
yduartep, line 111: I think **tableSentPackets, tableReceivedPackets, tableSentBytes, tableReceivedBytes, tableReadBytes, tableWriteBytes** don't need a **rate** function because are not CUMULATIVE metrics. You could convert it as `type=gauge` or better use directly existing signals **diskBytesWrite, diskBytesRead, networkReceived, networkSent, packetsReceived, packetsSent** converted to gauge. You don't need to define same signals twice just because are being used as table or as timeSeries.

yduartep (changes requested):
I added a few comments...

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: I will add the count in the `exprWrappers `, but keep the raw type because I want to do the `sum by` only by `instance_name`, otherwise it will not return the sum uniquely counted by the unique name, but by job which is the same for all

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,666 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+          + g.query.prometheus.withInterval('1m')
+          + g.query.prometheus.withFormat('table'),
+          this.signals.gcpceOverview.tableReadBytes.asTarget()
+          + g.query.prometheus.withInterval('1m')
+          + g.query.prometheus.withFormat('table'),
+          this.signals.gcpceOverview.tableWriteBytes.asTarget()
+          + g.query.prometheus.withInterval('1m')
+          + g.query.prometheus.withFormat('table'),
+        ]
+      )
```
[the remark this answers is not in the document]
v-zhuravlev: you could also use: asTable() for the first signal  and asTableColumn() for the rest:
like:
`this.signals.gcpceOverview.tableCpuUtilization.asTable(format="table")
+ this.signals.gcpceOverview.tableUptime.asTableColumn(format="table")`

See here: https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/README.md#signal
and example here: https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/panels/topic.libsonnet#L16-L20

csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,119 @@
...
+        type: 'counter',
+        unit: 'seconds',
+        aggFunction: 'avg',
+        sources: {
+          stackdriver: {
+            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_cpu_usage_time{%(queriesSelector)s}',
+            legendCustomTemplate: '{{instance_name}}',
+          },
+        },
+      },
```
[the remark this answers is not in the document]
anaivanov: This is a counter because the value is cumulative

**See the panel with rate:**
image
**and without:**
image

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+                  },
+                  {
+                    color: 'green',
+                    value: 30,
+                  },
+                  {
+                    color: 'red',
+                    value: 85,
+                  },
+                ],
```
yduartep, line 1322: Can we sync on the threholds we use? I am using
- Base Green 
- from 70 Orange
- from 90 Red

Let's decide together which one to use in all our dashboards. What do you think?

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+                type: 'color-text',
+              },
+            },
+            {
+              id: 'color',
+              value: {
+                fixedColor: 'blue',
+                mode: 'fixed',
+              },
+            },
```
yduartep, line 1349: Any particular reason you show text in blue and no in normal text color (white)? I want to use the same criteria on my dashboard

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+            {
+              id: 'displayName',
+              value: 'Network throughput Send [5m]',
```
yduartep, line 1431: Can you make this title shorter. It's hard to read now because the column is short. What about just **Network sent [5m]**?

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+              id: 'color',
+              value: {
+                fixedColor: 'blue',
+                mode: 'fixed',
+              },
+            },
+            {
+              id: 'displayName',
+              value: 'Network throughput Received [5m]',
+            },
```
yduartep, line 1461: Can you make this title shorter. It's hard to read now because the column is short. What about just **Network received [5m]**?

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+            {
+              id: 'color',
+              value: {
+                fixedColor: 'blue',
+                mode: 'fixed',
+              },
+            },
+            {
+              id: 'displayName',
+              value: 'Total Bytes count read [5m]',
```
yduartep, line 1489: Same here. What about just **Disk read** or **Disk read bytes**? The idea is that can be readable on the column.

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+            {
+              id: 'color',
+              value: {
+                fixedColor: 'blue',
+                mode: 'fixed',
+              },
+            },
+            {
+              id: 'displayName',
+              value: 'Total Bytes count write [5m]',
```
yduartep, line 1518: Same here. What about just **Disk write** or **Disk write bytes**?

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,667 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+          + g.query.prometheus.withFormat('table'),
+          this.signals.gcpceOverview.tableReceivedBytes.asTarget()
+          + g.query.prometheus.withInstant(true)
+          + g.query.prometheus.withFormat('table'),
+          this.signals.gcpceOverview.tableReadBytes.asTarget()
+          + g.query.prometheus.withInstant(true)
+          + g.query.prometheus.withFormat('table'),
+          this.signals.gcpceOverview.tableWriteBytes.asTarget()
+          + g.query.prometheus.withInstant(true)
+          + g.query.prometheus.withFormat('table'),
```
yduartep, line 1227: Can you show first **Disk read and write** and then **Network sent and received** columns? It’s just to have same order in GCP and Azure. And also to have the same order you use on the breakdown panels below.

csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,119 @@
...
+        sources: {
+          stackdriver: {
+            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_network_sent_bytes_count{%(queriesSelector)s}',
+            legendCustomTemplate: '{{instance_name}}',
+          },
+        },
+      },
+
+      diskReadBytes: {
+        name: 'Count of disk read bytes',
```
yduartep, line 66: Can you rename this to something shorter like “**Disk read bytes**” or **Disk read bytes (Total)**? You can also omit total if you want. I prefer to read first the type of element to be measured (disk, network,...) then the function (read, write, send, received,...) and then maybe the unit (bytes, total, count, ...)

csp-mixin/signals/gcpce.libsonnet
```diff
@@ -0,0 +1,119 @@
...
+        aggFunction: 'avg',
+        sources: {
+          stackdriver: {
+            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_disk_write_bytes_count{%(queriesSelector)s}',
+            legendCustomTemplate: '{{instance_name}}',
+          },
+        },
+      },
+      diskReadOperations: {
+        name: 'Count of disk read operations',
```
yduartep, line 93: also here what about **Disk read operations (count)**? or without count

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+        unit: 'short',
+        sources: {
+          stackdriver: {
+            expr: 'sum by (instance_name) (stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{%(queriesSelector)s})',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['instance_name'],
+            exprWrappers: [['count(', ')']],
+          },
+        },
+      },
```
yduartep, line 26: You can use **type=gauge** and remove extras from **expr** and you will have the same result because an instance cannot be in 2 different job and project at the same time. I tested and works

```suggestion
     instanceCount: {
        name: 'Instance count',
        description: 'Number of VM instances',
        type: 'gauge',
        unit: 'short',
        sources: {
          stackdriver: {
            expr: 'stackdriver_gce_instance_compute_googleapis_com_instance_cpu_utilization{%(queriesSelector)s}',
            legendCustomTemplate: '',
            aggKeepLabels: ['instance_name'],
            exprWrappers: [['count(', ')']],
          },
        },
      },
```

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+        type: 'gauge',
+        sources: {
+          stackdriver: {
+            expr: 'rate(stackdriver_gce_instance_compute_googleapis_com_instance_network_received_packets_count{%(queriesSelector)s}[5m])',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['instance_name'],
+          },
+        },
+      },
+      tableSentBytes: {
```
yduartep, line 148: Sent bytes of what? maybe is more clear if you rename it as **tableNetworkSent**. Imagine if somebody else read the code. I had to read the metric to understand the purpose of this signal.

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+        sources: {
+          stackdriver: {
+            expr: 'rate(stackdriver_gce_instance_compute_googleapis_com_instance_network_received_bytes_count{%(queriesSelector)s}[5m])',
+            legendCustomTemplate: '',
+            aggKeepLabels: ['instance_name'],
+          },
+        },
+      },
+      tableReadBytes: {
+        name: 'tableReadBytes',
```
yduartep, line 171: same here. Maybe **tableDiskRead** is more clear?

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+            legendCustomTemplate: '',
+            aggKeepLabels: ['instance_name'],
+          },
+        },
+      },
+      // end table
+
+      memoryUtilization: {
+        name: 'Memory Utilization',
+        description: 'Memory Utilization',
```
yduartep, line 196: Can we use always the same pattern? First word in uppercase, the rest in lowercase?
```suggestion
        description: 'Memory utilization',
```

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+        type: 'raw',
+        unit: 'bytes',
+        sources: {
+          stackdriver: {
+            expr: 'sum(stackdriver_gce_instance_compute_googleapis_com_instance_memory_balloon_ram_size{instance_name=~"$instance_name"} or on() vector(0)) + sum(stackdriver_gce_instance_compute_googleapis_com_guest_memory_bytes_used{instance_name=~"$instance_name"} OR on() vector(0))',
```
yduartep, line 201: Don't understand why you implemented the query in this way. I think if you move this signal and the memoryUsed signal to **gce.libsonnet**, you will have the filter by **instance_name** plus the other filters included and then the query can be changed to something like this:

```suggestion
           expr: 'sum(stackdriver_gce_instance_compute_googleapis_com_instance_memory_balloon_ram_size{%(queriesSelector)s}) + sum(stackdriver_gce_instance_compute_googleapis_com_guest_memory_bytes_used{%(queriesSelector)s})',
```

csp-mixin/signals/gcpceOverview.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+      },
+
+      memoryUsed: {
+        name: 'Memory Utilization2',
+        description: 'Memory used',
+        type: 'raw',
+        unit: 'bytes',
+        sources: {
+          stackdriver: {
+            expr: 'sum(stackdriver_gce_instance_compute_googleapis_com_instance_memory_balloon_ram_used{instance_name=~"$instance_name"} OR on() vector(0)) + sum(stackdriver_gce_instance_compute_googleapis_com_guest_memory_bytes_used{instance_name=~"$instance_name", state!="free"} OR on() vector(0))',
```
yduartep, line 214:
```suggestion
expr: 'sum(stackdriver_gce_instance_compute_googleapis_com_instance_memory_balloon_ram_used{%(queriesSelector)s}) + sum(stackdriver_gce_instance_compute_googleapis_com_guest_memory_bytes_used{%(queriesSelector)s, state!="free"})',
```

yduartep (changes requested):
I added a few comments that I resume here:

- Why instance count is in red?
- Can you show Top Instance by Disk read bytes first and write after? 
- On the table can you show first Disk read and write and then Network in and Out? It’s just to have same order in GCP and Azure. And also to have the same order you use on the breakdown panels below. Also rename the column titles with a shorter value that can be visible like Disk read bytes | Disk write bytes, Network received, Network sent?
- Can you rename Total packets count sent/recieved to something shorter like just “Total packets sent/recieved”?
- Can you rename Memory Utilization second word in lowercase?
- Can you rename Network throughput Sent/Received to just “Network sent/received”?
- Can you rename Total Bytes count read/write to just “Disk bytes (total)” or Disk read/write (bytes) or Disk read/write (total)? Using Total and count together is a bit confusing for a title. Also you are already displaying on the legend the Read and Write. Not sure if we should include them also in the title.
- Can you rename “Count of disk read bytes” to something shorter like “Disk read bytes (total)”? You can also omit total if you want. Same for operations.

csp-mixin/panels.libsonnet
anaivanov: I got these thresholds from Vasil, so perhaps you can update yours?

csp-mixin/panels.libsonnet
anaivanov: Good call!

csp-mixin/panels.libsonnet
anaivanov: I'll keep this order because it's the same as in the below panels

csp-mixin/signals/gcpceOverview.libsonnet
anaivanov: I don't see a strong rule about this, when I look at other dashbaords.
I think I prefer having Capital letters for each word except for prepositions

csp-mixin/panels.libsonnet
yduartep: but why not use like 70 for yellow or orange for example?

csp-mixin/panels.libsonnet
yduartep: as we use colors on threholds, If you have them in blue it makes me think that blue means something specific and ca be confused

csp-mixin/panels.libsonnet
anaivanov: I believe the ideal CPU value is between 30 and 85 (green) because it should not be under utilized, or over utilized

csp-mixin/signals/gcpceOverview.libsonnet
yduartep: in my PRs I have comments about having standard naming on the titles for that reason I think we can use the same. If people in the past didn't do it does not mean we can decide to start using it

csp-mixin/panels.libsonnet
anaivanov: I removed this https://github.com/grafana/jsonnet-libs/pull/1336/commits/45514b921f03f2a0266a9b73aaf71372a7400617

yduartep (approved):
yeahhhhh great work Ana. Thank you so much for taking care of my comments.

csp-mixin/panels.libsonnet
```diff
@@ -899,5 +899,636 @@ local commonlib = import 'common-lib/common/main.libsonnet';
...
+              id: 'thresholds',
+              value: {
+                mode: 'absolute',
+                steps: [
+                  {
+                    color: 'text',
+                    value: null,
+                  },
+                ],
+              },
```
rgeyer, line 932: I think you may be doing this in order to change the stats text color to "gray", but it doesn't appear to be taking effect.

I achieved the same thing in the queue storage dashboard using [this solution](https://github.com/grafana/jsonnet-libs/blob/547934640d91d6ed08e69229c5f6a4ec6e3ed250/csp-mixin/panels/azurequeuestore.libsonnet#L66-L69).

rgeyer (approved):
Looks great! Just one little nit about the color on the top stat panels. @yduartep caught me on this on the azure queue store dashboard, and I linked to how I solved it there.

Great work, and thank you for all the thrash addressing feedback!

image

csp-mixin/panels.libsonnet
anaivanov: Thank you, I like better your solution! https://github.com/grafana/jsonnet-libs/pull/1336/commits/94adb658561f72444ac7851c70e070dda64d55cd

v-zhuravlev: JFYI, you could also add default commonlib style to stat panels, to avoid seeing everything in green(<80) and red(=>80):
like so:
```
      this.signals.gcpceOverview.instanceCount.asStat()
      + commonlib.panels.generic.stat.base.stylize()
```
It has the same default style of color text, not green/red thresholds.
https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/panels/generic/stat/base.libsonnet

anaivanov:
Thank you all for the feedback! I learned a lot from you while learning this library!

PR #1338, opened 2024-09-27
update cpu utilization thresholds and rule value to trigger alert

yduartep:
- Update cpu utilization thresholds to sync value with gcp compute engine value.
- Update rule value used to trigger the alert based on the previous red value


anaivanov (approved):
LGTM

PR #1340, opened 2024-10-02
add dashboard uid, service and namespace to the azure alerts

yduartep:
add dashboard uid, service and namespace to the azure alerts


anaivanov (approved):
LGTM

PR #1342, opened 2024-10-04
Add google VPC dashboard to csp-mixin

rgeyer:
Adds one new dashboard for Google VPC data.

image
image

### DIFF
--- a/csp-mixin/panels/gcpvpc.libsonnet
+++ b/csp-mixin/panels/gcpvpc.libsonnet
@@ -66,7 +66,7 @@ local commonlib = import 'common-lib/common/main.libsonnet';
         [
           this.signals.gcpvpc.gcpvpc_service_topk_throughput_bytes.asTarget()
           + g.query.prometheus.withFormat('table')
-          + g.query.prometheus.withInstant(true)
+          + g.query.prometheus.withInstant(true),
         ],
         'Average throughput',
       ) + self._serviceTableCommon('Average throughput'),
@@ -79,7 +79,7 @@ local commonlib = import 'common-lib/common/main.libsonnet';
         [
           this.signals.gcpvpc.gcpvpc_service_topk_error_percent.asTarget()
           + g.query.prometheus.withFormat('table')
-          + g.query.prometheus.withInstant(true)
+          + g.query.prometheus.withInstant(true),
         ],
         'Average error rate',
       ) + self._serviceTableCommon('Average error rate'),
--- a/csp-mixin/panels/gcpvpc.libsonnet
+++ b/csp-mixin/panels/gcpvpc.libsonnet
@@ -1,0 +1,134 @@
+local g = import '../g.libsonnet';
+local commonlib = import 'common-lib/common/main.libsonnet';
+{
+  new(this): {
+    _serviceTableCommon(valueColNiceName)::
+      g.panel.table.queryOptions.withTransformations([
+        g.panel.table.queryOptions.transformation.withId('filterFieldsByName')
+        + g.panel.table.queryOptions.transformation.withOptions({
+          include: {
+            pattern: 'service_name.*|Value.*',
+          },
+        }),
+        g.panel.table.queryOptions.transformation.withId('joinByField')
+        + g.panel.table.queryOptions.transformation.withOptions({
+          byField: 'service_name',
+          mode: 'outer',
+        }),
+      ])
+      + g.panel.table.standardOptions.withOverrides([
+        {
+          matcher: {
+            id: 'byName',
+            options: 'service_name',
+          },
+          properties: [
+            {
+              id: 'displayName',
+              value: 'Service name',
+            },
+          ],
+        },
+        {
+          matcher: {
+            id: 'byName',
+            options: 'Value',
+          },
+          properties: [
+            {
+              id: 'displayName',
+              value: valueColNiceName,
+            },
+          ],
+        },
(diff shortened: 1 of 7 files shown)


csp-mixin/panels/gcpvpc.libsonnet
```diff
@@ -0,0 +1,134 @@
...
+      + g.panel.stat.standardOptions.color.withMode('fixed')
+      + g.panel.stat.standardOptions.color.withFixedColor('text'),
+
+    gcpvpc_fixed_usage_tier_network_egress:
+      this.signals.gcpvpc.gcpvpc_fixed_usage_tier_network_egress.asTimeSeries()
+      + g.panel.timeSeries.standardOptions.withNoValue('0'),
+
+    gcpvpc_service_topk_throughput:
+      this.signals.gcpvpc.gcpvpc_service_topk_throughput_bytes.common
+      + g.panel.table.standardOptions.withUnit('bps')
```
anaivanov, line 63: isn't this unit mentioned in the signal as well?

yduartep: I think, even if you use **asTable** the units are not propagated for any rare reason. This is happening just on tables. Maybe is an error on the common lib.

csp-mixin/panels/gcpvpc.libsonnet
```diff
@@ -0,0 +1,134 @@
...
+      + commonlib.panels.generic.table.base.new(
+        'Top 5 services - Average throughput',
+        [
+          this.signals.gcpvpc.gcpvpc_service_topk_throughput_bytes.asTarget()
+          + g.query.prometheus.withFormat('table')
+          + g.query.prometheus.withInstant(true),
+        ],
+        'Average throughput',
+      ) + self._serviceTableCommon('Average throughput'),
```
yduartep, line 72: I wonder, if in a future we includes a new parameter **instant** on the common-lib / **asTable** function , maybe we can stop doing this. Must of the time we use this kind of definition because we cannot define **instant** param into **asTable**. @v-zhuravlev what do you think? Should I create a new issue for this?

csp-mixin/panels/gcpvpc.libsonnet
```diff
@@ -0,0 +1,134 @@
...
+    gcpvpc_tunnel_throughput:
+      commonlib.panels.network.timeSeries.traffic.new('Total tunnel network throughput', targets=[])
+      + commonlib.panels.network.timeSeries.traffic.withNegateOutPackets()
+      + g.panel.timeSeries.fieldConfig.defaults.custom.stacking.withMode('normal')
+      + this.signals.gcpvpc.gcpvpn_tunnel_egress_bytes.asPanelMixin()
+      + this.signals.gcpvpc.gcpvpn_tunnel_ingress_bytes.asPanelMixin(),
+
+    gcpvpc_tunnel_by_protocol_throughput:
+      commonlib.panels.network.timeSeries.traffic.new('Total tunnel network throughput by protocol', targets=[])
+      + commonlib.panels.network.timeSeries.traffic.withNegateOutPackets()
```
yduartep, line 122: Nice!!

csp-mixin/signals/gcpvpc.libsonnet
```diff
@@ -0,0 +1,295 @@
...
+      // Overview counters
+      //
+      gcpvpc_services_in_use_count: {
+        name: 'Total services in use',
+        description: 'Count of unique service names (including "UNKNOWN") with ingress or egress traffic.',
+        type: 'raw',
+        unit: 'short',
+        sources: {
+          stackdriver: {
+            expr: 'count(count by (service_name) (stackdriver_google_service_gce_client_networking_googleapis_com_google_service_response_bytes_count{%(queriesSelector)s} OR stackdriver_google_service_gce_client_networking_googleapis_com_google_service_request_bytes_count{%(queriesSelector)s}))',
```
yduartep, line 40: Why do you use `count by service_name` OR the `same metric value `itself? What is the purpose?

yduartep, line 40: Why aren't you counting the requests for `service_name` as you do for responses?

yduartep (approved):
Great work!! I approved because I think is Ok but I left a few questions that I would like to understand.

csp-mixin/panels/gcpvpc.libsonnet
anaivanov: Removing it from the signal would make sense!

anaivanov (approved):
LGTM

csp-mixin/signals/gcpvpc.libsonnet
rgeyer: Just an oversight.. Possibly a copy/paste error. I'll make sure they're using the same metric, since that makes more sense.

I'm fairly sure I copy/pasted this from a "total throughput" query, which had both request and response metrics in it, and just failed to fully modify the query for this purpose.

Thanks for catching it!

rgeyer: Actually.. No.. I responded to this before looking at it closely enough, and before my coffee.

This query is first counting the number of samples for either requests, or responses, by `service_name`. This results in a list of unique service names. Then counting the number of samples returned by that query, gives me the total number of services in use.

Just using request, or response would likely be adequate, but because the count is aggregating on `service_name`, it's safe to include both as a "just in case" measure.

PR #1344, opened 2024-10-11
Bump logs-lib dep to grafana 11

v-zhuravlev:
Bump logs-lib dep to grafana 11
Also Add wrapPanels to calculate x,y properly


Dasomeone (approved):
LGTM, thanks Vitaly! 

What do you think, should we set up some regular check for these modular libs so we all help maintain them?

PR #1347, opened 2024-10-21
Cilium clustermesh alert

balous:
This PR adds two groups of rules:
- Cilium agent clustermesh alerts to detect when agent has problems to mesh with remote clusters
- Cilium kvstoremesh alerts to detect when kvstoremesh component has problems to cache remote cluster's identities


balous:
Fixed.

Dasomeone (approved):
Awesome, thanks for making the changes!

PR #1348, opened 2024-10-22
Add per-mixin workflow for linting

Dasomeone:
This PR reintroduces mixin linting via the use of [mixtool](https://github.com/monitoring-mixins/mixtool), but in an updated fashion to only account for changes made to any given mixins in the current working push request.

This is done to avoid blocking PR authors when mixins outwith the changes they're making fail linting.

### DIFF
--- a/.github/workflows/lint-mixins.yml
+++ b/.github/workflows/lint-mixins.yml
@@ -39,6 +39,7 @@ jobs:
           dir_names_max_depth: 1
           files: | 
             **-mixin/
+            **-observ-lib/
           matrix: true
           
       - name: List all changed mixins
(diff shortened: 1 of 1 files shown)


.github/workflows/lint-mixins.yml
```diff
@@ -0,0 +1,83 @@
...
+      
+      - name: Get changed mixins
+        id: changed-mixins
+        uses: tj-actions/changed-files@v44
+        with:
+          dir_names: true
+          dir_names_exclude_current_dir: true
+          dir_names_max_depth: 1
+          files: | 
+            **-mixin/
```
v-zhuravlev, line 41: some observ-libs would have sample mixin.libsonnet file too.
Should we try to lint it as well?

For example: kafk-observ-lib/mixin.libsonnet

v-zhuravlev (approved):

v-zhuravlev (approved):

PR #1349, opened 2024-10-22
Fix Incorrect panel name and description in NGINX Mixin

ishanjainn:
Just checked the panel name and description had a wrong placeholder in my previous PR. This PR just fixes them
cc @Dasomeone


Dasomeone (approved):
Thanks for catching these!

PR #1353, opened 2024-11-02
windows: Use instance label for hostname in hostname panel

v-zhuravlev:
As hostname label is not present in windows_os_info metric.


Dasomeone (approved):
Third time's the charm Github, please approve this time...

PR #1354, opened 2024-11-12
Add current all to clickhouse-mixin

v-zhuravlev:
This to ensure all nodes(instances) of clickhouse cluster are selected when nothing is set specifically.
Updated via old grafonnet funcion arg.


Dasomeone (approved):
LGTM, thanks Vitaly

PR #1355, opened 2024-11-12
Add support for empty (filteringSelector='')

v-zhuravlev:
1) Add support for empty (filteringSelector='') to support cases when we don't want to have any static filter at all. 
You can now set filteringSelector to ''
and in queries this:
`{job!="", job=~"$job"}`
would become
`{job=~"$job"}`
2) Turn `datasource` to `prometheus_datasource` when loki is enabled this helps to pass dashboard-linter rule:


Dasomeone (approved):
Nice, loving all these improvements!

PR #1356, opened 2024-11-12
Kafka-observ-lib: Set totalTime panels repeat configurable

v-zhuravlev:


Dasomeone (approved):
This is much cleaner, great work!

mshahzeb (approved):

PR #1360, opened 2024-11-21
fix/apache-solr-mixin - Fixes a number of issues for Apache Solr Mixin

StefanKurek:
Fixes issue with Resource Monitoring Dashboard where '~' was missing in queries for selector to `solr_cluster` variable.

Change custom allValues for solr_collection variable from `.+` to `.*` in Cluster Overview Dashboard to fix an issue in panels when `collection` label was not present in prometheus metrics.

Change custom allValues for solr_collection variable from `.+` to `.*` in Query Performance Dashboard to fix an issue in panels when `collection` label was not present in prometheus metrics.


Dasomeone (approved):
Thanks for addressing this Stefan, this is fine for metrics dashboards, but just a heads up `.*` will break the logs dashboard if you want to make changes to that in the future :)

PR #1361, opened 2024-11-22
static-exporter: Update httpd.conf file to include headers

Pokom:
Prometheus v3.0.0 has a stricter requirements on the headers that are returned from exporters. The static-exporter does not set any headers and causes scrapes to fail with the following error:

```
non-compliant scrape target sending blank Content-Type and no fallback_scrape_protocol specified for target ````
```

This adds an httpd.conf file to mount to the container that includes a directive to add a header that solves this problem. To generate the httpd.conf file, I ran the following command:

`docker run --rm httpd:2.4 cat /usr/local/apache2/conf/httpd.conf > my-httpd.conf`


Duologic (approved):

jjo:
1st pass:
* Please add some bits from this PR description into `README.md`, especially the httpd.conf should be updated, in case the (hardcoded) image changes
* nit: in above instructions, let's also use precisely the same image, i.e. `docker run --rm httpd:2.4-alpine ...`

static-exporter/jsonnetfile.json
```diff
@@ -1,6 +1,15 @@
...
   "version": 1,
   "dependencies": [
+    {
+      "source": {
+        "git": {
+          "remote": "https://github.com/grafana/jsonnet-libs.git",
+          "subdir": ""
+        }
+      },
+      "version": "static-exporters-add-content-type-headers"
```
Duologic, line 11: Seems like an accidental change

static-exporter/main.libsonnet
```diff
@@ -46,6 +48,18 @@ local k = import 'ksonnet-util/kausal.libsonnet';
         ),
     }),
 
+  withHttpConfig():: {
+    local configMap = k.core.v1.configMap,
+    local volumeMount = k.core.v1.volumeMount,
+    httpdConfig:
+      configMap.new(self.name, 'httpd-config')
+      + configMap.withData({
+        'httpd.conf': importstr 'httpd.conf',
```
Duologic, line 57:
```suggestion
  withHttpConfig(config=(importstr 'httpd.conf')):: {
    local configMap = k.core.v1.configMap,
    local volumeMount = k.core.v1.volumeMount,
    httpdConfig:
      configMap.new(self.name, 'httpd-config')
      + configMap.withData({
        'httpd.conf': config,
```
I'd set the default in the argument, that'll allow downstream to override it with their own.

PR #1365, opened 2024-12-05
nit: conditional valueMaps in overrides in signals lib

v-zhuravlev:
nit: conditional valueMaps in overrides in signals lib. Even though grafana handled the result before, this would produce cleaner output.


Dasomeone (approved):
Looks great. definitely in favour of defaulting towards cleaner output whenever possible

Dasomeone:
@v-zhuravlev this PR is getting some index out of bounds errors on the build

v-zhuravlev:
yes, fixed it now

PR #1375, opened 2024-12-17
add dashboard for azure front door

yduartep:
add dashboard for azure front door

### DIFF
--- a/csp-mixin/panels/azurefrontdoor.libsonnet
+++ b/csp-mixin/panels/azurefrontdoor.libsonnet
@@ -40,19 +40,15 @@ local commonlib = import 'common-lib/common/main.libsonnet';
               {
                 id: 'thresholds',
                 value: {
-                  mode: 'absolute',
+                  mode: 'percentage',
                   steps: [
                     {
-                      color: 'yellow',
-                      value: null,
-                    },
-                    {
-                      color: 'green',
-                      value: 30,
+                      color: 'text',
+                      value: 0,
                     },
                     {
                       color: 'red',
-                      value: 85,
+                      value: 1,
                     },
                   ],
                 },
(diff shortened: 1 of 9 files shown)


csp-mixin/panels/azurefrontdoor.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+                      value: null,
+                    },
+                    {
+                      color: 'green',
+                      value: 30,
+                    },
+                    {
+                      color: 'red',
+                      value: 85,
+                    },
```
anaivanov, line 56: should we switch `yellow` with `green` thresholds? so that when there are 0 errors the cell is green, then above ~~30~~ 15 it turns yellow and above (perhaps update the threshold to 50) it turns red. what do you think?

csp-mixin/panels/azurefrontdoor.libsonnet
```diff
@@ -0,0 +1,300 @@
...
+                      color: 'yellow',
+                      value: null,
+                    },
+                    {
+                      value: 30,
+                      color: 'green',
+                    },
+                    {
+                      color: 'red',
+                      value: 85,
```
anaivanov, line 228: Looks like the meaning of this percent is the number of `successful health probes from AFDX to backends.` Sounds to me that the closer it gets to 100 it should be green, so:
null - red
85 - yellow
90 - green

csp-mixin/panels/azurefrontdoor.libsonnet
anaivanov: I think if it's null, it can be text-color. Then anything above it can be red

csp-mixin/panels/azurefrontdoor.libsonnet
yduartep: THis is a percentage of successful health probes. I think 0 is no probes executed

anaivanov (approved):
LGTM!

PR #1383, opened 2025-01-10
Update rabbitmq alerts

v-zhuravlev:
Remove RabbitMQNodeDown as it could be false-positive for single-node "clusters".


Dasomeone (commented):
Couldn't we add a conditional check where this is only fired if rabbitmq_cluster labels are present? Feels a bit dangerous to outright remove it..

Also, is this in response to a bug report? Just wondering if we can link the issue and make sure it's closed and to have it for future context

PR #1384, opened 2025-01-14
feat(static-exporter): allow values to be printed as floats

Duologic:
For exporting CPU information, it might be useful to export the value as a float.


tinitiuset (approved):
LGTM

PR #1386, opened 2025-01-15
Fix withDataLink() in generic.timeSeries and move to base from topk_percentage

v-zhuravlev:
Fix withDataLink() in generic.timeSeries and move this function to base.libsonnet from topk_percentage.libsonnet so it available to all timeSeries panels in commonlib.


Dasomeone (approved):
Nice!

PR #1387, opened 2025-01-15
commonlib Add optional adHoc variables

v-zhuravlev:
Add optional adHoc variables to variables generated by commonlib.

### DIFF
--- a/common-lib/common/variables/variables.libsonnet
+++ b/common-lib/common/variables/variables.libsonnet
@@ -63,7 +63,7 @@ local utils = import '../utils.libsonnet';
        },
        adHoc:
          var.adhoc.new('adhoc', 'prometheus', '${' + root.datasources.prometheus.name + '}')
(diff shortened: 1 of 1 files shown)


common-lib/common/variables/variables.libsonnet
```diff
@@ -59,14 +61,30 @@ local utils = import '../utils.libsonnet';
            + var.datasource.generalOptions.withLabel(prometheusDatasourceLabel)
            + var.datasource.withRegex(''),
        },
+       adHoc:
+         var.adhoc.new('adhoc', 'prometheus', '${' + root.datasources.prometheus.name + '}')
+         + var.adhoc.generalOptions.withLabel('Adhoc filters')
```
Dasomeone, line 66: When we write it out, is it "Adhoc", "AdHoc" or "Ad hoc" 

Because my personal take is "ad hoc" but if we have a formal naming for this feature we should use that, just wanting to clarify :)

PR #1388, opened 2025-01-16
Update commonlib(signals)

v-zhuravlev:
- Variables: Add support for optional ad Hoc variabels
- Signals: Add support for info metrics in asTableColumn(format='table), now properly showing info metric as one more column
- Signals: asTable() update: Order and rename Instance/Group labes as first columns in table
- Signals: add new attribute "nameShort". if defined this optional alias is used in legends and table column names.
- Signals: Introduce new class of functions called "modifiers" and first member .withTopK(limit). They don't render signal immediately, instead, they modify the signal. Can be used in builder patterns before finally rendering final panel. Very useful when signals should be rendered a little bit different on different dashboards:
i..e
on fleet dashboard:
`signal.withTopK(25).asTimeSeries()`
on signle host dashboard:
`signal.asTimeSeries()`
- Signals: fix: group aggregations are now properly applied to info type metrics as well.

checklist:
- [x] README updated
- [x] Tests updated


common-lib/common/signal/base.libsonnet
```diff
@@ -182,36 +241,54 @@ local xtd = import 'github.com/jsonnet-libs/xtd/main.libsonnet';
...
+      ),
 
     //Return as timeSeriesPanel
-    asTimeSeries(name=signalName)::
-      g.panel.timeSeries.new(name)
-      + self.common,
+    asTimeSeries(name=signalName, prototype=g.panel.timeSeries.new(name)):
+
+      // local _prototype = std.split(prototype, '.');
+      // panels[_prototype[0]][_prototype[1]][_prototype[2]].new(title='', targets=[])
```
Dasomeone, line 266: Leftovers from testing?

common-lib/common/signal/signal.libsonnet
```diff
@@ -135,6 +140,7 @@ local stub = import './stub.libsonnet';
                name=name,
                type=metricType,
                unit=std.get(signalsJson.signals[s], 'unit', ''),
+               nameShort=std.get(signalsJson.signals[s], 'nameShort', name),
```
Dasomeone, line 143: shouldn't it be `nameShort` instead of `name` at the end here?

Dasomeone (commented):
This is super super cool, and pretty neatly done!
Despite that I'm pretty sure I'm overlooking a couple things as there's so many changes, so would prefer getting a second pair of eyes on for final approval once my comments have been addressed

common-lib/common/signal/signal.libsonnet
v-zhuravlev: nope, it is `name` as it is default fallback, if nameShort is not found in JSON.

common-lib/common/signal/base.libsonnet
v-zhuravlev: oops, indeed it is some garbage left. Removed.

common-lib/common/signal/signal.libsonnet
Dasomeone: Ahh of course that makes more sense, thank you for clarifying!

Dasomeone (approved):
Thanks for addressing the comments, lgtm, but as stated would prefer a second approval before merge :)

mshahzeb (approved):
LGTM! Great work.

v-zhuravlev:
FYI, started simple changelog.

PR #1392, opened 2025-01-24
Fix increase queries in clickhouse

v-zhuravlev:
Increase should be used together with $__interval, not $__rate_interval to get proper increments.


Dasomeone (approved):
Nice catch, thanks Vitaly!

PR #1394, opened 2025-02-07
nginx-directory: add light/dark style

asimpson:
Adds light/dark theme support to the nginx directory index.html page.

Adds some basic styling as well. Including making the help "tooltip" feature bigger since it's difficult to mouse over in its current size.

Dark mode:
image

Light mode:
image

### DIFF
--- a/nginx-directory/files/index.html
+++ b/nginx-directory/files/index.html
@@ -1,20 +1,122 @@
 <html>
-  <head>
-    <style>
-      .help {
-        color: grey;
-        font-size: 50%%; 
-        text-decoration-line: underline;
-        text-decoration-style: dashed;
-        vertical-align: super;
+
+<head>
+  <style>
+    *,
+    *:before,
+    *:after {
+      box-sizing: border-box;
+    }
+
+    body {
+      font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
+      margin: 0;
+      padding: 24px;
+      color: #424242;
+      line-height: 1.7;
+      background: #fff;
+    }
+
+    h1 {
+      border-bottom: 1px solid #e4e4e4;
+    }
+
+    .help {
+      color: #666;
+      text-decoration-line: underline;
+      text-decoration-style: dashed;
+      vertical-align: super;
+    }
+
+    ul {
+      list-style: none;
+      margin: 0;
+      padding: 0;
+    }
+
+    li {
+      margin: 0;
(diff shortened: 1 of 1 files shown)


sarahzinger (approved):
🌔

nginx-directory/files/index.html
```diff
@@ -1,20 +1,122 @@
...
+    *:before,
+    *:after {
+      box-sizing: border-box;
+    }
+
+    body {
+      font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif;
+      margin: 0;
+      padding: 24px;
+      color: #424242;
```
iainlane, line 15: could we consider using CSS variables to make the theme easier to grok / modify? could be a way to do the light/dark theme too.

nginx-directory/files/index.html
```diff
@@ -1,20 +1,122 @@
...
+      background: #f8f9fa;
+      font-weight: 500;
+    }
+
+    button,
+    input[type="button"] {
+      padding: 4px 12px;
+    }
+
+    @media (prefers-color-scheme: dark) {
```
iainlane, line 75: as I understand this, it will always sync with the "system" setting, right?

no opinion on that right now - just wanted to call it out and see what others think. might be that if we did this we'd like a toggle that saves the preference in local storage, because not all folks will like that.

iainlane (commented):
cheers! i think there's scope for visual improvement here... just had some questions to get the ball rolling and I'll ask for more input

Duologic (commented):
You have to consider that we stuff this file into a ConfigMap, the size matters (<1mb), I bet it does fit, just raising awareness.

Duologic:
I'd personally argue not to style this page, the point is to keep the page dead simple. Adding styles opens the door for all kinds of personal preferences.

There are plugins that can configure any page to dark-style: https://darkreader.org/ and there is also some browsers supporting user-stylesheets, if you must I'd suggest looking at that.

asimpson:
> You have to consider that we stuff this file into a ConfigMap, the size matters (<1mb), I bet it does fit, just raising awareness.

Yep, that was one of my main considerations in doing this as a few lines of inline CSS (total page weighs in around 2.9K) instead of anything heavier. 

> I'd personally argue not to style this page, the point is to keep the page dead simple. Adding styles opens the door for all kinds of personal preferences.

Fair enough. I'll close the PR, thanks for the review :+1:

PR #1395, opened 2025-02-11
add withExprWrappersMixin

v-zhuravlev:
1) add flexible withExprWrappersMixin , will be used in snmp observ lib.
2) handle asTable() cases when table panel has only single target.


Dasomeone (approved):
Nice, good to see all these incremental improvements coming in :D

mshahzeb (approved):

PR #1397, opened 2025-02-12
Add support for newer mysqld-versions.

nachogiljaldo:
The current default version is fairly aged and there are newer ones with less vulnerabilities.

Unfortunately on https://github.com/prometheus/mysqld_exporter/tree/v0.15.0 the configs changed and it now requires a different way to pass the user, password, etc. (i.e. not just env variables).

I also went ahead and changed the default version to be 0.16.0 which is the latest. Added support for both vanilla (tag starts with v) and bitnami (tag does not start with v) images where the tag starts with v.


mysqld-exporter/main.libsonnet
```diff
@@ -27,7 +33,9 @@ local container = k.core.v1.container;
       + container.withPorts(k.core.v1.containerPort.new('http-metrics', 9104))
       + container.withArgsMixin([
         '--collect.info_schema.innodb_metrics',
-      ])
+      ] + if needsConfigFile then [
+        '--config.my-cnf=/conf/config.cnf',
+      ] else [])
```
jvrplmlmn, line 38: Note that the suggestion might not be `jsonnetfmt` compliant. Wrapping the if/else block in parenthesis removes potential the _footgun_ about extending the code and only applying to one branch of the conditional block:

```suggestion
      ] + (if needsConfigFile then [
        '--config.my-cnf=/conf/config.cnf',
      ] else [])
      )
```

mysqld-exporter/main.libsonnet
```diff
@@ -43,10 +51,25 @@ local container = k.core.v1.container;
             DATA_SOURCE_NAME: '$(MYSQL_USER):$(MYSQL_PASSWORD)@tcp($(MYSQL_HOST):$(MYSQL_PORT))/?tls=$(MYSQL_TLS_MODE)',
           }),
         ]
-      ),
+      ) + if needsConfigFile then configVolumeMount(configMapName, '/conf') else {},
```
jvrplmlmn, line 54: Note that the suggestion might not be `jsonnetfmt` compliant. Wrapping the if/else block in parenthesis removes potential the _footgun_ about extending the code and only applying to one branch of the conditional block:

```suggestion
      ) + (if needsConfigFile then configVolumeMount(configMapName, '/conf') else {}),
```

mysqld-exporter/main.libsonnet
```diff
@@ -43,10 +51,25 @@ local container = k.core.v1.container;
...
+                                             client: {
+                                               host: '${MYSQL_HOST}',
+                                               port: '${MYSQL_PORT}',
+                                               user: '${MYSQL_USER}',
+                                               password: '${MYSQL_PASSWORD}',
+                                               tls: '${MYSQL_TLS_MODE}',
+                                             },
+                                           },
+                                         }),
+                                       }) else {},
```
jvrplmlmn, line 72: Note that the suggestion might not be `jsonnetfmt` compliant. Wrapping the if/else block in parenthesis removes potential the _footgun_ about extending the code and only applying to one branch of the conditional block:

```suggestion
    configMap: (if needsConfigFile then configMap.new(configMapName)
                                       + configMap.withData({
                                         'config.cnf': std.manifestIni({
                                           sections: {
                                             client: {
                                               host: '${MYSQL_HOST}',
                                               port: '${MYSQL_PORT}',
                                               user: '${MYSQL_USER}',
                                               password: '${MYSQL_PASSWORD}',
                                               tls: '${MYSQL_TLS_MODE}',
                                             },
                                           },
                                         }),
                                       }) else {}
                                       ),
```

Nit: Could we also make this multi-line?

mysqld-exporter/main.libsonnet
```diff
@@ -9,11 +11,15 @@ local container = k.core.v1.container;
...
     port=3306,
-    image='prom/mysqld-exporter:v0.13.0',
+    image='prom/mysqld-exporter:v0.16.0',
     tlsMode='preferred',
+    configMapName='%s-cfg' % [name],
   ):: {
     local this = self,
-
+    local versionParts = std.split(std.split(image, ':')[1], '.'),
+    local majorVersion = std.parseInt(std.strReplace(versionParts[0], 'v', '')),
```
jvrplmlmn, line 20: > Added support for **bitnami images** where the tag starts with v.

For my own learning, is this a convention?

mysqld-exporter/main.libsonnet
```diff
@@ -9,11 +11,15 @@ local container = k.core.v1.container;
...
+    local minorVersion = std.parseInt(versionParts[1]),
+    local needsConfigFile = majorVersion > 0 || minorVersion > 14,
```
jvrplmlmn, line 22: Nit: Could we add a comment that explains why?

```suggestion
    // Ref: https://github.com/prometheus/mysqld_exporter/releases/tag/v0.15.0
    local needsConfigFile = majorVersion > 0 || minorVersion > 14,
```

jvrplmlmn (approved):
LG2M modulo a couple of jsonnet suggestions.

mysqld-exporter/main.libsonnet
nachogiljaldo: Sorry, I meant the other way around, bitnami images do not have v while the vanilla mysqld-exporter image has tags starting with v.

nachogiljaldo:
@jvrplmlmn thanks for the feedback, I applied it and corrected the description since the v prefix on the tag was the other way around (vanilla image has it, whereas bitnami one does not).

cinaglia (approved):

PR #1398, opened 2025-02-12
static-exporter: Automatically set http headers

Pokom:
Automatically set http headers so that static exporters are prometheus 3.0.0 compliant. The configuration can be overwritten by setting
  `config=null` when instantiating the object.


nikimanoledaki:
@Pokom this can be closed since it was implemented by https://github.com/grafana/jsonnet-libs/pull/1406 🥳

PR #1399, opened 2025-02-13
Add enableLokiLogs to commonlib signals and new log signal type

v-zhuravlev:
Add enableLokiLogs to commonlib signals and new log signal type.

### DIFF
--- a/common-lib/CHANGELOG.md
+++ b/common-lib/CHANGELOG.md
@@ -1,6 +1,7 @@
 # 0.3.0
 
-- [Signal] add new signal `log`. add enableLokiLogs=true|false to signals init
+- [Signal] add new signal `log`. 
+- [Signal] add enableLokiLogs=true|false to signals init.
 - [Signal] `withExprWrappersMixin(offset)` - wrap signal expression into additional function on top of existing wrappers.
 
 
--- a/common-lib/CHANGELOG.md
+++ b/common-lib/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.3.0
+
+- [Signal] add new signal `log`. add enableLokiLogs=true|false to signals init
+- [Signal] `withExprWrappersMixin(offset)` - wrap signal expression into additional function on top of existing wrappers.
+
+
 # 0.2.0
 
 - [Signal] `withOffset(offset)` - add offset modifier to the expression.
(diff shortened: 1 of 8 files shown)


common-lib/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.3.0
+
+- [Signal] add new signal `log`. add enableLokiLogs=true|false to signals init
```
Dasomeone, line 3: Might be worth splitting this on a multi-line entry, and making `enableLokiLogs` a code block

common-lib/common/signal/log.libsonnet
```diff
@@ -0,0 +1,68 @@
...
+          self.asPanelExpression(),
+        )
+        + lokiQuery.withRefId(name),
+
+      asGauge()::
+        error 'asGauge() is not supported for log metrics.',
+      asStat()::
+        error 'asStat() is not supported for log metrics.',
+      asTimeSeries()::
+        error 'asTimeSeries() is not supported for log metrics.',
```
Dasomeone, line 51: I know you're not using `asTimeSeries()` above on line 35, but it is still pulling the `timeSeries.panelOptions.withDescription` method, so I wonder if we want to pull generics like `withDescription` out on its own?

common-lib/common/signal/signal.libsonnet
```diff
@@ -31,8 +32,8 @@ local stub = import './stub.libsonnet';
   // DEPRECATED. Use unmarshallJsonMulti instead.
```
Dasomeone, line 32: Do we have an issue for tracking this?

Dasomeone (approved):
Couple comments but nothing blocking, looks good!

common-lib/CHANGELOG.md
```diff
@@ -1,3 +1,9 @@
+# 0.3.0
```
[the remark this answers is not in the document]
v-zhuravlev: I think line should be that functions are stable enough and I don't intend introduce breaking changes. For now, I keep this possibility.

common-lib/common/signal/log.libsonnet
v-zhuravlev: thanks, changed to logs for cleaner looks.

common-lib/common/signal/signal.libsonnet
v-zhuravlev: Not really. I think it should be safe to remove for now, i'll check this repo for uses.

common-lib/CHANGELOG.md
v-zhuravlev: updated

PR #1401, opened 2025-02-17
Snmp observ lib update: New cisco memory counter.

v-zhuravlev:
Add support for new counters introduced here: https://github.com/prometheus/snmp_exporter/pull/1374 
Also improves version parsing from sysDescr.


Dasomeone (approved):
Apologies Vitaly, I just noticed I hadn't hit approve so here goes:

Looks good to me Vitaly, great to see even more additions to this :D 
Do you have a tracker somewhere with the total you want to add?

PR #1404, opened 2025-02-20
feat: add serviceId label to csp-mixin alerts

anaivanov:
It adds a new label prop `serviceId` for each alert.

### DIFF
--- a/csp-mixin/alerts/gcp-alerts.yml
+++ b/csp-mixin/alerts/gcp-alerts.yml
@@ -144,7 +144,7 @@ groups:
         labels:
           severity: warning
           service: 'Pub/Sub'
-          service: 'pubsub'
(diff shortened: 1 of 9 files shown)


csp-mixin/alerts/gcp-alerts.yml
```diff
@@ -135,6 +144,7 @@ groups:
         labels:
           severity: warning
           service: 'Pub/Sub'
+          service: 'pubsub'
```
yduartep, line 147: Here should be **serviceId**

anaivanov: Good catch! https://github.com/grafana/jsonnet-libs/pull/1404/commits

yduartep (approved):
lgtm

PR #1407, opened 2025-02-21
Add Check for unexpected changes in generated files

v-zhuravlev:
Due to usage of external libs (grafonnet,commonlib,xtd, other observ-libs) sometimes there could be untracked/unexpected changes to the dashboards. To control it, there is a new CI check suggested that will check a diff between generation result and dashboards_out/alerts out files.

So in order to pass this check, in contributors PR there should be not only jsonnet changes, but also dashboards_out changes that can be created by running `make build`.

Also adds common Makefile to all mixins/observ-libs.

Note: diff is so large for this fix specifically. Would be much smaller thanks to gitattributes `linguist-generated=true` and the thing that PR normally changes one mixin at a time.)


v-zhuravlev:
Example issue detected:
https://github.com/grafana/jsonnet-libs/actions/runs/13464020444/job/37625731590
And fix:
https://github.com/grafana/jsonnet-libs/pull/1407/commits/cd876c0cd34afe3ae95c543fcb28957ff07f43f1

.gitattributes
```diff
@@ -0,0 +1,2 @@
+**/prometheus_rules_out/*.yaml linguist-generated=true
+**/dashboards_out/*.json linguist-generated=true
```
mshahzeb, line 2: Missing new line

Makefile
```diff
@@ -5,7 +5,7 @@ install-ci-deps:
 	go install github.com/google/go-jsonnet/cmd/jsonnet@v0.20.0
 	go install github.com/google/go-jsonnet/cmd/jsonnetfmt@v0.20.0
 	go install github.com/google/go-jsonnet/cmd/jsonnet-lint@v0.20.0
-	go install github.com/monitoring-mixins/mixtool/cmd/mixtool@a8984473edc33da685bd5f2138eab9cf6a0ff2b7
+	go install github.com/monitoring-mixins/mixtool/cmd/mixtool@7472b3fc310e9d1aaa82d423ae3e7e775a5879f8
```
mshahzeb, line 8: Shouldnt we pin to main?

mshahzeb (approved):
Great work! great addition.

Makefile
v-zhuravlev: I doubt, as it may break linting without a warning :)

.github/workflows/lint-mixins.yml
```diff
@@ -75,8 +75,12 @@ jobs:
...
+        run: make build
 
-      - name: Lint Mixin
+      - name: Lint Mixin with Mixtool
         working-directory: ./${{ matrix.mixin }}
         run: mixtool lint mixin.libsonnet
+
+      - name: Check for unexpected changes in generated files
+        working-directory: ./${{ matrix.mixin }}
+        run: make && git diff --exit-code
```
dimitarvdimitrov, line 86: nitpick, but maybe include a message for how to fix CI (e.g. `"Run `make` and commit the local diff"`)

v-zhuravlev: updated

.gitattributes
Dasomeone: These won't work unless Github has changed how Linguist work fundamentally. It doesn't support double use of `**`, we already attempted this last year

Makefile
Dasomeone: I would prefer to have this repo pinned to main actually, as a reminder that things change and always lint against latest (and what we tell people to install). it's a good check to people to realise their local install of mixtool may be out of date

.gitattributes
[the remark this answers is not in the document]
v-zhuravlev: image

Looks like it works?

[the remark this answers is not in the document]
v-zhuravlev: yes, not ideal, but at least something for the start.

Makefile
v-zhuravlev: ok, let's update for main.

PR #1408, opened 2025-02-22
Enhance windows observ lib with detailed alerts and add promtool tests.

v-zhuravlev:
Add windows alerts expanded descriptions. Add promtool testing in CI. And add documentation that such tests should be added for all alerts with complex queries/complex golang templates in descriptions.

Note: promtool tests in CI would fire when target branch is switched to master.


mshahzeb (approved):

.github/workflows/lint-mixins.yml
```diff
@@ -83,4 +83,8 @@ jobs:
 
       - name: Check for unexpected changes in generated files
         working-directory: ./${{ matrix.mixin }}
-        run: make && git diff --exit-code
+        run: make dashboards_out prometheus_alerts.yaml prometheus_rules.yaml && git diff --exit-code
```
Dasomeone, line 86: Do you want to add a warning or act on the diff in any way here?
Correct me if I'm wrong but this will fire for every mixin that doesn't have generated dashboards present?

Dasomeone (commented):
Generally looks really nice, just one outstanding comment :)

.github/workflows/lint-mixins.yml
v-zhuravlev: yes, see #1407.

v-zhuravlev: I included all of them in #1407 but consul-mixin due to another reason. I'll add it soon.

PR #1428, opened 2025-03-26
fix azure front door dimensions

yduartep:
fix azure front door dimensions

**Now**:
Screenshot 2025-03-26 at 09 53 51

**After the fix**:
Screenshot 2025-03-26 at 09 57 44


anaivanov (approved):
LGTM

PR #1439, opened 2025-05-02
Update nginx versions

zalegrala:
These nginx versions are quite old.  Perhaps updating the default version is valuable for folks who don't specify.


zalegrala:
I see the failing test, but I'm not sure what to make of it.  The one I currently care about updating is the directory.  I can back out the others if this can help move it forward.

PR #1441, opened 2025-05-07
fix(mixin-utils): remove std.prune and use std.filter instead

Duologic:
The use of `std.prune(std.map())` is slow and uses a big stack, on big
rulesets this can cause a 'RUNTIME ERROR: max stack frames exceeded.'.

It is better to use `std.filter()` to remove an element from an array.

### DIFF
--- a/mixin-utils/test/test_remove_rules.libsonnet
+++ b/mixin-utils/test/test_remove_rules.libsonnet
@@ -1,0 +1,284 @@
+local utils = import '../utils.libsonnet';
+local test = import 'github.com/jsonnet-libs/testonnet/main.libsonnet';
+
+local config = {
+  prometheusAlerts: {
+    groups: [
+      {
+        name: 'group1',
+        rules: [
+          {
+            alert: 'alert1',
+          },
+          {
+            alert: 'alert2',
+          },
+        ],
+      },
+      {
+        name: 'group2',
+        rules: [
+          {
+            alert: 'alert3',
+          },
+          {
+            alert: 'alert4',
+          },
+        ],
+      },
+    ],
+  },
+  prometheusRules: {
+    groups: [
+      {
+        name: 'group3',
+        rules: [
+          {
+            record: 'record1',
+          },
+          {
+            record: 'record2',
+          },
+        ],
+      },
+      {
+        name: 'group4',
+        rules: [
+          {
+            record: 'record3',
+          },
+          {
+            record: 'record4',
+          },
+        ],
+      },
+    ],
+  },
+};
+
+test.new(std.thisFile)
+
++ test.case.new(
+  'removeAlertRuleGroup',
+  test.expect.eq(
+    actual=config + utils.removeAlertRuleGroup('group1'),
+    expected={
+      prometheusAlerts: {
+        groups: [
+          {
+            name: 'group2',
+            rules: [
+              {
+                alert: 'alert3',
+              },
+              {
+                alert: 'alert4',
+              },
+            ],
+          },
+        ],
+      },
+      prometheusRules: {
+        groups: [
+          {
+            name: 'group3',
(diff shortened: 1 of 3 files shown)


mixin-utils/utils.libsonnet
```diff
@@ -411,20 +409,17 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
-    local overrideInGroup(group) = group { rules: std.map(overrideRule, super.rules) },
+    local overrideInGroup(group) = group + { rules: std.map(overrideRule, super.rules) },
     prometheusAlerts+:: {
       groups: std.map(overrideInGroup, super.groups),
     },
   },
 
   removeAlerts(alerts):: {
-    local removeRule(rule) =
-      if 'alert' in rule && std.objectHas(alerts, rule.alert)
```
Duologic, line 422: Oh, I'm breaking something here, this function takes an object as input, not an array...

Duologic: covered, it is not backwards compatible.

mixin-utils/utils.libsonnet
```diff
@@ -411,20 +409,21 @@ local g = import 'grafana-builder/grafana.libsonnet';
       if 'alert' in rule && std.objectHas(overrides, rule.alert)
       then rule + overrides[rule.alert]
       else rule,
-    local overrideInGroup(group) = group { rules: std.map(overrideRule, super.rules) },
+    local overrideInGroup(group) = group + { rules: std.map(overrideRule, super.rules) },
```
pstibrany, line 412: super-nit: seems like unnecessary change that jsonfmt should remove (?)

pstibrany (approved):
Changes lgtm. (Could it be possible to have some tests? Don't block this PR on it though.)

mixin-utils/utils.libsonnet
```diff
@@ -392,18 +392,16 @@ local g = import 'grafana-builder/grafana.libsonnet';
...
-    local removeRuleGroup(rule) = if rule.name == ruleName then null else rule,
-    local currentRuleGroups = super.groups,
-    groups: std.prune(std.map(removeRuleGroup, currentRuleGroups)),
+  removeRuleGroup(groupName):: {
+    groups: std.filter(function(group) group.name != groupName, super.groups),
   },
 
   removeAlertRuleGroup(ruleName):: {
-    prometheusAlerts+:: $.removeRuleGroup(ruleName),
+    prometheusAlerts+: $.removeRuleGroup(ruleName),
```
Duologic, line 400: `+:` means 'keep visiblity as it was, so if the field was hidden upstream, then it will stay hidden.

mixin-utils/test/test_remove_rules.libsonnet
```diff
@@ -0,0 +1,284 @@
...
+    ],
+  },
+};
+
+test.new(std.thisFile)
+
++ test.case.new(
+  'removeAlertRuleGroup',
+  test.expect.eq(
+    actual=config + utils.removeAlertRuleGroup('group1'),
```
pstibrany, line 64: Thank you for adding tests. This also shows how these functions are supposed to be used! That's very useful!

mixin-utils/test/test_remove_rules.libsonnet
```diff
@@ -0,0 +1,284 @@
...
++ test.case.new(
+  'removeRecordingRuleGroup with groupname from alert rules (noop)',
+  test.expect.eq(
+    actual=config + utils.removeRecordingRuleGroup('group2'),
+    expected=config
+  )
+)
+
++ test.case.new(
+  'removeAlerts',
```
Duologic, line 177: I ran all but this test case against main. This test case represents a new feature, `removeAlerts` now accepts an array.

PR #1442, opened 2025-05-12
feat(tanka-utils): add `isKubernetesObject` test and `findKubernetesObject` utilities

alexander-bauer:
I was working on my home cluster last week, and ran into a case where I needed to traverse a rendered Helm chart in order to generate ServiceMonitor resources for each Service. The existing `patchKubernetesObjects` function couldn't quite cover this case, so I wrote a similar function, and figured it would do best in this library.

I also extracted the `isKubernetesObject` logic common to both functions, in the hopes that it'll also be independently useful.

### DIFF
--- a/tanka-util/k8s.libsonnet
+++ b/tanka-util/k8s.libsonnet
@@ -2,6 +2,19 @@ local d = import 'github.com/jsonnet-libs/docsonnet/doc-util/main.libsonnet';
 {
   local this = self,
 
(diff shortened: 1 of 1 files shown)


Duologic:
There's also this utility which has a function for finding kubernetes objects: https://github.com/jsonnet-libs/xtd/blob/master/docs%2Finspect.md

PR #1444, opened 2025-05-22
drop snmp non fc alerts

v-zhuravlev:
Remove FC alerts that are not actually problems.


aditya-konarde (commented):
There's a tradeoff between the level of noise and having this info for correlation. I'm wondering if dropping to `info` would be a better first-step. 

Edit: less noise is a good idea, let's iterate and add back as `info` if the community feels strongly about this.

aditya-konarde (approved):
lgtm

PR #1445, opened 2025-05-27
Update ClickHouse mixin to better utilize libraries

schmikei:
- Utilizing `common-lib`, `logs-lib` and `grafonnet` for most of the mixin generation
- Following inspiration for more recently updated mixins like Vsphere

We need to retake some screenshots I believe for this update as now logs are utilizing the logs-lib and will be on a separate dashboard now instead of baked into the overview.

Clickhouse Overview
image

Clickhouse Latency
image

Clickhouse replica
image

Logs
image


clickhouse-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,142 @@
...
+        + g.dashboard.withPanels(
+          g.util.grid.wrapPanels(
+            [
+              panels.interserverConnectionsPanel { gridPos+: { h: 8, w: 12, x: 0, y: 0 } },
+              panels.replicaQueueSizePanel { gridPos+: { h: 8, w: 12, x: 12, y: 0 } },
+              panels.replicaOperationsPanel { gridPos+: { h: 8, w: 12, x: 0, y: 8 } },
+              panels.replicaReadOnlyPanel { gridPos+: { h: 8, w: 12, x: 12, y: 8 } },
+              panels.zooKeeperWatchesPanel { gridPos+: { h: 8, w: 12, x: 0, y: 16 } },
+              panels.zooKeeperSessionsPanel { gridPos+: { h: 8, w: 12, x: 12, y: 16 } },
+              panels.zooKeeperRequestsPanel { gridPos+: { h: 8, w: 24, x: 0, y: 24 } },
```
v-zhuravlev, line 29: x,y coordinates pair is redundant if wrapPanels is used. It will be calculated.

v-zhuravlev (commented):
Thanks for contributing! Small comment above while still a draft.

clickhouse-mixin/rows.libsonnet
```diff
@@ -0,0 +1,86 @@
+local g = import './g.libsonnet';
+
+{
+  new(panels): {
+    row_1:
```
v-zhuravlev: seems like rows are not used to build dashboards.

clickhouse-mixin/targets.libsonnet
```diff
@@ -0,0 +1,236 @@
...
+      + g.panel.timeSeries.queryOptions.withInterval('30s')
+      + prometheusQuery.withLegendFormat('%s - Failed Insert Query' % utils.labelsToPanelLegend(config.legendLabels)),
+
+    ClickHouseProfileEvents_RejectedInserts:
+      prometheusQuery.new(
+        '${' + vars.datasources.prometheus.name + '}',
+        'rate(ClickHouseProfileEvents_RejectedInserts{%s}[$__rate_interval])' % selector
+      )
+      + g.panel.timeSeries.queryOptions.withInterval('30s')
+      + prometheusQuery.withLegendFormat('%s - Rejected Inserts' % utils.labelsToPanelLegend(config.legendLabels)),
```
v-zhuravlev:
```suggestion
      + prometheusQuery.withLegendFormat('%s - Rejected Inserts' % utils.labelsToPanelLegend(config.legendLabels)),
```
here and multiple places. 
Please use Sentence case.

clickhouse-mixin/targets.libsonnet
```diff
@@ -0,0 +1,236 @@
...
+        '${' + vars.datasources.prometheus.name + '}',
+        'rate(ClickHouseProfileEvents_NetworkSendBytes{%s}[$__rate_interval])' % selector
+      )
+      + g.panel.timeSeries.queryOptions.withInterval('30s')
+      + prometheusQuery.withLegendFormat('%s - network send bytes' % utils.labelsToPanelLegend(config.legendLabels)),
+
+    ClickHouseProfileEvents_DiskReadElapsedMicroseconds:
+      prometheusQuery.new(
+        '${' + vars.datasources.prometheus.name + '}',
+        'increase(ClickHouseProfileEvents_DiskReadElapsedMicroseconds{%s}[$__rate_interval])' % selector
```
v-zhuravlev:
```suggestion
        'increase(ClickHouseProfileEvents_DiskReadElapsedMicroseconds{%s}[$__rate_interval])' % selector
```

to get accurate increments, Increase() should be used with `[$__interval:] -offset $__interval`. 
rate() is best used with $__rate_interval, but it gives per second ratio.

clickhouse-mixin/variables.libsonnet
v-zhuravlev, line 1: this whole file could be ommited. Its contents could generated with `commonlib.variables.new()`

clickhouse-mixin/jsonnetfile.json
```diff
@@ -4,20 +4,20 @@
...
       },
       "version": "master"
     },
     {
       "source": {
         "git": {
-          "remote": "https://github.com/grafana/jsonnet-libs.git",
-          "subdir": "grafana-cloud-integration-utils"
+          "remote": "https://github.com/grafana/grafonnet.git",
+          "subdir": "gen/grafonnet-v10.0.0"
```
v-zhuravlev:
```suggestion
          "subdir": "gen/grafonnet-v11.0.0"
```
best to use  v11.

clickhouse-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,142 @@
...
+    local uid = g.util.string.slugify(this.config.uid);
+    local vars = this.grafana.variables;
+    local annotations = this.grafana.annotations;
+    local refresh = this.config.dashboardRefresh;
+    local period = this.config.dashboardPeriod;
+    local timezone = this.config.dashboardTimezone;
+    local panels = this.grafana.panels;
+
+    {
+      'clickhouse-replica':
```
v-zhuravlev:
```suggestion
      'clickhouse-replica.json':
```
.json suffix should be used for dashboards.

clickhouse-mixin/config.libsonnet
```diff
@@ -1,20 +1,26 @@
 {
-  _config+:: {
-    enableMultiCluster: false,
-    clickhouseSelector: if self.enableMultiCluster then 'job=~"$job", instance=~"$instance", cluster=~"$cluster"' else 'job=~"$job", instance=~"$instance"',
+  enableMultiCluster: false,
```
v-zhuravlev:
```suggestion
  enableMultiCluster: false,
```

overall, I think cluster could be considered a specific case of a member of groupLabel. Do we need any specific support for clusterLabel?

clickhouse-mixin/targets.libsonnet
```diff
@@ -0,0 +1,236 @@
+local g = import './g.libsonnet';
+local prometheusQuery = g.query.prometheus;
+local commonlib = import 'common-lib/common/main.libsonnet';
+local utils = commonlib.utils {
+  labelsToPanelLegend(labels): std.join(' - ', ['{{%s}}' % [label] for label in labels]),
+};
```
v-zhuravlev: I think this function is present in commonlib.utils now?

clickhouse-mixin/targets.libsonnet
v-zhuravlev, line 1: Is it possible to transform targets to signals?

v-zhuravlev (changes requested):
Thanks! Some comments inline.

clickhouse-mixin/config.libsonnet
schmikei: I'm curious about this as well. In this effort I'm mostly trying to refactor over to the new formats in the least destructive way. I've noticed that for like velero-2 we've not had this configuration field in config.libsonnet, but for others like pgbouncer we do something similar...

I'm not quite sure what the ideal preference is here, but the `enableMultiCluster` seems to at least make initial sense to me but happy to explore what is the preferred approach!

schmikei: Apologies as its been a while since fully completing one of these, but let me know what the ideal for this would be, it seems like we should keep the config enableMultiCluster in my opinion but I may be missing something.

If there's a good example or guidance I can go off of I'm happy to make further changes! Otherwise I believe this is probably what we'd like?

schmikei (commented):
Made the cutover to signals pretty smoothly 🤞 hoping to get some more direction around the `cluster` label/where that logic should go 🤔 thanks again for the reviews and guiding me back on the rails

clickhouse-mixin/signals/memory.libsonnet
```diff
@@ -0,0 +1,43 @@
...
+          prometheus: {
+            expr: 'ClickHouseMetrics_MemoryTracking{%(queriesSelector)s}',
+            legendCustomTemplate: '{{ instance }} - Memory tracking',
+          },
+        },
+      },
+      memoryUsagePercent: {
+        name: 'Memory usage percentage',
+        nameShort: 'Memory %',
+        type: 'gauge',
```
v-zhuravlev:
```suggestion
        type: 'raw',
```
best to use raw if you have actual PromQL expression (not just simple metric) in expression.

clickhouse-mixin/signals/queries.libsonnet
```diff
@@ -0,0 +1,100 @@
...
+        name: 'Failed select queries',
+        nameShort: 'Failed SELECT',
+        type: 'counter',
+        description: 'Rate of failed SELECT queries per second',
+        unit: '/ sec',
+        sources: {
+          prometheus: {
+            expr: 'ClickHouseProfileEvents_FailedSelectQuery{%(queriesSelector)s}',
+            legendCustomTemplate: '{{ instance }} - Failed select query',
+            interval: '30s',
```
v-zhuravlev:
```suggestion
            interval: '30s',
```
this doesn't do anything I think.

clickhouse-mixin/signals/queries.libsonnet
```diff
@@ -0,0 +1,100 @@
...
+          },
+        },
+      },
+      failedInsertQueries: {
+        name: 'Failed insert queries',
+        nameShort: 'Failed INSERT',
+        type: 'counter',
+        description: 'Rate of failed INSERT queries per second',
```
v-zhuravlev:
```suggestion
        description: 'Rate of failed INSERT queries per second.',
```
Here and everywhere else(descriptions). Shouldn't we end with dot?

clickhouse-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,142 @@
...
+              panels.interserverConnectionsPanel { gridPos+: { w: 12 } },
+              panels.replicaQueueSizePanel { gridPos+: { w: 12 } },
+              panels.replicaOperationsPanel { gridPos+: { w: 12 } },
+              panels.replicaReadOnlyPanel { gridPos+: { w: 12 } },
+              panels.zooKeeperWatchesPanel { gridPos+: { w: 12 } },
+              panels.zooKeeperSessionsPanel { gridPos+: { w: 12 } },
+              panels.zooKeeperRequestsPanel { gridPos+: { w: 24 } },
+            ]
+          )
+        )
```
v-zhuravlev: Could you add rows.libsonnet, group rows there and then reference rows here? That would make it possible to reuse rows as logical group of rows in other combined dashboards.

clickhouse-mixin/panels.libsonnet
```diff
@@ -0,0 +1,216 @@
...
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('decbytes')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withFillOpacity(0)
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withSpanNulls(false),
+
+      memoryUsageGaugePanel:
+        g.panel.gauge.new('Memory usage')
+        + g.panel.gauge.queryOptions.withTargets([signals.memory.memoryUsagePercent.asTarget()])
+        + g.panel.gauge.queryOptions.withDatasource('prometheus', '${' + this.grafana.variables.datasources.prometheus.name + '}')
+        + g.panel.gauge.panelOptions.withDescription('Percentage of memory allocated by ClickHouse compared to OS total')
```
v-zhuravlev:
```suggestion
        + g.panel.gauge.queryOptions.withTargets([signals.memory.memoryUsagePercent.asTarget()])
        + g.panel.gauge.panelOptions.withDescription('Percentage of memory allocated by ClickHouse compared to OS total')
```
This shouldn't be needed anymore.

clickhouse-mixin/config.libsonnet
v-zhuravlev: I think cluster label is just yet another groupLabel. It is only unique in a way that it could be missing (in cloud integrations context) in metrics , so allValue filter should be `.*` not `.+` so dashboards are not empty for "clusterless" sources.. Also this is more about integrations context and not really about mixin.

@Dasomeone , what do you think? I personal lean towards removing it everywere and finding a generic way to set some of the groupLabels "optional".

Dasomeone: Fully agreed here, let's chat with the k8s monitoring team to see if there's still a use for this `enableMultiCluster` flag, but I don't believe we use it at all in integrations. In fact I believe it was a pre-emptive setup for k8s support back in the day but it's not really in use? I don't think?
Additionally on the `cluster` and `groupLabels` I think it makes sense to check with the k8s team in case they have any reservations w. regards to removing `cluster`, but I'm personally in favour of making `groupLabels` the default everywhere, with some optionality allowed

clickhouse-mixin/main.libsonnet
```diff
@@ -0,0 +1,49 @@
...
+        for sig in std.objectFields(this.config.signals)
+      },
+
+    grafana: {
+      variables: commonlib.variables.new(
+        filteringSelector=this.config.filteringSelector,
+        groupLabels=this.config.groupLabels,
+        instanceLabels=this.config.instanceLabels,
+        varMetric='ClickHouseMetrics_InterserverConnection',
+        customAllValue='.*',
```
schmikei: This generally works but since all variables are now `.*` instead of `.+` I just wanted to confirm thats okay even though we mostly wanted this for `cluster` specifically. Or if there was an easier way of specifying that specific behavior?

v-zhuravlev: you should use `.+` here.

as of cluster `.+`, we can replace cluster value to `.*` with function
from 
"grafana-cloud-integration-utils/util.libsonnet":



```
  // This function can be used to patch dashboards variables, matching by name:
  // patch format:
  // {
  //   cluster: {
  //     allValues: ".*"
  //   },
  // }
  patch_variables(dashboard, patch)::
    {
      templating: {
        list+: [
          t + std.get(patch, t.name, default={})
          for t in dashboard.templating.list
        ],
      },
    },
```

it can be used on the integration level like in kafka (mixin.libsonnet):

```
  grafanaDashboards+::
    {
      local tags = config._config.dashboardTags,

      [fname]:
        local dashboard = util.decorate_dashboard(mixin.grafanaDashboards[fname], tags=tags)
                          + util.add_links_by_tag(tags);

        dashboard
        + util.patch_variables(dashboard, k8s_patch)


      for fname in std.objectFields(mixin.grafanaDashboards)
    },
```

where k8s_patch:
```
{
  cluster+: {
    allValue: '.*',
  },
}

```

clickhouse-mixin/config.libsonnet
v-zhuravlev: let's remove it and use k8s_patch instead suggested below.

clickhouse-mixin/main.libsonnet
v-zhuravlev: it is not necessary to use k8s patch here, since this is irrelevant on mixin level... this snippet should be applied on integration level though.

clickhouse-mixin/rows.libsonnet
```diff
@@ -0,0 +1,47 @@
...
+local panels = import './panels.libsonnet';
+
+{
+  new(this): {
+    replicaOperationsPanels: [
+      this.grafana.panels.interserverConnectionsPanel { gridPos+: { w: 12 } },
+      this.grafana.panels.replicaQueueSizePanel { gridPos+: { w: 12 } },
+      this.grafana.panels.replicaOperationsPanel { gridPos+: { w: 12 } },
+      this.grafana.panels.replicaReadOnlyPanel { gridPos+: { w: 12 } },
+    ],
```
v-zhuravlev: We found this format more appealing, as it allows to import separate rows and make them collapsed/uncollapsed later:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/rows.libsonnet#L5-L24
on the dashboard you need to add extra function:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/dashboards.libsonnet#L34

schmikei: I suppose this would change the dashboard's layout that would divert from the spec? The intention for this PR at least was to use the modular libs without changing the structure too much of the integration. 

Perhaps in a follow-up we could have actual row containers? Unless you feel like we should make this change in this PR or a quick follow-on PR?

schmikei: Ahh I wasn't sure if we wanted to update "structure" of the mixin while we were here. I personally think that the dashboard probably looks better this way, so I went ahead and made the change.

image

PR #1450, opened 2025-06-11
MSSQL Modernization

schmikei:
- Utilizing common-lib, logs-lib and grafonnet for most of the mixin generation
- Following inspiration for more recently updated mixins like Vsphere

This involves separating out the MSSQL Overview dashboard to separate logs and metrics. This would require some new screenshots. 

image
image
image


mssql-mixin/rows.libsonnet
```diff
@@ -0,0 +1,17 @@
...
+      [
+        g.panel.row.new('Database'),
+        this.grafana.panels.databaseWriteStallDurationPanel
+        + g.panel.timeSeries.gridPos.withW(12),
+        this.grafana.panels.databaseReadStallDurationPanel
+        + g.panel.timeSeries.gridPos.withW(12),
+        this.grafana.panels.transactionLogExpansionsPanel
+        + g.panel.timeSeries.gridPos.withW(24),
+      ],
+  },
```
v-zhuravlev: best if row panels are added not as array but inside
`g.panel.row.withPanels()` function. This way it is possible to toggle rows collapsed/uncollapsed.

example:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/rows.libsonnet#L25-L33.
and then:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/dashboards.libsonnet#L33-L52

mssql-mixin/dashboards_out/mssql_pages
Dasomeone, line 1: Missing .json filetype on this and the mssql_overview files

mssql-mixin/dashboards_out/mssql_pages
```diff
@@ -0,0 +1,300 @@
...
+            "keepTime": true,
+            "title": "Logs",
+            "type": "link",
+            "url": "/d/mssql-logs"
+         },
+         {
+            "keepTime": true,
+            "title": "MSSQL Overview",
+            "type": "link",
+            "url": "/d/mssql_mssql_overview"
```
Dasomeone: I know this is generated, but it feels like a weird change to have mssql twice in the URL?

schmikei:
@v-zhuravlev just curious, do you think we should just default to using signals for this PR as well? I think the API is nice to use and gives us more flexibility probably moving forward.

mssql-mixin/config.libsonnet
```diff
@@ -1,21 +1,39 @@
...
-    multiclusterSelector: 'job=~"$job"',
-    mssqlSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
-    dashboardTags: ['mssql-mixin'],
-    dashboardPeriod: 'now-1h',
-    dashboardTimezone: 'default',
-    dashboardRefresh: '1m',
+  local this = self,
+  enableMultiCluster: false,
+  filteringSelector: 'job=~"integrations/mssql"',
+  groupLabels: ['job', 'cluster'],  // Remove 'db' from here
```
v-zhuravlev: looks like db is removed.

mssql-mixin/config.libsonnet
```diff
@@ -1,21 +1,39 @@
...
+
+  // additional params
+  dashboardPeriod: 'now-1h',
+  dashboardTimezone: 'default',
+  dashboardRefresh: '1m',
+  metricsSource: 'prometheus',
+
+  // logs lib related
+  enableLokiLogs: true,
+  logLabels: ['job', 'cluster', 'instance'],
```
v-zhuravlev:
```suggestion
  logLabels: self.groupLabels + self.instanceLabels,
```
would it be a better default?

mssql-mixin/jsonnetfile.json
```diff
@@ -4,12 +4,39 @@
...
+          "subdir": "common-lib"
+        }
+      },
+      "version": "master"
+    },
+    {
+      "source": {
+        "git": {
+          "remote": "https://github.com/grafana/grafonnet.git",
+          "subdir": "gen/grafonnet-v10.0.0"
```
v-zhuravlev: could it be bumped to 11.0.0 or 11.4.0?

mssql-mixin/main.libsonnet
```diff
@@ -0,0 +1,72 @@
...
+            type: 'prometheus',
+            uid: '${prometheus_datasource}',
+          },
+          refresh: 2,
+          sort: 1,
+          type: 'query',
+          includeAll: true,
+          allValue: '.*',
+          multi: true,
+        };
```
v-zhuravlev: why do you need special value here?

mssql-mixin/signals/connections.libsonnet
```diff
@@ -0,0 +1,70 @@
...
+        type: 'raw',
+        description: 'Rate of severe connection errors per second.',
+        unit: '/ sec',
+        sources: {
+          prometheus: {
+            expr: 'increase(mssql_kill_connection_errors_total{%(queriesSelector)s}[$__rate_interval:])',
+            legendCustomTemplate: '{{ instance }}',
+          },
+        },
+      },
```
v-zhuravlev: wouldn't it be simplier to use type=counter and simple expression?
Alos, AFAIK, increase doesn't work well with rate_interval. Best if combined just $__interval to avoid increments duplication and wrong "Total" aggregations in the legend etc.

mssql-mixin/signals/connections.libsonnet
```diff
@@ -0,0 +1,70 @@
...
+        name: 'Deadlocks',
+        nameShort: 'Deadlocks',
+        type: 'raw',
+        description: 'Rate of deadlocks per second.',
+        unit: '/ sec',
+        sources: {
+          prometheus: {
+            expr: 'rate(mssql_deadlocks_total{%(queriesSelector)s}[$__rate_interval])',
+            legendCustomTemplate: '{{ instance }}',
+          },
```
v-zhuravlev: same, why not to use type=counter?

mssql-mixin/signals/database.libsonnet
```diff
@@ -0,0 +1,58 @@
...
+    signals: {
+      databaseWriteStallDuration: {
+        name: 'Database write stall duration',
+        nameShort: 'Write Stall',
+        type: 'raw',
+        description: 'Time spent waiting for database write operations.',
+        unit: 'seconds',
+        sources: {
+          prometheus: {
+            expr: 'increase(mssql_io_stall_seconds_total{%(queriesSelector)s, db=~"$db", operation="write"}[$__rate_interval:])',
```
v-zhuravlev: can `db` just added as the last element of instanceLabels? Can it be considered the smallest instnance you want to observ (single db within MSSQL host?) would it work?

mssql-mixin/signals/database.libsonnet
```diff
@@ -0,0 +1,58 @@
...
+            legendCustomTemplate: '{{ instance }} - {{ db }}',
+          },
+        },
+      },
+      databaseReadStallDuration: {
+        name: 'Database read stall duration',
+        nameShort: 'Read Stall',
```
v-zhuravlev:
```suggestion
        nameShort: 'Read stall',
```
Please use Sentence case.

mssql-mixin/signals/database.libsonnet
```diff
@@ -0,0 +1,58 @@
...
+        type: 'raw',
+        description: 'Time spent waiting for database read operations.',
+        unit: 'seconds',
+        sources: {
+          prometheus: {
+            expr: 'increase(mssql_io_stall_seconds_total{%(queriesSelector)s, db=~"$db", operation="read"}[$__rate_interval:])',
```
v-zhuravlev:
```suggestion
        type: 'counter',
        description: 'Time spent waiting for database read operations.',
        unit: 'seconds',
        sources: {
          prometheus: {
            expr: 'mssql_io_stall_seconds_total{%(queriesSelector)s, db=~"$db", operation="read"})',
```
again, I would defined counters as type=counters and simple expression. See signals README, you can swap rate to increase() as the default function that is applied to the counter if needed.

mssql-mixin/signals/memory.libsonnet
```diff
@@ -0,0 +1,121 @@
...
+        sources: {
+          prometheus: {
+            expr: 'mssql_os_page_file{%(queriesSelector)s}',
+            legendCustomTemplate: '{{ instance }} - {{ state }}',
+          },
+        },
+      },
+      bufferCacheHitPercentage: {
+        name: 'Buffer cache hit percentage',
+        nameShort: 'Buffer Cache Hit',
```
v-zhuravlev: Here and similar, please check it is Sentence case.

v-zhuravlev (commented):
Great work, some comments inline.

mssql-mixin/main.libsonnet
schmikei: I implemented this because only some targets need the `db` field as part of the filter in the original design. 

Old Source:
https://github.com/grafana/jsonnet-libs/blob/1cd23af7f31e8ee2f4e6abe16b466c7bfc0ce82a/mssql-mixin/dashboards/mssql-overview.libsonnet#L508
vs
https://github.com/grafana/jsonnet-libs/blob/1cd23af7f31e8ee2f4e6abe16b466c7bfc0ce82a/mssql-mixin/dashboards/mssql-overview.libsonnet#L590

Dropdown:
https://github.com/grafana/jsonnet-libs/blob/1cd23af7f31e8ee2f4e6abe16b466c7bfc0ce82a/mssql-mixin/dashboards/mssql-overview.libsonnet#L892-L901

As you can see we got around this last time by selectively using the filter on the specific targets we needed. 

With signals it seems a tad bit harder to specify these kind of scenarios where we want a filter dropdown but also want all the filters aside from the `db` selector. So changing the default value to `.*` makes it so that the filter still works but does not break other panels when filtering. 

Let me know if you think there's an easier way of accomplishing this but this was my solution thus far to keep this going forward!

v-zhuravlev: you just need db var to accept `.*` as all value?

mssql-mixin/dashboards_out/mssql_overview.json
```diff
@@ -192,7 +192,7 @@
                      "type": "prometheus",
                      "uid": "${prometheus_datasource}"
                   },
-                  "expr": "increase(mssql_kill_connection_errors_total{job=~\"integrations/mssql\",job=~\"$job\",cluster=~\"$cluster\",instance=~\"$instance\"}[$__rate_interval:])",
+                  "expr": "rate(mssql_kill_connection_errors_total{job=~\"integrations/mssql\",job=~\"$job\",cluster=~\"$cluster\",instance=~\"$instance\"}[$__rate_interval])",
```
v-zhuravlev: some expr got changed function from increase to rate at the end. Just to check that this is what you want. If not, you can change rateFunction to be increase for some signals. See Signals README.

mssql-mixin/main.libsonnet
schmikei: Right but its moreso that if any database filter is applied then it kind of breaks the other panels if I do it as an instanceLabel...

I suppose is this behavior fine from your perspective? I didn't want to remove the funcitionality of this being a single pane where a user could correlate database data with instance data.

https://github.com/user-attachments/assets/5b69d4af-a35a-41ea-8316-83bc16ce161e

if that doesn't make sense let me know but I figured the currently written way was more user friendly, but happy to make the change either way!

v-zhuravlev: ok got it. You check this example with Kafka instance and Kafka instance topics, very similar to yours.


1) First topic is added(hardcoded) as instance label to topics signals.
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/signals/topic.libsonnet#L7
note that for your case it could be instanceLabels + ['db'], there instanceLabels are ommitted for completely different reason.
2) Second, union of variables generated is used give variables for topics dashboard:
https://github.com/grafana/jsonnet-libs/blob/master/kafka-observ-lib/dashboards.libsonnet#L7-L13

schmikei: I believe I've implemented based off what you're saying. Let me know if you think its adequate!

mssql-mixin/dashboards_out/mssql_overview.json
schmikei: Thanks for the catch 👍 should be resolved now

mssql-mixin/prometheus_alerts.yaml
v-zhuravlev, line 1: Is this a duplicate?

mssql-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,114 @@
...
+    local timezone = this.config.dashboardTimezone;
+
+    {
+      'mssql_overview.json':
+        g.dashboard.new(prefix + ' overview')
+        + g.dashboard.withVariables(
+          std.setUnion(
+            this.signals.database.getVariablesMultiChoice(),
+            keyF=function(x) x.name
+          )
```
v-zhuravlev: Sorry, I think missguided you here ,

But having just one variables set in union() is not doing anything. I guess it could be removed, or you need to combine variables of signals.database and other signals group you use on that dashboard.

mssql-mixin/dashboards_out/mssql_overview.json
```diff
@@ -682,10 +674,18 @@
...
                "refresh": 2,
                "sort": 1,
                "type": "query"
+            },
+            {
+               "hide": 2,
+               "label": "Loki data source",
+               "name": "loki_datasource",
+               "query": "loki",
+               "regex": "(?!grafanacloud.+usage-insights|grafanacloud.+alert-state-history).+",
```
v-zhuravlev: Where is regex coming from? It doesn't make sense on the mixin level.

schmikei: I believe this is coming from our decorate_dashboard call in main.libsonnet:

```
  grafanaDashboards+:: {
    local tags = config.dashboardTags,
    [fname]:
      local dashboard = util.decorate_dashboard(mssql.grafana.dashboards[fname], tags=tags);
      dashboard + util.patch_variables(dashboard, optional_labels)

    for fname in std.objectFields(mssql.grafana.dashboards)
  },
```


https://github.com/grafana/jsonnet-libs/blob/1cd23af7f31e8ee2f4e6abe16b466c7bfc0ce82a/grafana-cloud-integration-utils/util.libsonnet#L119-L143


Is it preferrable to remove the decorate?

v-zhuravlev: I think decorate is not needed on mixin level.

PR #1452, opened 2025-06-18
feat: add netflow-mixin

theSuess:
This mixin is intended to be used with tools like
[ktranslate](https://github.com/kentik/ktranslate) but will also work with other tooling, as long as the metrics conform to the [networking semantic conventions](https://opentelemetry.io/docs/specs/semconv/registry/attributes/network/).


netflow-mixin/dashboards/netflow-overview.libsonnet
```diff
@@ -0,0 +1,88 @@
...
+    'netflow-overview.json':
+      g.dashboard.new('Netflow overview')
+      + g.dashboard.withUid(dashboardUid)
+      + g.dashboard.withTags($._config.dashboardTags)
+      + g.dashboard.time.withFrom($._config.dashboardPeriod)
+      + g.dashboard.withRefresh($._config.dashboardRefresh)
+      + g.dashboard.withTimezone($._config.dashboardTimezone)
+      + g.dashboard.withVariables([
+        var.datasource.new('prometheus_datasource', 'prometheus')
+        + var.datasource.withRegex('(ops|dev)-cortex')
```
v-zhuravlev: this filter looks a bit specific.

PR #1459, opened 2025-07-01
feat: add withLabels function to utils for enhanced rule labeling

andriikushch:
This PR adds a new utility function withLabels() to the mixin-utils library that allows adding labels to alert rules.

**Features**

- Add labels to all alerts in rule groups
- Optionally filter which alerts receive the labels using a custom filter function
- Preserves existing labels on alerts

Usage example:

```jsonnet
local utils = import 'mixin-utils/utils.libsonnet';

// Add environment and team
local alertGroups = [...];
local withEnvironment = utils.withLabels(
  { environment: 'production', team: 'myteam' },
  alertGroups
);
```

### DIFF
--- a/mixin-utils/utils.libsonnet
+++ b/mixin-utils/utils.libsonnet
@@ -392,6 +392,11 @@ local g = import 'grafana-builder/grafana.libsonnet';
       for group in groups
     ],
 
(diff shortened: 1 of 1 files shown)


tristanburgess (approved):
nice work!

PR #1462, opened 2025-07-02
handle large query ranges on azure deadlocks panels

yduartep:
The query used by the Deadlocks by database panel is likely to miss showing non-zero values the larger the query range is. Since the metric is a count and showing any occurrences of non-zero values is crucial, using `max_over_time` for the query interval is appropriate.

See [this thread for context](https://raintank-corp.slack.com/archives/C036J5B39/p1751047074742109).

### DIFF
--- a/csp-mixin/signals/azuresqldb.libsonnet
+++ b/csp-mixin/signals/azuresqldb.libsonnet
@@ -30,7 +30,7 @@ function(this)
         type: 'gauge',
         sources: {
           azuremonitor: {
-            expr: 'label_replace(azure_microsoft_sql_servers_databases_deadlock_total_count{%(queriesSelector)s}, "database", "$1", "resourceID", ".+/(.*)")',
+            expr: 'label_replace(max_over_time(azure_microsoft_sql_servers_databases_deadlock_total_count{%(queriesSelector)s}[$__interval]), "database", "$1", "resourceID", ".+/(.*)")',
             legendCustomTemplate: '{{database}}',
           },
         },


csp-mixin/signals/azuresqldb.libsonnet
```diff
@@ -30,7 +30,7 @@ function(this)
         type: 'gauge',
         sources: {
           azuremonitor: {
-            expr: 'label_replace(azure_microsoft_sql_servers_databases_deadlock_total_count{%(queriesSelector)s}, "database", "$1", "resourceID", ".+/(.*)")',
+            expr: 'label_replace(max_over_time(azure_microsoft_sql_servers_databases_deadlock_total_count{%(queriesSelector)s}[$__interval]), "database", "$1", "resourceID", ".+/(.*)")',
```
Dasomeone, line 33: What's the usual interval set to for this dashboard? Would have to be fairly large to see any performance increase here, no?

yduartep: @tristanburgess?

tristanburgess: The default query range appears to be 1 hour. As far as I know the interval is by default dynamically computed based on the query range and the max number of data points for the query to return. From my investigations [context](https://raintank-corp.slack.com/archives/C036J5B39/p1751047074742109) there are enough data points generated with the default Prometheus scrape interval of 15s and default max data points to cause the calculated interval to be above 1m with a query range of ~3 days, which triggered some of the non-zero values no longer appearing. The customer involved in the escalation expected to see deadlock events over the last 30 days.

Dasomeone: Ahh that's not a worry at all then, I was thinking more along the lines of our standalone integrations where intervals can sometimes be as low as once a minute :)

PR #1465, opened 2025-07-17
Introduce OTel Collector mixin

tpaschalis:
This PR introduces a new mixin for the OpenTelemetry collector that uses its internal Prometheus metrics.

It was constructed with grafonnet and tried to follow best practices. Let me know what you think!


opentelemetry-collector-mixin/dashboards/collector.libsonnet
```diff
@@ -0,0 +1,137 @@
...
+local variables = import './utils/variables.libsonnet';
+local panels = import './utils/panels.libsonnet';
+local queries = import './utils/queries.libsonnet';
+local cfg = import '../config.libsonnet';
+
+{
+  grafanaDashboards+:: {
+    'collector.json':
+      g.dashboard.new(
+        'OpenTelemetry Collector Health',
```
v-zhuravlev: We normally use Sentence case for Dashboard titles and panel titles

opentelemetry-collector-mixin/dashboards/utils/queries.libsonnet
```diff
@@ -0,0 +1,505 @@
...
+        '$' + variables.datasourceVariable.name,
+        |||
+          sum by (job, cluster, namespace, instance) (
+            {__name__=~"otelcol_process_runtime_total_sys_memory_bytes(_total)?", job=~"$job", cluster=~"$cluster", namespace=~"$namespace", instance=~"$instance"}
+          )
+        |||
+      )
+      + prometheusQuery.withIntervalFactor(2)
+      + prometheusQuery.withLegendFormat(|||
+        RSS - {{cluster}} - {{namespace}} - {{instance}}
```
v-zhuravlev: what if otel collector is deployed outsude k8s and there is no namespace label?

v-zhuravlev: take a look at the concept of three variables we use across all mixins:
1) `filteringSelector`. Static label filter that must added to all queries in panels and alerts. This allows to deploy mixin more than once (with different filteringSlectors) or filter our unncessary metrics coming from other sources (otel-collectors you don't want to see on that dashboard). Could be empty. Example: job="integrations/otel-collector".

2) `groupLabels`. a list of labels that represent a group of instances of SUO. Also labels you would like to keep in alerts aggregations. example: `groupLabels: ['job']` or `groupLabels: [`environment`, 'job', `cluster`].

3) `instanceLabels`. a list of labels that represent a single instance of SUO. 
example: `instanceLabels: ['instance']` or `instanceLabels: [`instance`, 'pod'].

Moving those into config and then reusing those variables instead of static labels in queries, dashboard variables , panel legends gives maximum flexibility no matter where mixin is deployed.

opentelemetry-collector-mixin/dashboards/utils/variables.libsonnet
```diff
@@ -0,0 +1,51 @@
...
+
+  namespaceVariable:
+    variable.query.new('namespace')
+    + variable.query.generalOptions.withLabel('Namespace')
+    + variable.query.withDatasourceFromVariable(self.datasourceVariable)
+    + variable.query.refresh.onTime()
+    + variable.query.withSort(type='alphabetical', asc=false)
+    + variable.query.selectionOptions.withIncludeAll(true, '.*')
+    + variable.query.selectionOptions.withMulti(true)
+    + variable.query.queryTypes.withLabelValues('namespace', metric='{__name__=~"otelcol_process_uptime.*"}'),
```
v-zhuravlev: nit: take a look at commonlib variables func from here: https://github.com/grafana/jsonnet-libs/tree/master/common-lib
you might find it handy to generated funnel-like chained variables:
https://github.com/grafana/jsonnet-libs/blob/master/windows-observ-lib/main.libsonnet#L24-L31

opentelemetry-collector-mixin/dashboards/utils/queries.libsonnet
```diff
@@ -0,0 +1,505 @@
...
+                    instance=~"$instance"
+                }
+            [$__rate_interval])
+        )
+      |||
+    )
+    + prometheusQuery.withIntervalFactor(2)
+    + prometheusQuery.withLegendFormat(|||
+      {{cluster}} - {{namespace}} - {{instance}}
+    |||),
```
v-zhuravlev: nit: You may also use 'signals' declarative format , where functions like rate and histogram_quantile are applied automatically:
https://github.com/grafana/jsonnet-libs/blob/master/common-lib/common/signal/README.md

lots of examples:
https://github.com/grafana/jsonnet-libs/tree/master?tab=readme-ov-file#observability-libraries-signal-extention

opentelemetry-collector-mixin/dashboards/collector.libsonnet
tpaschalis: Thanks, I've amended this in [2bee8e1](https://github.com/grafana/jsonnet-libs/pull/1465/commits/2bee8e1544128f7c0acef5c03032c20e40771082)

opentelemetry-collector-mixin/dashboards/utils/variables.libsonnet
tpaschalis: Yeah, that was much nicer to work with, thanks for the suggestion!

opentelemetry-collector-mixin/dashboards/utils/queries.libsonnet
tpaschalis: Thanks for the tip! 🙌 

I've tried to introduce these, as well as using commonlibs for variables in [744cd6a](https://github.com/grafana/jsonnet-libs/pull/1465/commits/744cd6a19f4d47628aef1ad7d69ec275c1681a66). Would you mind taking another look?

opentelemetry-collector-mixin/dashboards/utils/queries.libsonnet
tpaschalis: I like this approach but opted against it for now. Due to some inconsistencies in OTel instrumentation, the metrics do not always have the same consistent names (say, they might have or not have a `_total`) suffix.

This made my queries look like `__name__=~"otelcol_process_cpu_seconds(_total)?"` in the above snippet, which IIUC is not possible with signals.

I want this mixin to also contain support for Alloy as an OTel collector, so for now I think I'd leave the queries as-is and I'd take another stab at improving this once all metrics have a consistent naming scheme.

tpaschalis:
When I import the dashboard onto Grafana, I get the layout I want and all panels are populated _but_ I also get this warning (tried on two different stacks).

Does anyone have any clue what's that about?
image

opentelemetry-collector-mixin/rules/rules.libsonnet
```diff
@@ -0,0 +1,12 @@
+{
+  prometheusRules+:: {
+    groups+: [
+      {
+        name: 'otelcol-rules',
+        rules: [
+          // TODO: Add prometheus rules.
```
Dasomeone: I'm not sure we'll want to add a blank rulegroup here, might be better to wait until we have something?

opentelemetry-collector-mixin/rules.jsonnet
Dasomeone, line 1: +1 to my comment on the rules file, feels a bit premature to add as-is?

opentelemetry-collector-mixin/dashboards/utils/variables.libsonnet
```diff
@@ -0,0 +1,14 @@
+local cfg = import '../../config.libsonnet';
+local commonlib = import 'common-lib/common/main.libsonnet';
+
+commonlib.variables.new(
+  filteringSelector=cfg._config.filteringSelector,
+  groupLabels=cfg._config.groupLabels,
+  instanceLabels=cfg._config.instanceLabels,
+  varMetric='{__name__=~"otelcol_process_uptime.*"}',
```
Dasomeone: this is where your extra `{` is coming from that you mentioned in your error [here](https://github.com/grafana/jsonnet-libs/pull/1465#issuecomment-3244339701). Specifically this results in the templated variables having queries with multiple blocks in curly braces, like so:
image

opentelemetry-collector-mixin/rules/rules.libsonnet
tpaschalis: Hm, OK I'd mainly added this so I'd learn how to structure the mixin, but I'm OK with removing it as you're saying.

opentelemetry-collector-mixin/dashboards/utils/variables.libsonnet
tpaschalis: Hmm, I see. 

I have some collectors reporting `otelcol_process_uptime` and some reporting `otelcol_process_uptime_seconds_total` so I wanted a catch-all way for both of these. The discrepancy is due to some Prometheus<->OTel conventions and unit conversions IIUC.

Is there a way to achieve that, or should I choose the more slightly 'correct' version of this and have older collectors not represented here?

Dasomeone: The issue here is twofold.

First is that `varMetric` does not represent a query, but rather a singular metric to be used for the instance and job template variables. As such you should not be wrapping it in curly braces, nor defining the  `__name__=~` part manually. 

The second is that this older `varMetric` approach by default assumes your metric does not use  `__name__`, meaning that it will automatically wrap the job, instance, and further chained label selectors in curly braces to append to the metric.

To use this approach you would need to modify or override the logic to not include the job label, and then manually add that to your `varMetric` above, but that may be prone to other issues.

Instead the newer signal approach has support for chained and regex'd variable metrics. The Kafka observ lib uses this to great effect to generate chained variables for selectors as seen here 
image

This updated approach uses the `discoveryMetric` in the signals approach, which lets you define multiple signal sources, which in Kafka's case is `prometheus`, `grafanacloud`, and `bitnami` as commonly available. Notably, this requires `varMetric` to not be set.

In general I'd recommend both the [Kafka-observ-lib](https://github.com/grafana/jsonnet-libs/tree/master/kafka-observ-lib) and [snmp-observ-lib](https://github.com/grafana/jsonnet-libs/tree/master/snmp-observ-lib) as the current best practices for complex signals-based mixins that support multiple metric sources

https://github.com/grafana/jsonnet-libs/blob/e10cd2ae1352eaba653b52f7004e2e4a1354b014/kafka-observ-lib/signals/broker.libsonnet#L10C1-L13C73

I suspect it's why Vitaly strongly suggested you adopt the Signals approach earlier, as it would let you define multiple discovery metrics as needed and you can, I believe, have wildcards on them like in your example here.

Dasomeone: In your case there's three options I see that are all valid:
1. Modify your current query to have `varMetric` represent only the current collector metric
2. Modify `varMetric` and the logic to not inject chained variables like `job` and manually define them
3. Adopt the modern Signals approach, using `discoveryMetrics` which lets you define and use multiple sources in all panels and variables


I'm fine with all of them, so it comes down to what your preference is!

opentelemetry-collector-mixin/rules/rules.libsonnet
Dasomeone: That's totally understandable, I get that. It's a good practice to do to learn it, though in this case we generally tend to avoid blocked out code like this for mixins.

Might be something to change going forward, I could see this being more approachable for new maintainers and contributors :thinking: 

Maybe an issue to track this with the boilerplate included so it's there if needed in the future?

opentelemetry-collector-mixin/dashboards/utils/variables.libsonnet
tpaschalis: Ok, I've checked again and the only collectors that output that secondary metric I was trying to catch with the regex are from a quite older version of the collector, we should be good to go with the simpler `varMetric='otelcol_process_uptime'`.

In the future we should revisit the Prometheus compatibility with other OTel SMEs here and see if we should simplify the rest of the queries as well.

I took a look at the signals approach, but it did seem a little heavy-handed for these simpler dashboards. If you insist, I can migrate it over, up to you.

Dasomeone: Then let's stick with that, as I mentioned I'm good with all three approaches!
If there's little impact due to restricting the `varMetric` to just the one metric let's do that, and we can revisit Signals in the future if it becomes necessary to support multiple metric/OTEL sources.

opentelemetry-collector-mixin/dashboards/utils/queries.libsonnet
Dasomeone: Resolving this on Vitaly's behalf given we resolved it in the other thread

PR #1471, opened 2025-08-14
feat: Update Apache ActiveMQ  mixin to use modern libraries

schmikei:
This PR modernizes the Apache ActiveMQ mixin to utilize an updated common-lib, logs-lib and best match the `kafka-observ-lib`. 

image

image

image

image


apache-activemq-mixin/signals/clusters.libsonnet
```diff
@@ -0,0 +1,131 @@
...
+    },
+
+    enqueueRate: {
+      name: 'Enqueue / $__interval',
+      type: 'raw',
+      description: 'Number of messages that have been sent to destinations in a cluster.',
+      unit: 'none',
+      sources: {
+        prometheus: {
+          expr: 'sum by (activemq_cluster, job) (increase(activemq_queue_enqueue_count{%(queriesSelector)s}[$__interval:])) + sum by (activemq_cluster, job) (increase(activemq_topic_enqueue_count{%(queriesSelector)s, destination!~"ActiveMQ.Advisory.*"}[$__interval:]))',
```
Dasomeone: Just a thought, what do you think of adding a specific selector variable for `job` + `activemq_cluster`? Seems to me they're used together frequently

apache-activemq-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,298 @@
...
+      memoryUsage: {
+        name: 'Average broker memory usage',
+        nameShort: 'Memory usage',
+        type: 'gauge',
+        description: 'The percentage of memory used by both topics and queues across brokers.',
+        unit: 'percentunit',
+        sources: {
+          prometheus: {
+            expr: 'avg(activemq_memory_usage_ratio{%(queriesSelector)s})',
+            legendCustomTemplate: '{{activemq_cluster}} - {{instance}}',
```
Dasomeone: Thinking out loud here, would poke vitaly but he's out on PTO currently.

Thoughts on using `instanceLabels` and `activemqLabels` in the legend here? I know there's a slight risk if there's multiple labels defined, but I think it might make sense for most uses?

apache-activemq-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,298 @@
...
+
+      topicEnqueueCount: {
+        name: 'Topic enqueue count',
+        nameShort: 'Topic enqueue count',
+        type: 'raw',
+        description: 'The number of messages that have been sent to destinations in a cluster.',
+        unit: 'none',
+        sources: {
+          prometheus: {
+            expr: 'sum by (instance, activemq_cluster, job) (increase(activemq_topic_enqueue_count{%(queriesSelector)s})[$__interval:])',
```
Dasomeone: do we use the basic `cluster` anywhere?

I'm commenting as I go here, so I'm just wondering because I've not seen `groupLabels` used yet

apache-activemq-mixin/signals/clusters.libsonnet
```diff
@@ -0,0 +1,131 @@
+function(this) {
+  filteringSelector: this.filteringSelector,
+  groupLabels: this.groupLabels,
+  instanceLabels: this.instanceLabels,
+  enableLokiLogs: this.enableLokiLogs,
```
Dasomeone, line 6: Seems we don't use these at all outside the `queriesSelector` abstraction?

apache-activemq-mixin/signals/queues.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      totalQueueSize: {
+        name: 'Total queue size',
+        nameShort: 'Total queue size',
+        type: 'gauge',
+        description: 'Total number of messages in all queues.',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'sum (activemq_queue_queue_size{%(queriesSelector)s})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: Where at all possible, I'd like to avoid using auto. Are there no good consistent labels we could use here? Even a hardcoded name might be better than the formatting auto provides?

apache-activemq-mixin/signals/queues.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      totalProducers: {
+        name: 'Total producers',
+        nameShort: 'Total producers',
+        type: 'gauge',
+        description: 'Total number of queue producers.',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'sum (activemq_queue_producer_count{%(queriesSelector)s})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: Same thing goes here

apache-activemq-mixin/signals/queues.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      totalConsumers: {
+        name: 'Total consumers',
+        nameShort: 'Total consumers',
+        type: 'gauge',
+        description: 'Total number of queue consumers.',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'sum (activemq_queue_consumer_count{%(queriesSelector)s})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: and here

apache-activemq-mixin/signals/topics.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      topicCount: {
+        name: 'Topic count',
+        nameShort: 'Topic count',
+        type: 'gauge',
+        description: 'Number of active topics (excluding advisory topics).',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'count (activemq_topic_queue_size{%(queriesSelector)s, destination!~"ActiveMQ.Advisory.*"})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: Another auto here that I'd prefer to avoid

apache-activemq-mixin/signals/topics.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      totalProducers: {
+        name: 'Total producers',
+        nameShort: 'Total producers',
+        type: 'gauge',
+        description: 'Total number of topic producers.',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'sum (activemq_topic_producer_count{%(queriesSelector)s, destination!~"ActiveMQ.Advisory.*"})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: +1 auto

apache-activemq-mixin/signals/topics.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      totalConsumers: {
+        name: 'Total consumers',
+        nameShort: 'Total consumers',
+        type: 'gauge',
+        description: 'Total number of topic consumers.',
+        unit: 'short',
+        sources: {
+          prometheus: {
+            expr: 'sum (activemq_topic_consumer_count{%(queriesSelector)s, destination!~"ActiveMQ.Advisory.*"})',
+            legendCustomTemplate: '__auto',
```
Dasomeone: I'm gonna stop commenting on them all, but you know what I'm getting at here :d

apache-activemq-mixin/signals/topics.libsonnet
```diff
@@ -0,0 +1,192 @@
...
+      dequeueRate: {
+        name: 'Top topics by dequeue rate',
+        nameShort: 'Top topics by dequeue rate',
+        type: 'raw',
+        description: 'Rate of messages being dequeued from topics.',
+        unit: 'mps',
+        sources: {
+          prometheus: {
+            expr: 'topk by (instance, activemq_cluster, job) ($k_selector, rate(activemq_topic_dequeue_count{%(queriesSelector)s, destination!~"ActiveMQ.Advisory.*"}[$__rate_interval]))',
+            legendCustomTemplate: '{{activemq_cluster}} - {{instance}} - {{destination}}',
```
Dasomeone: Looking at the screenshots, some of these legends get *very* long, at least with the test values. Might make sense to swap to a table based legend for these? 

Maybe a larger topic around the best-practices for dashboard design like the doc we produced a while back, but specifically for these topic-based targets I wonder if we care *that* much about instance being included?

Dasomeone: image

apache-activemq-mixin/signals/clusters.libsonnet
Dasomeone: I was just checking our best practices doc for something, and I realised we have guidance specifically for `increase` that I'd forgotten about.

Specifically, and I'm quoting here:
> increase() should use
`[$__interval:] -offset $__interval `
for proper increments sum.

Might be something to revisit throughout?

apache-activemq-mixin/README.md
```diff
@@ -54,19 +56,35 @@ Default thresholds can be configured in `config.libsonnet`.
 ```js
 {
   _config+:: {
-    dashboardTags: ['apache-activemq-mixin'],
-    dashboardPeriod: 'now-1h',
```
Dasomeone, line 59: Not used for anything

apache-activemq-mixin/config.libsonnet
```diff
@@ -1,23 +1,39 @@
...
-    activemqAlertsSelector: if self.enableMultiCluster then 'job=~"${job:regex}", cluster=~"${cluster:regex}"' else 'job=~"${job:regex}"',
-    dashboardTags: ['apache-activemq-mixin'],
-    dashboardPeriod: 'now-30m',
-    dashboardTimezone: 'default',
-    dashboardRefresh: '1m',
-    grafanaDashboardIDs: {
-      'apache-activemq-logs-overview.json': 'apache-activemq-logs',
-    },
+  local this = self,
+  enableMultiCluster: false,
```
Dasomeone: Not used for anything

Dasomeone: Thinking about it more, this might be a decent solution to the concurrent use of `activemq_cluster`, if you take a similar approach to the couchbase mixin, with optionally updating the grouplabels?

apache-activemq-mixin/signals/queues.libsonnet
schmikei: Ah this was mostly to match parity with the current mixin... this is rendered as a single statistical sum so I'd imagine its a single latestValue that gets rendered. I suppose with this being a signal it might make sense to remove the `legendCustomTemplate` = `__auto` but the legend is a static string like `queues`? 

image
=>

image


I'll go ahead and make that change but let me know if you want a revert or anything :)

apache-activemq-mixin/signals/clusters.libsonnet
schmikei: Ahh good catch 👍

apache-activemq-mixin/signals/instance.libsonnet
schmikei: `groupLabels` gets mostly propogated via the `commonlib.variables.new()` but I am curious if you are expecting another usage outside of the queriesSelector?

apache-activemq-mixin/signals/instance.libsonnet
schmikei: Ah yep I can explore some options for templating the legends a tad bit better 👍

apache-activemq-mixin/signals/clusters.libsonnet
schmikei: Hmm correct me if I'm wrong but my at least initial understanding was  that we were trying to use the `queriesSelector` for anything related to `groupLabels`? Or is this comment pertaining to the `sum by` expression? Apologies if I'm misunderstanding the suggestion!

apache-activemq-mixin/signals/topics.libsonnet
schmikei: Hmm so something like this?

image

Without Instance:

image

I feel like instance could probably be dropped in the legend for this case as I would imagine that the queues would be unique enough to discover at this level? 🤔  I'll go ahead and make these changes but do want to express that in some instances having the instance information could be useful in some environments.

apache-activemq-mixin/signals/clusters.libsonnet
schmikei: Sure was matching the original queries as those were more heavily tested against, but will go ahead and follow best practices. Good idea 👍

apache-activemq-mixin/signals/clusters.libsonnet
schmikei: Just realized that these change the behavior of the signals API if not present on the signal... i.e. we change datasource specifically if `enableLokiLogs` is set

with enableLokiLogs
```
```
=>

without enableLokiLogs
```
"uid": "${datasource}"
```

I'm not sure if this is an intentional but does break my local setup by doing this. Do you have any insight into the reasoning for this? My local setup does get broken without these fields set on the signals...

apache-activemq-mixin/signals/clusters.libsonnet
Dasomeone: Excellent question, I should've been more clear! My comment was specifically with regards to the sum and group by used throughout, where it appears we might want another selector. I may be wrong, but that's just my gut feeling here to avoid hardcoding too many things (and combining hard-coded with selectors)

apache-activemq-mixin/signals/instance.libsonnet
Dasomeone: Ah I think for this comment I meant in the sum and group by as well, given that we specifically call out activemq_cluster, and users may want to optionally have sum by k8s cluster as well

apache-activemq-mixin/signals/clusters.libsonnet
Dasomeone: This is done for legacy linting reasons, as the linter expects ${datasource}, but this truly is something we're moving away from. Fine to leave it as-is for now then, we'll revisit in another clean-up pass later on though I'm curious what errors you get?

apache-activemq-mixin/signals/queues.libsonnet
Dasomeone: Ah that's a great point, I was fully thinking of it from a point of view of a timeseries panel and not wanting `{instance=xyz, cluster=123}=value` as the legend. For stat panels auto is just fine.

E.g. could you update any use of `__auto` for anything but stat panels to be more descriptive, and leave it as `__auto` for stat panels?

apache-activemq-mixin/signals/topics.libsonnet
Dasomeone: Agreed, it's a very case by case basis and in this case it's fine to drop instance. I'll leave it up to your discretion, but if we keep the instance label, I do prefer having the table legend! :)

apache-activemq-mixin/signals/clusters.libsonnet
Dasomeone: Appreciate it, thank you!

apache-activemq-mixin/README.md
```diff
@@ -60,8 +60,7 @@ Default thresholds can be configured in `config.libsonnet`.
     filteringSelector: 'job=~"integrations/apache-activemq"',
     groupLabels: ['job', 'cluster'],
```
Dasomeone: We should add the `activemq_cluster` here given it's in the config

PR #1473, opened 2025-08-28
group gcp and azure alerts by service

yduartep:
group gcp and azure alerts by service to avoid limit on total alerts by group


anaivanov (approved):
LGTM

PR #1454, opened 2025-06-19
feat: grafana modular lib Couchbase

schmikei:
This introduces more usage of grafana modular lib and logs lib to the couchbase mixin. Functionality should retain the same.

Cluster Overview:
image

Node Overview:
image

Bucket Overview:
image

Logs:
image

PR #1475, opened 2025-08-28
chore: Aerospike Mixin Modernization

schmikei:
This modernizes the aerospike mixin to use signals and commonlib to hopefully provide a more maintainable mixin for the future.

Throughout this we have also added support for some breaking changes around Aerospike 7.0+ versions which removed a couple of prometheus metrics that this mixin relied upon: 

https://aerospike.com/docs/database/reference/metrics#namespace__device_available_pct
https://aerospike.com/docs/database/reference/metrics#namespace__memory_free_pct

We had to change some of the dashboards accordingly. As of right now it just has side-by-side panels i.e.

image

as the workarounds aren't immediately clear to myself how to calculate an equivalent value given the new metrics. This has impacted these objects predominantly:

- Aerospike namespace overview
- Aerospike instance overview
  * two panels now for `Namespace disk usage`
- Aerospike overview
  * two tables now required to display `Top nodes by disk usage`

image
image
image
image
image


aerospike-mixin/mixin.libsonnet
```diff
@@ -1,3 +1,39 @@
...
-(import 'config.libsonnet')
+local mixinlib = import './main.libsonnet';
+local config = (import './config.libsonnet');
+local util = import 'grafana-cloud-integration-utils/util.libsonnet';
+
+local mixin =
+  mixinlib.new()
+  + mixinlib.withConfigMixin(
+    {
+      filteringSelector: 'job=~"integrations/aerospike"',
```
Dasomeone: Same comment as recent PR, can you pull this from the config just to avoid conflicts please?

aerospike-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,135 @@
...
+          )
+        ) + root.applyCommon(
+          vars.multiInstance + [
+            g.dashboard.variable.custom.new(
+              'k',
+              values=['2', '4', '6', '8', '10'],
+            ) + g.dashboard.variable.custom.generalOptions.withCurrent('2')
+            + g.dashboard.variable.custom.generalOptions.withLabel('Top node count')
+            + g.dashboard.variable.custom.selectionOptions.withMulti(false)
+            + g.dashboard.variable.custom.selectionOptions.withIncludeAll(false),
```
Dasomeone, line 36: Neat solution, I like this a lot actually

aerospike-mixin/signals/overview.libsonnet
```diff
@@ -0,0 +1,329 @@
...
+    signals: {
+      nodesClusterSize: {
+        name: 'Nodes cluster size',
+        nameShort: 'Nodes',
+        type: 'raw',
+        description: 'Number of nodes in an Aerospike cluster.',
+        unit: 'none',
+        sources: {
+          prometheus: {
+            expr: 'sum by(job, aerospike_cluster, instance) (aerospike_namespace_ns_cluster_size{%(queriesSelector)s})',
```
Dasomeone: Same comment as on other PRs on whether we want to have a selector for sum (and others?)
I'm generally in favour, but if you have strong concerns I'm good to leave it as-is

schmikei: Ah I think I caught on within my recent commit in #1454, seems like a great idea for extensibility :) I'll go ahead and make similar changes here shortly for this PR 👍

PR #1477, opened 2025-09-08
chore: Apache Hadoop Mixin Modernization 

schmikei:
Modernizing the Apache Hadoop mixin

NameNode Overview
image

Datanode overview
image

NodeManager overview
image
image

ResourceManager overview
image

image

### DIFF
--- a/apache-hadoop-mixin/.lint
+++ b/apache-hadoop-mixin/.lint
@@ -20,10 +20,5 @@ exclusions:
       - panel: "Applications state"
       - panel: "Available virtual cores"
       - panel: "Garbage collection count"
-  alert-name-length:
-    reason: "Alerts are best described with the longer names."
-    entries:
-      - alert: "ApacheHadoopHighResourceManagerVirtualCoreCPUUsage"
-      - alert: "ApacheHadoopHighResourceManagerMemoryUsage"
   template-datasource-rule:
     reason: "Based on new convention we are using variable names prometheus_datasource and loki_datasource where as linter expects 'datasource'"
--- a/apache-hadoop-mixin/README.md
+++ b/apache-hadoop-mixin/README.md
@@ -11,14 +11,14 @@ The Apache Hadoop mixin contains the following dashboards:
 
 and the following alerts:
 
-- ApacheHadoopLowHDFSCapacity
-- ApacheHadoopHDFSMissingBlocks
-- ApacheHadoopHDFSHighVolumeFailures
-- ApacheHadoopHighDeadDataNodes
-- ApacheHadoopHighNodeManagerCPUUsage
-- ApacheHadoopHighNodeManagerMemoryUsage
-- ApacheHadoopHighResourceManagerVirtualCoreCPUUsage
-- ApacheHadoopHighResourceManagerMemoryUsage
+- HadoopLowHDFSCapacity
+- HadoopHDFSMissingBlocks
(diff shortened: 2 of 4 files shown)


apache-hadoop-mixin/.lint
```diff
@@ -20,5 +20,10 @@ exclusions:
       - panel: "Applications state"
       - panel: "Available virtual cores"
       - panel: "Garbage collection count"
+  alert-name-length:
```
Dasomeone: This is not an acceptable bypass, it's a limit enforced by us as hosted grafana/mimir can run into issues whenever a name string is treated as a UID.

It should've been caught in the past too, but we missed these. The tl;dr; is that any name or UID should be less than or equal to 40 characters long to stay safe.

apache-hadoop-mixin/config.libsonnet
```diff
@@ -1,23 +1,39 @@
...
-    alertsCriticalNodeManagerCPUUsage: 80,  // %
-    alertsCriticalNodeManagerMemoryUsage: 80,  // %
-    alertsCriticalResourceManagerVCoreCPUUsage: 80,  // %
-    alertsCriticalResourceManagerMemoryUsage: 80,  // %
 
-    enableLokiLogs: true,
-    enableMultiCluster: false,
-    multiclusterSelector: 'job=~"$job"',
-    hadoopSelector: if self.enableMultiCluster then 'job=~"$job", cluster=~"$cluster"' else 'job=~"$job"',
+  dashboardTags: [self.uid + '-mixin'],
```
Dasomeone, line 8: Ah hmm this is an annoying one, because `-mixin` already is part of the tags, but we try to avoid that nowadays as it's pretty verbose. 

Let's leave it in for now, just in case people use the tag search as a bookmark in Grafana

apache-hadoop-mixin/.lint
schmikei: Sweet yeah I was just going for parity here, but yeah I've got an idea that we could maybe do something like this to stay within 40 characters

```
ApacheHadoopHighRMMemoryUsage
```

Do you mind if we change some of the alert naming conventions then to match so that alerts match that pattern or we could drop the Apache maybe?

```
HadoopHighRMMemoryUsage
```

I think I'm leaning towards the first option but curious if you have a preference!

Dasomeone: Personally I'm fine dropping the `apache` part of the alert name, so long as we keep it in all other text/description for licensing and trademark reasons. The metrics are prefixed `hadoop_` so I see no reason why alerts can't be.

Oh I hit save before I meant to; the last part of my thoughts:

Basically I would prefer us to have descriptive verbose alerts that describe what's failing, e.g. the actual system in full instead of abbreviated, and then drop the relatively unnecessary `apache_` prefix

schmikei: I can get on board with that 👍 I pushed up some renamed alerts with just the `Hadoop` prefix. Also changed the virtual core to vcore for one alert `ApacheHadoopHighResourceManagerVirtualCoreCPUUsage` =>  `HadoopHighResourceManagervCoreCPUUsage`. Feel free to make any suggestions but just giving that a shot for now!

Dasomeone: Perfect, that's super legible, and I agree with the vCore one, that's a common spelling anyways! Appreciate it!

PR #1478, opened 2025-09-10
Update the TensorFlow Mixin to use modern libraries

schmikei:
Tensorflow overview
image
image

Tensorflow logs (new dashboard that gets its own dashboard controls)
image

This initially was to just get the parity of the dashboard, let me know if we want to get any more changes here since its sort of an older mixin :)


tensorflow-mixin/config.libsonnet
```diff
@@ -1,17 +1,34 @@
...
+  dashboardPeriod: 'now-30m',
+  dashboardTimezone: 'default',
+  dashboardRefresh: '1m',
 
-    // for alerts
-    alertsModelRequestErrorRate: '30',  // %
-    alertsBatchQueuingLatency: '5000000',  // µs
+  // logs lib related
+  enableLokiLogs: true,
+  logLabels: if self.enableMultiCluster then ['job', 'instance', 'cluster', 'level'] else ['job', 'instance', 'level'],
```
Dasomeone, line 18: is `model_name` not present in the log labels for us to group by?

tensorflow-mixin/signals/core.libsonnet
```diff
@@ -0,0 +1,66 @@
...
+      },
+      graphBuildTime: {
+        name: 'Average graph build time',
+        nameShort: 'Build time',
+        type: 'raw',
+        description: 'Average time taken to build graphs.',
+        unit: 'µs',
+        sources: {
+          prometheus: {
+            expr: 'increase(:tensorflow:core:graph_build_time_usecs{%(queriesSelector)s}[$__rate_interval])/increase(:tensorflow:core:graph_build_calls{%(queriesSelector)s}[$__rate_interval])',
```
Dasomeone: Could you update the `increase()` queries to use the `[$__interval:] -offset $__interval` time slice instead of `$__rate_interval` from our best practices doc?

https://docs.google.com/document/d/1K9ljGlgQsXkq1prLm3SxwMYNqRCVoowHujUeFs_mxCY/edit?tab=t.0#heading=h.4md3humnajpb

Should be a bit more resilient to low throughput systems and sum correctly

tensorflow-mixin/signals/serving.libsonnet
```diff
@@ -0,0 +1,80 @@
...
+    signals: {
+      requestCount: {
+        name: 'Model request rate',
+        nameShort: 'Request rate',
+        type: 'raw',
+        description: 'Rate of requests over time for the selected model. Grouped by statuses.',
+        unit: 'reqps',
+        sources: {
+          prometheus: {
+            expr: 'rate(:tensorflow:serving:request_count{%(queriesSelector)s, model_name=~"$model_name"}[$__rate_interval])',
```
Dasomeone: `model_name=~"$model_name` here is redundant as its in your groupSelector, which is added to the `queriesSelector` upon dashboard generation.

The output for this expression when rendered is:
```
rate(:tensorflow:serving:request_count{job=~\"integrations/tensorflow\",job=~\"$job\",cluster=~\"$cluster\",model_name=~\"$model_name\",instance=~\"$instance\", model_name=~\"$model_name\"}[$__rate_interval])
```

I would suggest potentially dropping the model_name from the group selector if we don't want it by default everywhere, or if we do then drop it here.

Note job is also added twice, but that's done on purpose as we code in the default value

tensorflow-mixin/signals/serving.libsonnet
```diff
@@ -0,0 +1,80 @@
...
+      },
+      requestLatency: {
+        name: 'Model predict request latency',
+        nameShort: 'Request latency',
+        type: 'raw',
+        description: 'Average request latency of predict requests for the selected model.',
+        unit: 'µs',
+        sources: {
+          prometheus: {
+            expr: 'increase(:tensorflow:serving:request_latency_sum{%(queriesSelector)s, model_name=~"$model_name"}[$__rate_interval])/increase(:tensorflow:serving:request_latency_count{%(queriesSelector)s, model_name=~"$model_name"}[$__rate_interval])',
```
Dasomeone: +1 to the `increase()` queries needing to be rewritten (all of them in this file too)

Dasomeone: Same comment re: `model_name` as well here and throughout as we're not selecting something specific here, there shouldn't be a need to duplicate it

tensorflow-mixin/panels.libsonnet
```diff
@@ -0,0 +1,121 @@
...
+      core_graphBuildTimePanel:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Average graph build time',
+          targets=[
+            signals.core.graphBuildTime.asTarget(),
+          ],
+          description='Average time taken to build graphs.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('µs')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withFillOpacity(0)
```
Dasomeone: we should always use `withFillOpacity(10)` or greater to comply with our best practices guide for time series panels https://docs.google.com/document/d/1K9ljGlgQsXkq1prLm3SxwMYNqRCVoowHujUeFs_mxCY/edit?tab=t.0#heading=h.erqj2ugjulx5

Also applies throughout

tensorflow-mixin/signals/core.libsonnet
schmikei: Ah yep thanks for the catch 👍 Apologies for the miss!

tensorflow-mixin/config.libsonnet
schmikei: There is no model_name in the logs as those are the ModelServer logs and not related to a specific model. Makes me think about the groupLabels and whether or not to include `model_name` as a group label as not all metrics include a `model_name`

image


I'll go ahead and make that change and update the filtering approach :)

PR #1480, opened 2025-09-15
Update Varnish Mixin to use modern libraries

schmikei:
Updates the Varnish Mxin to use more modern libaries and signals API

Overview
image
image

Logs Overview
image

Have some questions around why we're using irate and wanted to discuss before merging as of submission of this PR.


varnish-mixin/panels.libsonnet
```diff
@@ -0,0 +1,267 @@
...
+      // Backend connections timeseries
+      backendConnectionsPanel:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Backend connections',
+          targets=[
+            signals.backend.backendConnectionsAccepted.asTarget(),
+            signals.backend.backendConnectionsRecycled.asTarget(),
+            signals.backend.backendConnectionsReused.asTarget(),
+            signals.backend.backendConnectionsBusy.asTarget(),
+            signals.backend.backendConnectionsUnhealthy.asTarget(),
```
schmikei: All these signals previously used the `irate()` function and I'm not quite sure how useful it is to know rate per second of the connection... I've currently translated as originally implemented but it seemed like we should maybe be displaying using `increase`. Happy to hear other thoughts, but just figured that knowing a whole number connection value here would be more user friendly/intuitive.

varnish-mixin/panels.libsonnet
```diff
@@ -0,0 +1,267 @@
...
+        + g.panel.timeSeries.fieldConfig.defaults.custom.stacking.withMode('normal'),
+
+      // Sessions timeseries
+      sessionsPanel:
+        commonlib.panels.generic.timeSeries.base.new(
+          'Sessions',
+          targets=[
+            signals.sessions.sessionsConnected.asTarget(),
+            signals.sessions.sessionsQueued.asTarget(),
+            signals.sessions.sessionsDropped.asTarget(),
```
schmikei: Same thing on usage of `irate` here which could be useful but feel as though the `increase` value may be more useful?

varnish-mixin/panels.libsonnet
Dasomeone: Agreed, and thank you for calling this out! I believe the later best-practices around the correct offset for `increase` actually stems from a discussion we had around a bug with the Varnish integration (or feedback related to it at least?) so I'm good with converting these to using increase.

varnish-mixin/panels.libsonnet
Dasomeone: +1, I agree with swapping them.

varnish-mixin/signals/backend.libsonnet
```diff
@@ -0,0 +1,96 @@
...
+      backendConnectionsRecycled: {
+        name: 'Backend connections recycled',
+        nameShort: 'Recycled',
+        type: 'counter',
+        description: 'Rate of recycled backend connections.',
+        unit: 'conn/s',
+        sources: {
+          prometheus: {
+            expr: 'varnish_main_backend_recycle{%(queriesSelector)s}',
+            rangeFunction: 'irate',
```
Dasomeone: Like you rightly pointed out, we can convert these to `increase()` using the proper offset and it should be a little more reliable than `irate()`

varnish-mixin/rows.libsonnet
```diff
@@ -0,0 +1,32 @@
...
+          // line of stat panels so width is generally small within the grid
+          this.grafana.panels.cacheHitRatePanel { gridPos+: { w: 3 } },
+          this.grafana.panels.frontendRequestsPanel { gridPos+: { w: 3 } },
+          this.grafana.panels.backendRequestsPanel { gridPos+: { w: 3 } },
+          this.grafana.panels.sessionsRatePanel { gridPos+: { w: 3 } },
+          this.grafana.panels.cacheHitsPanel { gridPos+: { w: 3 } },
+          this.grafana.panels.cacheHitPassPanel { gridPos+: { w: 3 } },
+          this.grafana.panels.sessionQueueLengthPanel { gridPos+: { w: 3 } },
+          this.grafana.panels.poolsPanel { gridPos+: { w: 3 } },
+
```
Dasomeone: Please split this into a separate row as we deliberately wanted the stat panels to be 4 units tall, while the current approach ends up being 8 units tall

varnish-mixin/panels.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+      cacheHitPassPanel:
+        commonlib.panels.generic.stat.base.new(
+          'Cache hit pass',
+          targets=[
+            signals.cache.cacheHitPass.asTarget(),
+          ],
+          description='Rate of cache hits for pass objects (fulfilled requests that are not cached).'
+        )
+        + g.panel.stat.standardOptions.withUnit('/ sec')
+        + g.panel.stat.panelOptions.withTransparent(true),
```
Dasomeone: Please add this for consistency with the other stat panels 
```suggestion
        + g.panel.stat.options.withGraphMode('none')
 ```

File: varnish-mixin/panels.libsonnet [hunk 6]
```diff
@@ -0,0 +1,275 @@
...
+          + g.panel.gauge.standardOptions.threshold.step.withValue(80),
+        ])
+        + g.panel.gauge.options.withOrientation('auto')
+        + g.panel.gauge.options.withShowThresholdLabels(false)
+        + g.panel.gauge.options.withShowThresholdMarkers(true)
+        + g.panel.gauge.panelOptions.withDescription('Rate of cache hits to misses.')
+        + g.panel.gauge.panelOptions.withTransparent(true),
+
+      // Frontend requests stat
+      frontendRequestsPanel:
```
Dasomeone, line 39: for all the stat panels can you please modify them and the queries to be sums/stats across all instances such that the stat panels do not split into one each? Gets pretty cluttered pretty quickly with multiple instances

Image

This was a complaint we saw with the old Varnish mixin as well so easy to miss as it wasn't something we fixed

varnish-mixin/panels.libsonnet
```diff
@@ -0,0 +1,275 @@
...
+          targets=[
+            signals.threads.threadsFailed.asTarget() { interval: '2m' },
+            signals.threads.threadsCreated.asTarget() { interval: '2m' },
+            signals.threads.threadsLimited.asTarget() { interval: '2m' },
+            signals.threads.threadsTotal.asTarget(),
+          ],
+          description='Number of failed, created, limited, and current threads.'
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('none')
+        + g.panel.timeSeries.fieldConfig.defaults.custom.withFillOpacity(0)
```
Dasomeone, line 272: Reminder here that we want at least `withFillOpacity(10)`, though `20` is perfectly fine as well

PR #1482, opened 2025-09-17
Update variables in Traefik dashboard to follow template format

annanay25:


annanay25:
done! should be all set this time.

PR #1484, opened 2025-09-25
Modernize the InfluxDB mixin to use modern libraries (WIP)

schmikei:
WIP; think its close just going to do some testing


influxdb-mixin/dashboards_out/influxdb-cluster-overview.json
```diff
@@ -134,85 +117,87 @@
...
+                  "legendFormat": "Buckets",
+                  "refId": "Buckets"
                },
                {
                   "datasource": {
+                     "type": "prometheus",
                      "uid": "${prometheus_datasource}"
                   },
-                  "expr": "influxdb_users_total{job=~\"$job\", influxdb_cluster=~\"$influxdb_cluster\"}",
+                  "expr": "influxdb_users_total{job=\"integrations/influxdb\",job=~\"$job\",influxdb_cluster=~\"$influxdb_cluster\",instance=~\"$instance\"}",
```
Dasomeone: Just a note here, we'll probably want to avoid including `instance` as a filtering selector for the cluster overview panels if they weren't already present?
Specifcally, if memory serves, we deliberately excluded them for overview-type panels where we wanted the picture for the whole cluster

PR #1616, opened 2026-03-31
Fix memcached dashboard hit rate query 

fionaliao:
The hit rate query was missing the `instance` filter in denominator.


karlskewes (approved):
Good find. Can't see any others missing `instance`.

PR #1617, opened 2026-04-01
Bump jsonnet to 0.22

v-zhuravlev:
The new version of jsonnet brings some nice features such as negative offset for arrays, removing the dependency on xtd package in some mixins: https://github.com/jsonnet-libs/xtd
jb is also updated to the latest version that fixes transient dependencies behaviour 
https://github.com/jsonnet-bundler/jsonnet-bundler/issues/160


Dasomeone (dismissed):
Oh this is huge actually, thanks for flagging this @v-zhuravlev ! :) 
Edit: looks like the golang toolchain needs a bump as well, would you please do that as well? :)

Dasomeone (approved):

PR #1621, opened 2026-04-15
update link for oracleDB mixin to point to official exporter

schmikei:
This mixin is primarily built around [alloy's component for oracledb ](https://grafana.com/docs/alloy/latest/reference/components/prometheus/prometheus.exporter.oracledb/) which uses https://github.com/oracle/oracle-db-appdev-monitoring as the exporter now since `v1.9.0` of Grafana Alloy.

Because of this we should update the README of the mixin to update the reference to be the official exporter.


Dasomeone (approved):
LGTM, thanks Keith!

PR #1625, opened 2026-04-17
fix: mssql link casing inconsistency

schmikei:
Fixes a style guide inconsistency where the links should be sentence cased in this context


Dasomeone (approved):
Great catch, thanks Keith!

PR #1626, opened 2026-04-21
chore(commonlib): panel requests.stat.errors use default palette color 

v-zhuravlev:
panel `requests.stat.errors` use default palette color (defined in tokens) if there are no errors, and red color if errors > 0.


Dasomeone (approved):
Nice little improvement here, thanks again Vitaly! :D

PR #1629, opened 2026-04-23
chore: Modernize mongodb-mixin

mariaalons:
Update the MongoDB mixin to match the modernized mixins:                                         
                                                                                                    
  - Replace hand-written dashboard JSON with generated dashboards                                   
  - Add MongoDB Overview dashboard                                                                  
  - Add conditional MongoDB Logs dashboard (enableLokiLogs flag)                                                                                                              
  - Update jsonnetfile to grafonnet v11.4.0
  - Use signals
  
Screenshot 2026-06-26 at 15 55 27
Screenshot 2026-06-26 at 16 09 19
Screenshot 2026-06-26 at 16 59 27
Screenshot 2026-06-26 at 16 59 43
Screenshot 2026-06-26 at 17 00 16


mongodb-mixin/.lint
```diff
@@ -1,15 +1,18 @@
...
+    reason: "Based on new convention we are using variable names prometheus_datasource and loki_datasource where as linter expects 'datasource'"
   template-instance-rule:
-    reason: "This integration is designed to work with 2 specific labels, service_name and mongodb_cluster, instead of instance."
-  template-job-rule:
-    reason: "mixtool upgrade made this rule stricter. TODO: Fix errors and remove the warning exclusion"
-  template-on-time-change-reload-rule:
-    reason: "mixtool upgrade made this rule stricter. TODO: Fix errors and remove the warning exclusion"
+    reason: "Based on new convention we are using variable names prometheus_datasource and loki_datasource where as linter expects 'datasource'"
+  panel-units-rule:
+    reason: "Custom units are used for better user experience in these panels"
```
Dasomeone: Unless this is all over the place, we should specify the specific panels in question (like the text implies)

mongodb-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,166 @@
...
+        ),
+
+      'mongodb-instance.json':
+        g.dashboard.new(this.config.dashboardNamePrefix + ' instance')
+        + g.dashboard.withPanels(
+          g.util.panel.resolveCollapsedFlagOnRows(
+            g.util.grid.wrapPanels([
+              this.grafana.rows.instanceOverview,
+              this.grafana.rows.instanceServiceSummary,
+            ])
```
Dasomeone, line 44: I think it'd make sense to slightly rearrange the layout of this dashboard, specifically the summary as we do have 12 panels in 2 column layout. Would be good to have *some* breaks, or split it into two separate groups

mongodb-mixin/rows.libsonnet
```diff
@@ -0,0 +1,166 @@
...
+        this.grafana.panels.instanceLatencyDetail { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceQueuedOps { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceCursors { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceScannedAndMoved { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceAsserts { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceGetLastErrorWriteOps { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceQueryEfficiency { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceGetLastErrorWriteTime { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instancePageFaults { gridPos+: { w: 12, h: 8 } },
+      ]),
```
Dasomeone, line 52: This is the section I think could do with a rearrangement. Would suggest 2-3-2-3-2 or similar, same as we suggested to Keith. If not possible due to right-aligned legends then an option would be alternating width to overlap them, or inserting multiple rows to create a gap between

mongodb-mixin/signals/cluster.libsonnet
```diff
@@ -0,0 +1,421 @@
...
+      collectionsInShards: {
+        name: 'Collections in shards',
+        description: 'Number of collections per shard.',
+        type: 'gauge',
+        unit: 'short',
+        aggFunction: 'max',
+        sources: {
+          percona_mongodb: {
+            expr: 'mongodb_mongos_db_collections_total{%(queriesSelector)s, db!~"admin|config"}',
+            legendCustomTemplate: '{{db}} | {{shard}} | Collections',
```
Dasomeone: What is the benefit of appending `| Collections` to all legends here?

Dasomeone (commented):
Seeing myself blind on these queries, going to have to get back to this PR on Monday, but for now here's my comments so far :)

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,356 @@
+local g = import '../g.libsonnet';
+local commonlib = import 'common-lib/common/main.libsonnet';
+
+{
+  new(signals, config): {
+    local tsPanel = g.panel.timeSeries,
+    local withFillOpacity = tsPanel.fieldConfig.defaults.custom.withFillOpacity(10),
```
Dasomeone: This shouldn't be necessary if you create panels from the common-lib base as they contain opinionated defaults. (and if they don't, we should update the common-lib so they do :D)

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,356 @@
...
+    local withTableLegend =
+      tsPanel.options.legend.withPlacement('right')
+      + tsPanel.options.legend.withDisplayMode('table')
+      + tsPanel.options.legend.withCalcs(['lastNotNull', 'min', 'mean', 'max']),
```
Dasomeone, line 10: Small utility builder options like this might be a good idea for future common-lib improvement as well. 

Perfectly fine as is here, but just something we might want to extract for a future improvement pass and think about other things that we want

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,356 @@
...
+    local withStacking = tsPanel.fieldConfig.defaults.custom.stacking.withMode('normal'),
+
+    // Sharding overview stats
+    clusterShardsTotal:
+      commonlib.panels.generic.stat.base.new(
+        '# of shards',
+        targets=[signals.cluster.shardsTotal.asTarget()],
+        description='Total number of shards.'
+      )
+      + g.panel.stat.standardOptions.withUnit('short')
```
Dasomeone, line 36: This one we definitely want to add to the generic bases for all our panels, such an easy one to forget otherwise

mongodb-mixin/rows.libsonnet
```diff
@@ -0,0 +1,166 @@
...
+        this.grafana.panels.instanceLatencyDetail { gridPos+: { w: 8, h: 8 } },
+        this.grafana.panels.instanceQueuedOps { gridPos+: { w: 8, h: 8 } },
+        this.grafana.panels.instanceCursors { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceScannedAndMoved { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instanceAsserts { gridPos+: { w: 8, h: 8 } },
+        this.grafana.panels.instanceGetLastErrorWriteOps { gridPos+: { w: 8, h: 8 } },
+        this.grafana.panels.instanceQueryEfficiency { gridPos+: { w: 8, h: 8 } },
+        this.grafana.panels.instanceGetLastErrorWriteTime { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.instancePageFaults { gridPos+: { w: 12, h: 8 } },
+      ]),
```
Dasomeone: Can you put both `getLastError` on the same row, or would that squish the `Page Faults` panel too much? (Can't verify as sample-app doesn't produce metrics for this)

mongodb-mixin/README.md
```diff
@@ -1,22 +1,113 @@
...
+
+The MongoDB overview dashboard provides fleet-wide health and performance: total instances, instances up/down, total connections, total operations, max replication lag, and per-instance state, connections, operations, and replication lag.
+
+## MongoDB instance
+
+The MongoDB instance dashboard provides per-instance details: uptime, QPS, replica set state, latency, command/document operations, connections, queued operations, cursors, scanned/moved objects, asserts, page faults, and query efficiency.
+
+## MongoDB replica set
+
+The MongoDB replica set dashboard provides per-replica-set details: members, last election, average lag, version distribution, member states, replication lag, operations, elections, heartbeat and member ping times, and oplog details (buffered ops, getmore time, recovery window, processing time, operations).
```
Dasomeone: Both of these two chunks might benefit from being broken out into lists as there's so many to mention? thoughts? can group where it makes sense, such as your parenthesis

mongodb-mixin/README.md
```diff
@@ -1,22 +1,113 @@
...
+
+## MongoDB cluster
+
+The MongoDB cluster dashboard provides sharded-cluster details: number of shards, sharded/unsharded databases, draining shards, sharded collections, total chunks, balancer status, collections and chunks per shard, indexes per shard, connections, operations, and cursors.
```
Dasomeone: This is the other one that could do with being bulletpointed

mongodb-mixin/README.md
```diff
@@ -1,22 +1,113 @@
...
+```
+
+The files in `dashboards_out` need to be imported
+into your Grafana server. The exact details will be depending on your environment.
+
+`prometheus_alerts.yaml` needs to be imported into Prometheus.
+
+## Generate dashboards and alerts
+
+Edit `config.libsonnet` if required and then build JSON dashboard files for Grafana:
```
Dasomeone: Minor nit, let's call it `and then generate the dashboard files`, no need to say build or JSON

mongodb-mixin/alerts.libsonnet
```diff
@@ -76,34 +76,42 @@ local xtd = import 'github.com/jsonnet-libs/xtd/main.libsonnet';
...
+            local agg = std.join(',', this.config.groupLabels + this.config.instanceLabels),
             alert: 'MongodbTooManyConnections',
             annotations: {
-              description: 'Too many connections to MongoDB instance {{ $labels.%(instanceLabel)s }} (> 80%%).' % config,
               summary: 'MongoDB has too many connections.',
+              description: 'Too many connections to MongoDB instance {{ $labels.%(instanceLabel)s }} (> %(alertsWarningConnectionUtilization)s%%%%).' % this.config {
+                instanceLabel: this.config.instanceLabels[std.length(this.config.instanceLabels) - 1],
+              },
             },
-            expr: 'avg by (%(agg)s) (rate(mongodb_connections{state="current",%(filteringSelector)s}[1m])) / avg by (%(agg)s) (sum (mongodb_connections) by (%(agg)s)) * 100 > 80' % config,
```
Dasomeone: As this alert here has changed quite a bit, I just want to make sure the logic is preserved properly.

Was it broken before? What do we gain by the rework?

mongodb-mixin/signals/cluster.libsonnet
```diff
@@ -0,0 +1,421 @@
...
+        },
+      },
+      cursorsTotal: {
+        name: 'MongoDB cursors',
+        description: 'Total number of open cursors across the cluster.',
+        type: 'raw',
+        unit: 'short',
+        sources: {
+          percona_mongodb: {
+            expr: 'sum by () (max_over_time(mongodb_mongod_metrics_cursor_open{%(queriesSelector)s, state="total"}[$__rate_interval]) or max_over_time(mongodb_mongos_metrics_cursor_open{%(queriesSelector)s, state="total"}[$__rate_interval]) or max_over_time(mongodb_mongod_cursors{%(queriesSelector)s, state="total_open"}[$__rate_interval]) or max_over_time(mongodb_mongos_cursors{%(queriesSelector)s, state="total_open"}[$__rate_interval]))',
```
Dasomeone: Can probably safely drop the `by ()` at the start here, no?

Additionally, some of these stacked expressions might benefits from being broken over multiple lines (see the [snmp-observ-lib](https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/signals/memory.libsonnet#L53 ) for a direct example)

This really applies on all of the bigger expressions across all the signals

mongodb-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,267 @@
...
+        unit: 's',
+        sources: {
+          percona_mongodb: {
+            expr: 'mongodb_instance_uptime_seconds{%(queriesSelector)s}',
+          },
+        },
+      },
+      replicaSetState: {
+        name: 'Replica set state',
+        description: 'An integer between 0 and 10 that represents the replica state of the current member.',
```
Dasomeone: This would benefit to a link to the source/definition of what the 0-10 states mean or represent

mongodb-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,267 @@
...
+          percona_mongodb: {
+            expr: 'avg by (%(agg)s, type) (irate(mongodb_mongod_op_counters_repl_total{%(queriesSelector)s, type!~"(command|query|getmore)"}[$__rate_interval]) or irate(mongodb_mongos_op_counters_repl_total{%(queriesSelector)s, type!~"(command|query|getmore)"}[$__rate_interval]))',
+            legendCustomTemplate: '{{type}}',
+          },
+        },
+      },
+      ttlDeletedDocuments: {
+        name: 'TTL deleted documents',
+        description: 'Rate of documents deleted by TTL indexes.',
+        type: 'raw',
```
Dasomeone, line 128: Not sure on the naming here, think it could be a bit clearer.
Description is fine, but panel naming I mean

mongodb-mixin/alerts.libsonnet
```diff
@@ -1,15 +1,5 @@
...
-    local config =
-      $._config
-      {
-        agg: std.join(',', $._config.groupLabels + $._config.instanceLabels),
-        aggCluster: std.join(',', $._config.groupLabels),
-        instanceLabel: xtd.array.slice($._config.instanceLabels, -1)[0],
-        groupLabel: xtd.array.slice($._config.groupLabels, -1)[0],
-
-      },
+  new(this): {
```
Dasomeone, line 2: One more thing on the alerts, we could move to using signals here as well, similar to how Vitaly did it in the [snmp-observ-lib](https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/signals/alerts.libsonnet)

Dasomeone (commented):
Thanks for the updates Maria! We'll definitely want to update the sample-app in lockstep with this as currently it doesn't cover appropriately to test/validate

mongodb-mixin/panels/cluster.libsonnet
mariaalons: This is a really good one, we can add it in the future

mongodb-mixin/README.md
mariaalons: I grouped, or added it as pointed list so it's easier to read

mongodb-mixin/alerts.libsonnet
mariaalons: It was broken before or at least it would probably never fire. The old one used rate() on mongodb_connections, which the percona exporter exposes as Untyped so the numerator was basically 0 while the denominator summed every state (including totalCreated), which dominated. Ratio stayed near zero, alert wouldn't fire.

mongodb-mixin/panels/cluster.libsonnet
Dasomeone: Can you create an issue to track this one please ? :)

mariaalons: yes! here https://github.com/grafana/jsonnet-libs/issues/1637

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,184 @@
...
+      + tsPanel.options.legend.withDisplayMode('table')
+      + tsPanel.options.legend.withCalcs(['lastNotNull', 'min', 'mean', 'max']),
+
+    // Replica set state value mappings
+    local replicaSetStateMappings =
+      g.panel.stat.standardOptions.mapping.ValueMap.withType()
+      + g.panel.stat.standardOptions.mapping.ValueMap.withOptions({
+        '0': { index: 0, text: 'STARTUP' },
+        '1': { index: 1, text: 'PRIMARY' },
+        '2': { index: 2, text: 'SECONDARY' },
```
Dasomeone: These two may need/benefit from a slightly more descriptive healthy state, given they already have y-axis that label them as primary/secondary.

I'm thinking perhaps `HEALTHY (PRIMARY)` or `PRIMARY - Healthy`, or similar. 

Image

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,184 @@
...
+      commonlib.panels.generic.stat.base.new(
+        'Average replica set lag',
+        targets=[signals.replicaset.avgReplicationLag.asTarget()],
+        description='Average replication lag across secondary members.'
+      )
+      + g.panel.stat.standardOptions.withUnit('s')
+      + g.panel.stat.options.reduceOptions.withCalcs(['lastNotNull'])
+      + g.panel.stat.options.withGraphMode('none'),
+
+    replicasetVersions:
```
Dasomeone, line 58: Slight nit on column naming here, might benefit from some cleanup so it's not just the label names?

Not 100% sold on this, so feel free to push back :D 

Image

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,184 @@
...
+
+    replicasetHeartbeatTime:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Max heartbeat time',
+        targets=[signals.replicaset.heartbeatTime.asTarget()],
+        description='Time since last heartbeat from each member.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('s'),
+
+    replicasetMemberPing:
```
Dasomeone, line 126: Duplicate primary/secondary in the legends on this panel 

Image

mongodb-mixin/panels/overview.libsonnet
```diff
@@ -0,0 +1,147 @@
...
+    // Per-instance panels
+    overviewInstanceStates:
+      commonlib.panels.generic.statusHistory.base.new(
+        'Instance states',
+        targets=[signals.overview.replicaSetStateByInstance.asTarget()],
+        description='Replica set state per instance over time.'
+      )
+      + g.panel.statusHistory.standardOptions.withMappings([replicaSetStateMappings])
+      + g.panel.statusHistory.options.withShowValue('auto'),
+
```
Dasomeone, line 119: I think you'll need to adjust the visualisation for this panel slightly, seems a little unwieldy.

Perhaps cleaning up the number display and y axis might help? 
Image

mongodb-mixin/panels/overview.libsonnet
```diff
@@ -0,0 +1,147 @@
...
+        '1': { index: 1, text: 'PRIMARY' },
+        '2': { index: 2, text: 'SECONDARY' },
+        '3': { index: 3, text: 'RECOVERING' },
+        '5': { index: 4, text: 'STARTUP2' },
+        '6': { index: 5, text: 'UNKNOWN' },
+        '7': { index: 6, text: 'ARBITER' },
+        '8': { index: 7, text: 'DOWN' },
+        '9': { index: 8, text: 'ROLLBACK' },
+        '10': { index: 9, text: 'REMOVED' },
+      }),
```
Dasomeone, line 25: Same comment here as on replicaset, probably need a healthy status for primary/secondary

mongodb-mixin/dashboards.libsonnet
```diff
@@ -0,0 +1,166 @@
...
+          tags,
+          links { mongodbCluster+:: {} },
+          annotations,
+          timezone,
+          refresh,
+          period,
+        ),
+
+    } + if this.config.enableLokiLogs then {
+      'mongodb-logs.json':
```
Dasomeone, line 125: Similar to the work done by Keith in recent PRs, you'll need to set a custom allValues for the logs of `.+` either here or in the final integration, as otherwise we get the usual error

Image

mongodb-mixin/config.libsonnet
```diff
@@ -1,8 +1,36 @@
...
+  dashboardTimezone: 'default',
+  dashboardRefresh: '1m',
+  metricsSource: ['percona_mongodb'],
+
+  // Logging configuration
+  enableLokiLogs: true,
+  extraLogLabels: ['level'],
+  logsVolumeGroupBy: 'level',
+  showLogsVolume: true,
+  customAllValue: '.+',
```
Dasomeone, line 19: Oh hm i just commented on tihs in the dashboards.libsonnet. This doesn't seem to get pulled in correctly. 

Might be the common-lib / logs-lib versions need an update?

mongodb-mixin/config.libsonnet
```diff
@@ -1,8 +1,36 @@
...
-    groupLabels: ['job', 'mongodb_cluster'],
-    instanceLabels: ['service_name'],
+  local this = self,
+  filteringSelector: '',
+  groupLabels: ['job', 'mongodb_cluster'],
+  instanceLabels: ['service_name'],
+  uid: 'mongodb',
+  dashboardTags: [self.uid + '-mixin'],
+  dashboardNamePrefix: 'MongoDB',
+  dashboardPeriod: 'now-1h',
```
Dasomeone: Our default is 30 minutes, but if there's a good reason to set it to 1h that's fine too

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,198 @@
...
+    instanceReplicaSetState:
+      commonlib.panels.generic.stat.base.new(
+        'Replica set',
+        targets=[signals.instance.replicaSetState.asTarget()],
+        description='An integer between 0 and 10 that represents the replica state of the current member. See https://www.mongodb.com/docs/manual/reference/replica-states/ for the meaning of each value.'
+      )
+      + g.panel.stat.standardOptions.withMappings([replicaSetStateMappings])
+      + g.panel.stat.options.reduceOptions.withCalcs(['lastNotNull'])
+      + g.panel.stat.options.withGraphMode('none'),
+
```
Dasomeone, line 97: Description and legends for this needs to be updated. Currently doesn't match the mapping used, and the legend is leaking through in the background

Image

mongodb-mixin/dashboards.libsonnet
Dasomeone, line 125: Additionally, with the[ updated sample-app](https://github.com/grafana/integration-sample-apps/pull/188/), logs are shipped to loki but the current dashboard does not pick them up.

mongodb-mixin/config.libsonnet
Dasomeone, line 19: Hmm this is causing issues with the labels for `mongodb_cluster` and `service_name`, along with others used throughout the updated mixin, wherein the metric is present but the labels are not (in the updated sample app)

As a result, the default queries don't match anything. Example here: 

Image

What you may need to do is a deliberate hardcoding of `.*` of labels that must be optional (or vice versa, default to `.*` and set `.+` with hardcoded on required)

mongodb-mixin/panels/cluster.libsonnet
Dasomeone, line 1: This entire dashboard is broken due to the `.+` allvalue currently

Dasomeone (changes requested):
Great progress, but we'll need to iron out the customAllValue as it's causing some issues. Basically we'll probably want to either have hardcoded values, or a very specific customAllValues just for the logs lib as it currently affects the main mixin, but not the logs lib for whatever reason

mariaalons:
@Dasomeone Thank you for the thorough review 🙏  I applied the requested changes 🙇

mongodb-mixin/panels/replicaset.libsonnet
Dasomeone: Much nicer now, but I just realised I'd forgotten to comment that there's a duplicate "replica set states" on each set. Might want to drop it since it's already in the title

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+        targets=[signals.replicaset.lastElection.asTarget()],
+        description='Time since the last replica set election.'
+      )
+      + g.panel.stat.standardOptions.withUnit('s')
+      + g.panel.stat.options.reduceOptions.withCalcs(['lastNotNull'])
+      + g.panel.stat.options.withGraphMode('none'),
+
+    replicasetAvgLag:
+      commonlib.panels.generic.stat.base.new(
+        'Average replica set lag',
```
Dasomeone, line 50: another legend issue here 

Image

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+        '6': { index: 5, text: 'UNKNOWN' },
+        '7': { index: 6, text: 'ARBITER' },
+        '8': { index: 7, text: 'DOWN' },
+        '9': { index: 8, text: 'ROLLBACK' },
+        '10': { index: 9, text: 'REMOVED' },
+      }),
+
+    replicasetMembers:
+      commonlib.panels.generic.stat.base.new(
+        'Replica set members',
```
Dasomeone, line 30: Legend issue
Image

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+      commonlib.panels.generic.timeSeries.base.new(
+        'Oplog getmore time',
+        targets=[signals.replicaset.oplogGetmoreTime.asTarget()],
+        description='Rate of time spent on oplog getmore operations.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('ms'),
+
+    replicasetOplogRecoveryWindow:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Oplog recovery window',
```
Dasomeone, line 153: Missing service name info or hostname in legend. 

No data for the rest of the oplog panels so can't verify, but might extend to all of them
Image

mongodb-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,319 @@
...
+    instanceLabels: this.instanceLabels,
+    enableLokiLogs: this.enableLokiLogs,
+
+    aggLevel: 'instance',
+    aggFunction: 'avg',
+    discoveryMetric: {
+      percona_mongodb: 'mongodb_up',
+    },
+    signals: {
+      uptime: {
```
Dasomeone, line 15: Duplicate "uptime" mention in all legends 

Image

mongodb-mixin/signals/instance.libsonnet
```diff
@@ -0,0 +1,319 @@
...
+        description: 'The uptime of the MongoDB instance.',
+        type: 'gauge',
+        unit: 's',
+        sources: {
+          percona_mongodb: {
+            expr: 'mongodb_instance_uptime_seconds{%(queriesSelector)s}',
+          },
+        },
+      },
+      replicaSetState: {
```
Dasomeone, line 27: Interesting value mapping issue here, you may need to break out the shards individually 

Image

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+      + g.panel.stat.standardOptions.withMappings([replicaSetStateMappings])
+      + g.panel.stat.standardOptions.withDisplayName('${__field.labels.service_name}')
+      + g.panel.stat.options.reduceOptions.withCalcs(['lastNotNull'])
+      + g.panel.stat.options.withGraphMode('none'),
+
+    instanceLatency:
+      commonlib.panels.generic.stat.base.new(
+        'Latency',
+        targets=[signals.instance.commandLatency.asTarget()],
+        description='Average command latency.'
```
Dasomeone, line 102: Is latency not covered by the updated sample app?


Image

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+          signals.instance.opCountersTotal.asTarget(),
+          signals.instance.opCountersReplTotal.asTarget(),
+          signals.instance.ttlDeletedDocuments.asTarget(),
+        ],
+        description='Rate of command operations by type.'
+      )
+      + withStacking
+      + g.panel.timeSeries.standardOptions.withUnit('ops'),
+
+    instanceConnections:
```
Dasomeone, line 144: Missing instance detail in legend 

Image

mongodb-mixin/panels/instance.libsonnet
Dasomeone, line 1: I won't comment on it throughout, but most if not all the panels in this dashboard are missing instance detail in the legends. A little hard to validate due to sample-app coverage (might need more comprehensive load-gen?)

Image

mongodb-mixin/panels/cluster.libsonnet
Dasomeone, line 1: Still largely broken on latest sample-app

Image

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,337 @@
...
+    clusterShardServicesQps:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Shard services QPS - $shard',
+        targets=[signals.cluster.shardServicesQps.asTarget()],
+        description='Queries per second for shard services.'
+      )
+      + withTableLegend
+      + g.panel.timeSeries.standardOptions.withUnit('ops')
+      + g.panel.timeSeries.panelOptions.withRepeat('shard'),
+
```
Dasomeone, line 161: Nothing wrong here, just wanted to comment and say I like it. Only downside is this *can* result in some whitespace on odd configurations/amounts of replicasets but that's acceptable

Dasomeone (commented):
Couple more comments, I think some of this is just due to sample-app loadgen/coverage but the cluster dashboard definitely is still broken

mongodb-mixin/panels/cluster.libsonnet
mariaalons: Added loadgen to the sample-app and make some changes to the dashboard 🙏

mongodb-mixin/panels/instance.libsonnet
mariaalons: covered now, it was a load generation gap, Latency uses type="command" and without activity there wasn't any data

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+
+    replicasetElections:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Elections',
+        targets=[signals.replicaset.elections.asTarget()],
+        description='Number of replica set elections.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    replicasetHeartbeatTime:
```
Dasomeone, line 118: Looks like a unit or calculation issue here as its off by several orders of magnitude.


Image

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+    replicasetReplicationLag:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Replication lag',
+        targets=[signals.replicaset.replicationLag.asTarget()],
+        description='Replication lag per secondary member.'
+      )
+      + withTableLegend
+      + g.panel.timeSeries.standardOptions.withUnit('s'),
+
+    replicasetOperations:
```
Dasomeone, line 98: No data on this one in the sample-app

mongodb-mixin/panels/replicaset.libsonnet
```diff
@@ -0,0 +1,186 @@
...
+
+    replicasetOplogBufferedOps:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Oplog buffered operations',
+        targets=[signals.replicaset.oplogBufferCount.asTarget()],
+        description='Number of operations in the oplog buffer.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    replicasetOplogGetmoreTime:
```
Dasomeone, line 143: Also no loadgen for oplog, and the recovery window seems to be scaling despite no operations (being shown, at least)

Image

mongodb-mixin/panels/instance.libsonnet
Dasomeone: A little better now, but still missing a lot of key information like command and document operations that I'd have expected the updated loadgen to add.

Added a few more comments that may apply more broadly, but I'd be curious to see if you have screenshots with more data? Maybe there's something up with my sample-app setup

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+        '3': { index: 3, text: 'RECOVERING' },
+        '5': { index: 4, text: 'STARTUP2' },
+        '6': { index: 5, text: 'UNKNOWN' },
+        '7': { index: 6, text: 'ARBITER' },
+        '8': { index: 7, text: 'DOWN' },
+        '9': { index: 8, text: 'ROLLBACK' },
+        '10': { index: 9, text: 'REMOVED' },
+      }),
+
+    instanceUptime:
```
Dasomeone, line 67: A table overview of instances may be better for the entire overview row. Currently unreadable with the default replicaset

Thoughts?

Image

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,201 @@
...
+
+    instanceQueuedOps:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Queued operations',
+        targets=[signals.instance.queuedOps.asTarget()],
+        description='Number of operations queued due to a lock.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    instanceCursors:
```
Dasomeone, line 176: We'll want to have a filter on values that are not 0, this is currently very hard to read
Image

mongodb-mixin/panels/instance.libsonnet
Dasomeone: Hmm there's still no latency on mine despite updating the sample app to latest and verifying the loadgen is running

mongodb-mixin/panels/overview.libsonnet
```diff
@@ -0,0 +1,147 @@
...
+    overviewConnectionsByInstance:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Connections by instance',
+        targets=[signals.overview.connectionsByInstance.asTarget()],
+        description='Current connections per instance.'
+      )
+      + withTableLegend
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    overviewOpsByInstance:
```
Dasomeone, line 129: Aside from this and "total qps" this dashboard has data in all the panels.
Seems to be a recurring issue for operations metrics?

mongodb-mixin/panels/cluster.libsonnet
Dasomeone: Can you share a screenshot? While the latest sample-app with loadgen did improve things somewhat it's still largely empty for me

Dasomeone: Suspect we may need to adapt the alloy config if it's not just an issue on my end? 

Tl;dr, I **have**:
- Cursor panels working
- Connections panels working
- indexes per shard and size of indexes per shard, but no dynamic panels
- # of chunks in shards
- First & second row of overview working


What **I do not have** is:
- **Anything with QPS**
- **Anything with Operations**
- Dynamic chunks + chunk split events & move event
- Dynamic indexes and Dynamic of indexes sizes (which are also oddly named?)

mongodb-mixin/panels/instance.libsonnet
Dasomeone: Image

Image

Dasomeone (commented):
Some improvements with the latest sample-app, but still basically no working panels for operations, latency, and qps

mariaalons:
> Some improvements with the latest sample-app, but still basically no working panels for operations, latency, and qps

Found the problem with the no data panels, it was related to the scrape interval not being set on the sample-app, I added it to the open PR https://github.com/grafana/integration-sample-apps/pull/191 and now all panels have data 🙇

mongodb-mixin/panels/instance.libsonnet
mariaalons: You right looks much better like a table 

Screenshot 2026-07-10 at 17 32 16

mariaalons:
Sharing some screenshots with the latest changes 


Screenshot 2026-07-10 at 12 34 24
Screenshot 2026-07-10 at 17 29 04
Screenshot 2026-07-10 at 17 30 43
Screenshot 2026-07-10 at 17 34 21

mongodb-mixin/panels/replicaset.libsonnet
mariaalons: yes 8 years doesn't sound right 🤦  fixed 

Screenshot 2026-07-10 at 17 49 16

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,244 @@
...
+        'Command operations',
+        targets=[
+          signals.instance.opCountersTotal.asTarget() + g.query.prometheus.withInterval('2m'),
+          signals.instance.opCountersReplTotal.asTarget() + g.query.prometheus.withInterval('2m'),
+          signals.instance.ttlDeletedDocuments.asTarget() + g.query.prometheus.withInterval('2m'),
+        ],
+        description='Rate of command operations by type.'
+      )
+      + withStacking
+      + g.panel.timeSeries.standardOptions.withUnit('ops'),
```
Dasomeone, line 142: This one probably shouldn't have `withStacking` as it's not a total of operations, nor is it properly grouped by type or service like the description implies.

Image

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,244 @@
...
+
+    instanceConnections:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Connections',
+        targets=[signals.instance.connectionsCurrent.asTarget()],
+        description='Number of current connections.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    instanceDocumentOps:
```
Dasomeone, line 152: +1 for not using `withStacking` unless you're wanting absolute totals 

Image

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,244 @@
...
+    instanceQueryEfficiency:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Query efficiency',
+        targets=[
+          signals.instance.queryEfficiencyDoc.asTarget() + g.query.prometheus.withInterval('2m'),
+          signals.instance.queryEfficiencyIndex.asTarget() + g.query.prometheus.withInterval('2m'),
+        ],
+        description='Ratio of documents returned vs scanned, and index usage.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('percentunit'),
```
Dasomeone: This one might benefit from some added description or coloured thresholds for what's considered good efficiency and what's not, assuming we have solid numbers on that (not sure)

mongodb-mixin/panels/instance.libsonnet
```diff
@@ -0,0 +1,244 @@
...
+
+    instanceGetLastErrorWriteTime:
+      commonlib.panels.generic.timeSeries.base.new(
+        'getLastError write time',
+        targets=[signals.instance.getLastErrorWriteTime.asTarget() + g.query.prometheus.withInterval('2m')],
+        description='Rate of getLastError write wait time.'
+      )
+      + g.panel.timeSeries.standardOptions.withUnit('ms'),
+
+    instancePageFaults:
```
Dasomeone, line 234: Not sure what's causing it here, but the default y-axis size here is rather larger than it really should be. At a glance it's hard to see they're all at 0 given the max is 100. 

I'd recommend maybe having a `>0` filter on the query and/or adjusting the axis scaling

Image

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,337 @@
...
+          + g.query.prometheus.withFormat('table'),
+        ],
+        description='Size of collections per shard and database.'
+      )
+      + g.panel.table.standardOptions.withUnit('bytes'),
+
+    // QPS by service type
+    clusterShardServicesQps:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Shard services QPS - $shard',
```
Dasomeone, line 154: Given this is duplicated per shard we don't need a right-aligned table since it's guaranteed to have one entry. Instead here I'd recommend a bottom-aligned entry but keeping the table format so we still have name/last/min/mean/max like so

Image

mongodb-mixin/rows.libsonnet
```diff
@@ -0,0 +1,163 @@
...
+
+    clusterChunks:
+      g.panel.row.new('Chunks in shards')
+      + g.panel.row.withCollapsed(false)
+      + g.panel.row.withPanels([
+        this.grafana.panels.clusterChunksInShardsTable { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.clusterChunksDynamic { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.clusterChunkSplitEvents { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.clusterChunkMoveEvents { gridPos+: { w: 12, h: 8 } },
+      ]),
```
Dasomeone, line 116: Missing a value override for the number, which currently just shows `Value`

In general the columns here need a consistency pass on capitalisation and such

Image

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,337 @@
...
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
+
+    clusterConnectionsAvailable:
+      commonlib.panels.generic.timeSeries.base.new(
+        'Available connections per instance',
+        targets=[signals.cluster.connectionsAvailable.asTarget()],
+        description='Available connections per instance.'
+      )
+      + withTableLegend
+      + g.panel.timeSeries.standardOptions.withUnit('short'),
```
Dasomeone, line 278: For these two I'd suggest slightly taller panels + bottom-aligned tables if we agree there'll only be 3-4 entries. Generally we keep right-aligned tables for >5 more metrics

Example here where I modified one panel
 

Image

mongodb-mixin/panels/cluster.libsonnet
```diff
@@ -0,0 +1,337 @@
...
+    clusterCursorsTotal:
+      commonlib.panels.generic.stat.base.new(
+        'MongoDB cursors',
+        targets=[signals.cluster.cursorsTotal.asTarget()],
+        description='Total number of open cursors across the cluster.'
+      )
+      + g.panel.stat.standardOptions.withUnit('short')
+      + g.panel.stat.options.reduceOptions.withCalcs(['lastNotNull']),
+
+    clusterCursorsByInstance:
```
Dasomeone, line 332: +1 bottom aligned table

mongodb-mixin/rows.libsonnet
```diff
@@ -0,0 +1,163 @@
...
+        this.grafana.panels.clusterCursorsByInstance { gridPos+: { w: 24, h: 8 } },
+      ]),
+
+    clusterAdditionalInfo:
+      g.panel.row.new('Additional info')
+      + g.panel.row.withCollapsed(true)
+      + g.panel.row.withPanels([
+        this.grafana.panels.clusterReplicationLagBySet { gridPos+: { w: 12, h: 8 } },
+        this.grafana.panels.clusterOplogRangeBySet { gridPos+: { w: 12, h: 8 } },
+      ]),
```
Dasomeone, line 161: This is very vague and hidden away at the end. Generally don't leave single rows collapsed on dashboards that require scrolling, especially at the end as they never get used.

I'd give this a more descriptive name and have it open by default. Replication lag would be fairly important at times

mongodb-mixin/panels/instance.libsonnet
Dasomeone, line 1: There's a couple panels on this panel file and the cluster panels that have a mismatch of table legends and normal legends.

Sometimes there's table legends for 1-3 metrics (which is fine at times if we want to display columnar data), but other times like on the instance dashboard's `command overview` we have many metrics in normal format when we default to table format at `> 5` series

Example here, might be worth doing a sanity pass or adding additional filtering to remove series at 0.

Image

Dasomeone (commented):
Another batch of comments, will leave it here for now :) 

Sample app works fantastically now though, so very well done on that front!

mariaalons:
> Another batch of comments, will leave it here for now :)
> 


Thank you for the review @Dasomeone 
Applied all the requested changes, I also applied some of the comments to other dashboards 🤞

mongodb-mixin/panels/instance.libsonnet
Dasomeone, line 1: Couple issues with the legends on this dashboard, appearing across multiple panels. Looks like it might be the `or vector(0)` entries doing it. We've seen that before :/

Image

Dasomeone (commented):
Just one item left, otherwise looks stellar!

mongodb-mixin/panels/instance.libsonnet
mariaalons: I missed that one 🤦‍♀️  I applied the change 🙏

Dasomeone (approved):
Fantastic work, thank you Maria!

PR #1630, opened 2026-04-27
fix: influxdb improper self-contained import of g.libsonnet

schmikei:
InfluxDB mixin: fix g.libsonnet import for vendored use

Importing the mixin with jsonnet-bundler failed with 

```
couldn't open import "../g.libsonnet" from influxdb-mixin/dashboards.libsonnet. The rest of the mixin already uses ./g.libsonnet; only dashboards.libsonnet used ../g.libsonnet,

```

which resolves to the jsonnet-libs repo root but not when only the influxdb-mixin subtree is vendored.

Use ./g.libsonnet so the dashboard code resolves g from the same package as the other mixin files, with no dependency on the parent tree.

### DIFF
--- a/influxdb-mixin/dashboards.libsonnet
+++ b/influxdb-mixin/dashboards.libsonnet
@@ -1,4 +1,4 @@
-local g = import '../g.libsonnet';
+local g = import './g.libsonnet';
 local logslib = import 'logs-lib/logs/main.libsonnet';
 
 {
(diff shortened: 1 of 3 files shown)


mariaalons (approved):
Nice catch! LGTM! 🙇

Dasomeone (approved):

PR #1633, opened 2026-05-12
feat: add metrics-endpoint-observ-lib

theSuess:
Adds a new observ-lib to give an overview on scrape statistics for metrics-endpoint based jobs in Grafana Cloud (e.g Supabase, Temporal, Anthropic + more to come)

Looks like this (without any filter):
image

And can be customized to be shipped as part of specific integrations (also filters for relevant scrape jobs only):
image

### DIFF
--- a/metrics-endpoint-observ-lib/README.md
+++ b/metrics-endpoint-observ-lib/README.md
@@ -1,0 +1,30 @@
+# Metrics-endpoint observability lib
+
+This lib can be used to generate dashboards, rows, or panels of metrics-endpoint scraping signals.
+
+## Import
+
+```sh
+jb init
+jb install https://github.com/grafana/jsonnet-libs/metrics-endpoint-observ-lib
+```
+
+## Usage
+
+This library is intended to be used together with other integrations to offer
+insight into the Grafana Cloud metrics endpoint scraping process.  As such, it
+can be modified to show branding and documentation of another integration:
+
+```jsonnet
+local observlib = import 'main.libsonnet';
+local mixin =
+  (observlib.new()
+   + observlib.withConfigMixin({
+     filteringSelector: 'job=~"some-specific-scrape"',
+     parentIntegration: {
+       name: 'My application',
+       id: 'my-application',
+       logoURL: '<application-logo-url>',
+     },
+   })).asMonitoringMixin();
+```
--- a/metrics-endpoint-observ-lib/config.libsonnet
+++ b/metrics-endpoint-observ-lib/config.libsonnet
@@ -1,0 +1,29 @@
+{
+  // any modular library should include as inputs:
+  // 'dashboardNamePrefix' - Use as prefix for all Dashboards and (optional) rule groups
+  // 'filteringSelector' - Static selector to apply to ALL dashboard variables of type query, panel queries, alerts and recording rules.
+  // 'groupLabels' - one or more labels that can be used to identify 'group' of instances. In simple cases, can be 'job' or 'cluster'.
+  // 'instanceLabels' - one or more labels that can be used to identify single entity of instances. In simple cases, can be 'instance' or 'pod'.
+  // 'uid' - UID to prefix all dashboards original uids
+  local this = self,
+  filteringSelector: 'job!=""',
+  groupLabels: ['job'],
+  instanceLabels: ['scrape_job'],
+  dashboardTags: [self.uid],
+  uid: 'metrics-endpoint',
+  dashboardNamePrefix: 'Metrics endpoint ',
+
+  // additional params can be added if needed
+  dashboardPeriod: 'now-1h',
+  dashboardTimezone: 'default',
+  dashboardRefresh: '1m',
+
+  metricsSource: ['prometheus'],
+  signals: (import './signals/metrics-endpoint.libsonnet')(this),
+
+  parentIntegration: {
+    id: 'metrics-endpoint',
+    name: 'Metrics Endpoint',
+    logoURL: 'https://storage.googleapis.com/grafanalabs-integration-logos/metrics-endpoint-prometheus.svg',
+  },
+}
--- a/metrics-endpoint-observ-lib/jsonnetfile.json
+++ b/metrics-endpoint-observ-lib/jsonnetfile.json
@@ -1,0 +1,33 @@
+{
+  "version": 1,
+  "dependencies": [
+    {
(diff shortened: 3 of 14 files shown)


metrics-endpoint-observ-lib/jsonnetfile.json
```diff
@@ -0,0 +1,33 @@
...
+          "subdir": "common-lib"
+        }
+      },
+      "version": "master"
+    },
+    {
+      "source": {
+        "git": {
+          "remote": "https://github.com/grafana/grafonnet.git",
+          "subdir": "gen/grafonnet-v11.0.0"
```
mariaalons: I think the version here should be `grafonnet-v11.4.0` the same one that `g.libsonnet` imports

metrics-endpoint-observ-lib/prometheus_rules_out/prometheus_alerts.yaml
```diff
@@ -0,0 +1,3 @@
+groups:
+    - name: golang-alerts-golang
+      rules: []
```
mariaalons: This one looks like a left over, if there is no alerts this should be just and empty file like `prometheus_rules.yaml`, you can remove it and run `make prometheus_alerts.yaml`

metrics-endpoint-observ-lib/config.libsonnet
```diff
@@ -0,0 +1,29 @@
...
+
+  metricsSource: ['prometheus'],
+  signals: (import './signals/metrics-endpoint.libsonnet')(this),
+
+  parentIntegration: {
+    id: 'metrics-endpoint',
+    name: 'Metrics Endpoint',
+    logoURL: 'https://storage.googleapis.com/grafanalabs-integration-logos/metrics-endpoint-prometheus.svg',
+  },
+}
```
Dasomeone, line 29: On the fence with this, we generally want to avoid referencing the integrations here, but I think in this case it makes sense. 

Fine either way, but think it's worth keeping in mind. Thoughts @mariaalons @theSuess?

metrics-endpoint-observ-lib/README.md
```diff
@@ -0,0 +1,30 @@
...
+This library is intended to be used together with other integrations to offer
+insight into the Grafana Cloud metrics endpoint scraping process.  As such, it
+can be modified to show branding and documentation of another integration:
+
+```jsonnet
+local observlib = import 'main.libsonnet';
+local mixin =
+  (observlib.new()
+   + observlib.withConfigMixin({
+     filteringSelector: 'job=~"integrations/supabase/.+"',
```
Dasomeone: I know you used this for debugging but please refrain from using the `integrations/` jobs in jsonnet-libs. Any other example would work here in the readme (Debug is fine, I know that'll actually get used)

metrics-endpoint-observ-lib/panels.libsonnet
```diff
@@ -0,0 +1,55 @@
...
+      ,
+      topCardinality: signals.topCardinality.asTimeSeries()
+                      + commonlib.panels.generic.timeSeries.base.stylize()
+                      + g.panel.timeSeries.options.legend.withDisplayMode('table')
+                      + g.panel.timeSeries.options.legend.withPlacement('right')
+                      + g.panel.timeSeries.options.legend.withCalcs(['lastNotNull'])
+      ,
+      scrapeDuration: signals.scrapeDuration.asTimeSeries()
+                      + commonlib.panels.generic.timeSeries.base.stylize(),
+    },
```
Dasomeone, line 54: Not sure if it's just my review/file here but while it works the indentation is really wonky here

Dasomeone (commented):
Couple comments in addition to Maria's, but overall this is a solid idea! Thanks for doing this!

metrics-endpoint-observ-lib/prometheus_rules_out/prometheus_alerts.yaml
theSuess: Yup, used go as the base and commited this by accident ^^

metrics-endpoint-observ-lib/jsonnetfile.json
theSuess: True! Will update

metrics-endpoint-observ-lib/panels.libsonnet
theSuess: Advantage of this style is easy reordering of lines without worrying about the trailing comma. I'm fine with changing this though

metrics-endpoint-observ-lib/config.libsonnet
theSuess: We have [grafana-cloud-integration-utils](https://github.com/grafana/jsonnet-libs/tree/master/grafana-cloud-integration-utils) in this repo as well so I assumed mentioning integrations is fine. We can of course move this to the internal repo as well to not confuse people

metrics-endpoint-observ-lib/README.md
theSuess: I've replaced this with generic text now. Still, this lib only makes sense in the context of grafana-cloud integrations (`metrics_endpoint_export_response_success` is a metric only generated by metrics-endpoint scrapes). As mentioned in another comment, we can also move this to the interal repo

metrics-endpoint-observ-lib/panels.libsonnet
Dasomeone: No it's fine, just wanted to make sure it was on purpose! Actually forgot to remove that comment as I wrote it before testing if the mixin generated properly :)

metrics-endpoint-observ-lib/README.md
Dasomeone: We could move it, but for now generic text will be alright. Can look at moving it in the future if we want to?
We do have a couple other mentions of `job: integrations/` elsewhere here but we're trying to thin them out as we update :)

metrics-endpoint-observ-lib/config.libsonnet
Dasomeone: Yeah it's fine to mention, just we try to avoid putting it in the readme's/snippets to copy. So debugging files are fine, but the rest we try to strip it from

Dasomeone (approved):

mariaalons (approved):
Looks good!! thank you!

PR #1635, opened 2026-05-14
fix(apache-hbase-mixin) panel not hiding irrelevant columns

schmikei:
Not sure when this changed from Value #A vs Value #B but we were getting this columns improperly getting excluded

Without the fix:

image

With the fix:

image


mariaalons (approved):
LGTM, Good catch!

PR #1638, opened 2026-05-15
chore(opensearch): remove hardcoded datasource

v-zhuravlev:
breaks when logsEnabled: false is set.


Dasomeone (approved):
Thanks for catching this @v-zhuravlev, looks like a leftover from the migration :)

PR #1644, opened 2026-06-03
feat(common-lib): add signals.collectMetricExprs() helper  

mariaalons:
## Description 
 
 Adds `collectMetricExprs(signalsObj)` to the signal lib. It returns the templated PromQL expression for each signal, with   
 stub signals filtered out. No deduplication, that's left to the caller.
                                                                     
  This is a building block for letting metric-extraction tooling pick up signal-defined metrics, additive only nothing calls it yet. Includes a test.


Dasomeone (approved):
Nice, thanks for tracking this down! Looks clean :)

PR #1645, opened 2026-06-04
feat: preserve asserts_env and asserts_site in GCP and Azure alert rules

anaivanov:
Added `asserts_env, asserts_site` to every `by()` aggregation clause in gcp and azure alerts.

Once these rules are deployed and new alert instances fire, they will carry `asserts_env` and `asserts_site` labels.

Also, update grafonnet from `v11.0.0` to `v11.4.0`


crowleash (dismissed):

Dasomeone (approved):
LGTM 🚀

PR #1649, opened 2026-06-09
fix(signals): collect metric exprs from grouped signals

mariaalons:
## Description 

 Follow up on #1644. The previous version didn't pick up grouped signals `collectMetricExprs` now recurses into nested maps so both flat and grouped signals are collected.

### DIFF
--- a/common-lib/common/signal/signal.libsonnet
+++ b/common-lib/common/signal/signal.libsonnet
@@ -388,11 +388,14 @@ local xtd = import 'github.com/jsonnet-libs/xtd/main.libsonnet';
   collectMetricExprs(signalsObj):
     std.filter(
       function(e) std.isString(e) && e != '',
-      [
-        signalsObj[s].asPanelExpression()
+      std.flattenArrays([
+        if std.isObject(signalsObj[s]) && std.objectHasAll(signalsObj[s], 'asPanelExpression') then
+          [signalsObj[s].asPanelExpression()]
+        else if std.isObject(signalsObj[s]) then
+          self.collectMetricExprs(signalsObj[s])
+        else
+          []
         for s in std.objectFields(signalsObj)
-        if std.isObject(signalsObj[s])
-           && std.objectHas(signalsObj[s], 'asPanelExpression')
-      ]
+      ])
     ),
 }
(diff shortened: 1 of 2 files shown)


common-lib/common/signal/signal.libsonnet
```diff
@@ -388,11 +388,14 @@ local xtd = import 'github.com/jsonnet-libs/xtd/main.libsonnet';
   collectMetricExprs(signalsObj):
     std.filter(
       function(e) std.isString(e) && e != '',
-      [
-        signalsObj[s].asPanelExpression()
+      std.flattenArrays([
```
Dasomeone, line 391: Just double checking here, `flattenArrays` flattens only one layer of nesting. Are we good on this front? (I think so, but I've not double checked)

common-lib/common/signal/test_collect_metric_exprs.libsonnet
```diff
@@ -74,6 +80,18 @@ local exprs = signal.collectMetricExprs(signals);
...
+        expect: 4,
+      },
+      testGroupedAllStrings: {
+        actual: std.all([std.isString(e) for e in groupedExprs]),
+        expect: true,
+      },
+      testGroupedContainsAbc: {
+        actual: std.member(groupedExprs, 'avg by (job,xxx) (\n  abc{job="integrations/agent",job=~"$job",instance=~"$instance"}\n)'),
+        expect: true,
+      },
```
Dasomeone: Minor nit, but since these use distinctly different test input, could you please make a new highlevel group for them so they're more easily distinguished from the ungrouped metrics in `exprs`?

Dasomeone (commented):
Two minor questions

common-lib/common/signal/signal.libsonnet
mariaalons: The `flattenArrays` here only needs one level because the recursion guarantees each `collectMetricExprs` call returns an already flatten array of strings so at every level its input is just an array of flat arrays. 
The recursion handles depth, flatten just does the single join per level. I tested it and confirmed with a 3 level deep fixture 🙏

Dasomeone: Oh shoot I missed the recursion :sweat_smile:

Dasomeone (approved):

PR #1650, opened 2026-06-09
fix!(opensearch) change false-positive OpenSearch memory alerts (Switch from OS Memory to JVM heap)

v-zhuravlev:
The previous alerting rules evaluated `opensearch_os_mem_used_percent` metric. 

Because OpenSearch aggressively relies on the OS page cache to store Lucene segments for fast querying, this metric naturally could float above 95% up to 100%, while OS working perfectly normal.

This resulted in constant false-positive alerts.

On the contrary, in my environment, the actual K8s pod memory metrics showed a healthy working set (~10-11 GiB out of the 16 GiB limit).

This changes:
- Renamed the alert to OpenSearchHighJVMHeapUsage for clarity.
- Replaced the node-level OS memory metric with `opensearch_jvm_mem_heap_used_percent` to control JVM memory pressure instead.

more ref:
https://repost.aws/knowledge-center/opensearch-high-sysmemoryutilization
https://forum.opensearch.org/t/default-consumption-of-os-memory-usage/17542

### DIFF
--- a/opensearch-mixin/README.md
+++ b/opensearch-mixin/README.md
@@ -16,7 +16,7 @@ and the following alerts:
 - OpenSearchUnstableShardUnassigned
 - OpenSearchHighNodeDiskUsage (warning and critical)
 - OpenSearchHighNodeCpuUsage (warning and critical)
-- OpenSearchHighNodeMemoryUsage (warning and critical)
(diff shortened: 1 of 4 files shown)


Dasomeone (approved):
Makes sense to me, thanks for catching this one Vitaly!

PR #1653, opened 2026-06-17
temporal-mixin: add Provisioned Capacity (TRU) section to overview dashboard

jpvotta:
## What
Adds a **Provisioned Capacity (TRU)** section:
image
- Namespaces using provisioned capacity and TRUs, APS, RPS, and OPS
- 1/7/30-day approximate Capacity utilization (Provisioned only)
- Per-namespace capacity utilization (Provisioned only)
- Per-namespace APS vs Limit (Provisioned only)

Also tidies the Usage & Quotas panel:
image
- Remove `or vector(...)` limit fallbacks since these can be inaccurate if manually tuned, and also these previously reflected the old world where namespaces had 400/1600/3200 APS/RPS/OPS by default
- Add namespace to legends
- Color by query (usage green, limit dark-red dashed) so it holds across namespaces

Also did the same for the Service panel:
image


temporal-mixin/dashboards/temporal-overview.json
```diff
@@ -8113,6 +8173,636 @@
...
+      },
+      "pluginVersion": "12.3.0-18186912471",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "sum by (temporal_namespace) (last_over_time(temporal_cloud_v1_action_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]) != last_over_time(temporal_cloud_v1_action_on_demand_envelope_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]))",
```
theSuess, line 8260: Since this is a time series panel, can we get rid of the `last_over_time` function? I don't think it's needed for time series panels (same for the other time series panels)

temporal-mixin/dashboards/temporal-overview.json
```diff
@@ -8113,6 +8173,636 @@
...
+      },
+      "pluginVersion": "12.3.0-18186912471",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "sum ( sum_over_time( ( sum by (temporal_namespace) (temporal_cloud_v1_total_action_count{temporal_namespace=~\"$temporal_namespace\"}) and on (temporal_namespace) ( last_over_time(temporal_cloud_v1_action_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]) != last_over_time(temporal_cloud_v1_action_on_demand_envelope_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]) ) )[30d:1m]) ) / sum ( sum_over_time( ( last_over_time( temporal_cloud_v1_action_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]) != last_over_time(temporal_cloud_v1_action_on_demand_envelope_limit{temporal_namespace=~\"$temporal_namespace\"}[1h]) )[30d:1m] ) )",
```
theSuess: You don't need to specify the `1m` resolution here. Grafana will pick the best value automatically based on selected time range & data source settings (same for the other queries where this is used)

theSuess (changes requested):
Thanks for the PR!
Overall, this looks good but I left two comments on the queries which should be easy to fix

temporal-mixin/dashboards/temporal-overview.json
jpvotta: Unfortunately these capacity metrics are only being emitted every ~6-7 minutes, so we need the lookback. We are investigating internally why this is the case, and I'll come update this again to remove the lookback once we get these metrics onto the regular cadence :)

temporal-mixin/dashboards/temporal-overview.json
jpvotta: Agreed and done

theSuess (approved):

temporal-mixin/dashboards/temporal-overview.json
theSuess: Emitted meaning they're only part of the endpoint response _sometimes_ or that changes are delayed by this amount? In the second case, I'd prefer using the specific values but if this is an issue on the exporter, I'm fine with merging this for now

jpvotta: It's the former. The metrics are being generated by an internal workflow which takes ~5 minutes to complete, so we are only actually emitting one datapoint every ~5-7 minutes. We have plans to eventually improve the performance here, but for now this is the state.

PR #1661, opened 2026-06-29
chore: Modernize signals catchpoint-mixin

postnati:
This PR modernizes catchpoint-mixin to use the signals architecture pattern.


catchpoint-mixin/panels.libsonnet
```diff
@@ -5,18 +5,68 @@ local utils = commonlib.utils;
...
+
+      // Overview: topk(1, sum by (<by>) (sum_over_time(<metric>[$__interval:])))
+      local topSum(signal, by) =
+        signal
+        .withFilteringSelectorMixin(pureSel)
+        .withExprWrappersMixin(['sum_over_time(', '[$__interval:])'])
+        .withExprWrappersMixin(['sum by (' + by + ') (', ')'])
+        .withTopK(1)
+        .withLegendFormat('{{' + by + '}}')
+        .asTarget(),
```
Dasomeone: What's all this for?

Dasomeone (changes requested):
Stylistically differs from the other PRs and needs the same overhaul as mentioned in comments around panels here https://github.com/grafana/jsonnet-libs/pull/1668#pullrequestreview-4763979143

catchpoint-mixin/panels.libsonnet
postnati: topAvg and sumBy each wrap the same aggregation chain across 10 and 76 panels. Just trying to remove duplication.

catchpoint-mixin/signals/network.libsonnet
```diff
@@ -0,0 +1,187 @@
...
+            expr: 'catchpoint_hosts_count{%(queriesSelector)s}',
+          },
+        },
+      },
+      cachedCount: {
+        name: 'Cache access',
+        nameShort: 'Cache hits',
+        type: 'gauge',
+        description: 'Number of cached elements accessed.',
+        unit: '',
```
mariaalons, line 119: nit: We can use `unit: 'short'` instead of a mix of `'none' and ' '`

catchpoint-mixin/panels.libsonnet
```diff
@@ -340,7 +407,10 @@ local utils = commonlib.utils;
       contentHandlingNodeName:
         commonlib.panels.generic.timeSeries.base.new(
           'Content handling',
-          targets=[t.contentHandlingLoad, t.contentHandlingRender],
+          targets=[
+            sumBy(signals.timing.contentLoadTime, 'node_name', '{{node_name}} - load').asTarget(),
+            sumBy(signals.timing.renderStartTime, 'node_name', '{{node_name}} - render').asTarget(),
+          ],
```
mariaalons, line 413: This two queries are the same as the one on the `contentHandling:` panel, I think this is a bug from before, this one should be `'test_name', '{{test_name}}` instead of ` 'node_name', '{{node_name}}`

mariaalons (commented):
Nice work! I see the asTarget() was fixed , but there is still some parts mentioned on the previous review that haven't been addressed like the topAvg/sumBy is still rebuild aggregation at the panel layer, it might be better to move those to the signal side.

catchpoint-mixin/signals/network.libsonnet
Dasomeone: Not just can, that's a requirement :)

PR #1666, opened 2026-06-29
chore: Modernize signals openstack-mixin

postnati:
This PR modernizes openstack-mixin to use the signals architecture pattern.


openstack-mixin/panels.libsonnet
Dasomeone, line 1: Needs the same overhaul as mentioned in comments around panels here https://github.com/grafana/jsonnet-libs/pull/1668#pullrequestreview-4763979143

Dasomeone (changes requested):

openstack-mixin/panels.libsonnet
postnati: Updated. Ready for review.

postnati: Converted the signals from raw to typed and used appropriate aggFunction, rangeFunction, and exprWrappers.

PR #1667, opened 2026-06-29
chore: Modernize signals apache-solr-mixin

postnati:
This PR modernizes apache-solr-mixin to use the signals architecture pattern.


Dasomeone (changes requested):
This doesn't have any use of the signals? While yes they've been added there's no changes to panel setup.

postnati:
The style of this one was pretty old. All the panels were still in the dashboards, and instead of breaking out the panels into a separate file, I just updated the dashboards with the signals directly. I'll try to break out the panels into a separate file and see if that causes any unintended issues.

postnati:
Converted the signals from raw to typed and used appropriate aggFunction, rangeFunction, and exprWrappers.

mariaalons (commented):
Good work with this one!,  I see the rows.libsonnet is not present here all the
panels are being placed directly in dashboards.libsonnet with the gridPos
hardcoded. It might be better to move those into a rows.libsonnet

PR #1668, opened 2026-06-29
chore: Modernize signals vsphere-mixin

postnati:
This PR modernizes vsphere-mixin to use the signals architecture pattern.


vsphere-mixin/panels.libsonnet
```diff
@@ -529,35 +529,35 @@ local utils = commonlib.utils;
...
             + g.query.prometheus.withRefId('disk_utilization')
             + g.query.prometheus.withRange(true)
             ,
-            t.hostVMNetworkThroughput + g.query.prometheus.withFormat('table')
+            signals.host.hostVMNetworkThroughput.asTarget() + g.query.prometheus.withFormat('table')
             + g.query.prometheus.withRefId('netowrk_throughput')
             + g.query.prometheus.withRange(true)
             ,
-            t.hostVMPacketDropRate + g.query.prometheus.withFormat('table')
+            signals.host.hostVMPacketDropRate.asTarget() + g.query.prometheus.withFormat('table')
```
schmikei, line 474: I think that we should use `asTableTarget()` here since these are tables we are populating (mostly inventory viewing)


```suggestion
            signals.host.hostVMPacketDropRate.asTableTarget() 
```

This also sets the "instant" field to `true` which for a table I think would be an improvement for these panels. LMK if you guys agree @Dasomeone @mariaalons 😄

Dasomeone: Sorry, missed this question in my inbox, but you're right in that we want the `instant` field set `true`. 

`asTableTarget()` is a good option here as we want to minimise direct calls to grafonnet via `g` unless strictly necessary.

Visual changes are acceptable compared to base mixin, but not what we're primarily looking for

vsphere-mixin/signals/host.libsonnet
```diff
@@ -0,0 +1,111 @@
...
+    type: 'raw',
+    unit: unit,
+    description: description,
+    sources: {
+      prometheus: {
+        expr: expr,
+        legendCustomTemplate: legend,
+      },
+    },
+  };
```
Dasomeone: Please don't do this, will make adding additional sources harder, and actively reduces the readability. Please see https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/signals/fiber.libsonnet for a better example on how to format this, and the following for the reason why we don't want to collapse it like that

https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/signals/cpu.libsonnet

**This'll need to be undone throughout both this PR and the others.**

vsphere-mixin/panels.libsonnet
```diff
@@ -14,87 +14,87 @@ local utils = commonlib.utils;
       clustersCountStatus:
         commonlib.panels.generic.stat.info.new(
           'Clusters',
-          targets=[t.clustersCount],
+          targets=[signals.overview.clustersCount.asTarget()],
           description='The number of clusters in the datacenter.'
         )
         + stat.options.withGraphMode('none'),
```
Dasomeone, line 17: You're sort of reinventing / duplicating your own work by sticking with `asTarget()` throughout. Signals has support for `asTimeSeries()`, `asTable()`, and much more, letting it pull in description and styling directly from the underlying signal.

Generally speaking `asTarget()` should only be used on compound/complex panels that use multiple signals or want to override/remix them.

See https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/panels/memory.libsonnet for a quick example of how minimal this could be by comparison, with the matching signal definition [here](https://github.com/grafana/jsonnet-libs/blob/master/snmp-observ-lib/signals/memory.libsonnet#L33)

Dasomeone (changes requested):
While the output looks great, the actual porting to Signals is fairly half-baked and doesn't really live up to the spirit of the thing. Please have a look at my comments and apply them throughout both this PR and the others, as it seems to be a recurring theme

postnati:
Updated the use of signals and removed the builder helper.

PR #1669, opened 2026-06-29
chore: Modernize signals windows-active-directory-mixin

postnati:
This PR modernizes windows-active-directory-mixin to use the signals architecture pattern.


postnati:
This is unneeded. Signals are already defined in the shared windows-observ-lib.

PR #1678, opened 2026-07-23
chore: Modernize signals velero-2-mixin

postnati:
This PR modernizes velero-2-mixin to use the signals architecture pattern.


Dasomeone (changes requested):
Needs the same overhaul as mentioned in comments around panels here https://github.com/grafana/jsonnet-libs/pull/1668#pullrequestreview-4763979143

postnati:
Updated and ready for review

velero-2-mixin/signals/clusteroverview.libsonnet
```diff
@@ -0,0 +1,224 @@
...
+    datasource: 'prometheus_datasource',
+    aggLevel: 'none',
+    discoveryMetric: {
+      prometheus: 'velero_backup_success_total',
+    },
+    signals: {
+      successfulBackups: {
+        name: 'Successful backups / $__interval ',
+        nameShort: 'Backups',
+        type: 'raw',
```
mariaalons: This was part of the previous review, haven't been addressed. All the declared signals are `type: raw` with the legacy expression baked in, this bypasses everything the signals layer is for.
This one for example everything in this raw expr is what the lib generates from `type: 'counter'` with the appropriate agg settings

Most signals in both files are this shape, raw should be reserved for the genuinely awkward ones

mariaalons (changes requested):
Nice work! I see the signal constructor feedback was applied to the stats and timeseries panels, but parts of the previous review still stand 🙏

velero-2-mixin/signals/clusteroverview.libsonnet
postnati: Converted the signals from raw to typed and used appropriate aggFunction, rangeFunction, and exprWrappers.

mariaalons (commented):
This one is looking better but I see some similarities with the Apache-solr one https://github.com/grafana/jsonnet-libs/pull/1667#pullrequestreview-4908257722
It is missing the rows.libsonnet file, also the cluster overview doesn't have any rows at all, could we group those 10 panels the same way the overview does?

Issue #1683, opened 2026-08-03
Update license file to a full Apache 2.0 license instead of Apache 2.0 file header


EmilVorre:
*Summary*
Im working with auto extracting licenses from repo's, and i have run into a problem as jsonnet-libs LICENSE file as not a standard recognised format of SPDX as it is a file header text and not the license text. To fix this i would suggest replacing the file with the full Apache License 2.0 text. so that automated tools can recognize it.

PR #1684, opened 2026-08-03
chore: update license file to include full Apache License 2.0 text

EmilVorre:
added the full text for an Apache license so that automatic tools can recognize it as an Apache License 2.0


EmilVorre:
This is done as a fix for issue https://github.com/grafana/jsonnet-libs/issues/1683

PR #1685, opened 2026-08-05
feat(nginx-directory): allow extra http config in ngifx.conf

ashwanthgoli:
Allows additional directives to be added to http block. It indents the provided config by two spaces.

```
extra_http_config: |||
    map $uri $loki_tenant_id {
      default "";
      "~^/[^/]+/config/tenant/([^/]+)/v1/limits/?$" $1;
    }
  |||,

-- rendered as --

http {
  map $uri $loki_tenant_id {
    default "";
    "~^/[^/]+/config/tenant/([^/]+)/v1/limits/?$" $1;
  }
...
}
```

### DIFF
--- a/nginx-directory/files/nginx.conf
+++ b/nginx-directory/files/nginx.conf
@@ -7,7 +7,7 @@ events {
   worker_connections  4096;  ## Default: 1024
 }
 
-http {
+http {%(extra_http_config_rendered)s
   default_type application/octet-stream;
   log_format   main '%(log_format)s';
   access_log   /dev/stderr  main;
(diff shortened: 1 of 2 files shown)


nginx-directory/files/nginx.conf
```diff
@@ -7,7 +7,7 @@ events {
   worker_connections  4096;  ## Default: 1024
 }
 
-http {
+http {%(extra_http_config_rendered)s
```
chaudum, line 10: I assume 
```suggestion
http {
%(extra_http_config_rendered)s
```

would cause all existing usages to render the output differently, right?

chaudum (dismissed):

nginx-directory/files/nginx.conf
ashwanthgoli: that would add a empty line everywhere

ashwanthgoli:
pushed another commit to simplify to allow a single extra config entry instead of a slice.
This keeps the logic simple and maintainable.

chaudum (approved):

PR #1686, opened 2026-08-05
fix(haproxy-mixin): read proxy status from the state label

mariaalons:
Fix the Frontend/Backend/Server status tables
Since HAProxy 2.4 the exporter emits haproxy_{frontend,backend,server}_status as one series per state with a state label, so every proxy rendered twice (as Up and Down) and the status color followed the sample value instead of the actual state.

- Filter the status queries with == 1 and build the Status column from the state label 
- Drop option `http-use-htx` from the dev HAProxy config

### DIFF
--- a/haproxy-mixin/dashboards/dashboards.libsonnet
+++ b/haproxy-mixin/dashboards/dashboards.libsonnet
@@ -59,6 +59,9 @@ local g = import 'github.com/grafana/dashboard-spec/_gen/7.0/jsonnet/grafana.lib
     processConnectionRateLimit: 'haproxy_process_limit_connection_rate{%s}' % $._config.baseMatchers,
     processSessionRateLimit: 'haproxy_process_limit_session_rate{%s}' % $._config.baseMatchers,
     processSslRateLimit: 'haproxy_process_limit_ssl_rate{%s}' % $._config.baseMatchers,
+    // Since HAProxy 2.4 PROMEX emits one *_status series per possible state with a
+    // `state` label; all have value 0 except the actual state, which is 1.
+    // Filter with == 1 and read the status from the `state` label.
     backendStatus: 'haproxy_backend_status{%s} == 1' % $._config.baseMatchers,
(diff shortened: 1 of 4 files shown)


haproxy-mixin/dashboards/dashboards.libsonnet
```diff
@@ -59,9 +59,9 @@ local g = import 'github.com/grafana/dashboard-spec/_gen/7.0/jsonnet/grafana.lib
     processConnectionRateLimit: 'haproxy_process_limit_connection_rate{%s}' % $._config.baseMatchers,
     processSessionRateLimit: 'haproxy_process_limit_session_rate{%s}' % $._config.baseMatchers,
     processSslRateLimit: 'haproxy_process_limit_ssl_rate{%s}' % $._config.baseMatchers,
-    backendStatus: 'haproxy_backend_status{%s}' % $._config.baseMatchers,
-    frontendStatus: 'haproxy_frontend_status{%s}' % $._config.baseMatchers,
-    serverStatus: 'haproxy_server_status{%s}' % $._config.backendMatchers,
+    backendStatus: 'haproxy_backend_status{%s} == 1' % $._config.baseMatchers,
+    frontendStatus: 'haproxy_frontend_status{%s} == 1' % $._config.baseMatchers,
+    serverStatus: 'haproxy_server_status{%s} == 1' % $._config.backendMatchers,
```
theSuess, line 67: Maybe we could document the way the haproxy exporter reports this.
For me, it was quite a surprise that it returns metrics for all states but only those with value `1` are the actual state

theSuess (dismissed):
Took a closer look and the haproxy exporter is indeed a bit weird around status metrics 😅

haproxy-mixin/dashboards/dashboards.libsonnet
mariaalons: I added a comment about this on the panel 🙇

theSuess (approved):

PR #1690, opened 2026-08-13
chore: Modernize velero-2-mixin

postnati:
Modernized the mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.

Cosmetic changes from row headers and library defaults. Also fixed three expressions that were querying the wrong thing, so those panels now show different values. Two were reading  `*_success_total` for attempt/failure series, and the backup success rate was dividing a count by a rate. Overview job/cluster are now multi-select.


velero-2-mixin/links.libsonnet
```diff
@@ -4,17 +4,17 @@ local g = import './g.libsonnet';
   new(this):
     {
       veleroClusterOverview:
-        link.link.new('Velero cluster view', '/d/' + this.grafana.dashboards.clusterOverview.uid)
+        link.link.new('Velero cluster view', '/d/' + this.grafana.dashboards['clusterOverview.json'].uid)
         + link.link.options.withKeepTime(true),
```
Dasomeone: Minor addition here, we'll want `withKeepVars(true)` as well for all these, should've been in the mixin already but we must've overlooked it

velero-2-mixin/main.libsonnet
```diff
@@ -17,13 +17,24 @@ local variables = import './variables.libsonnet';
 
     local this = self,
     config: config,
+    signals:
+      {
+        // legacy panel queries filter by dashboard variables only, so the static
+        // filteringSelector is kept out of dashboard queries here; the signal
+        // specs retain it for alert expressions
+        [sig]: commonlib.signals.unmarshallJsonMulti(
+          this.config.signals[sig] { filteringSelector: '' },
```
Dasomeone: This should not be stripped/overwritten here, but you can add this comment to the config as a default if you want (we'll want to override the config on the integration side)

velero-2-mixin/signals/overview.libsonnet
```diff
@@ -0,0 +1,327 @@
...
+      },
+      backupSuccessRate: {
+        name: 'Backup success rate / $__interval',
+        nameShort: 'Backup success rate',
+        type: 'raw',
+        unit: 'percentunit',
+        description: 'Success rate of backups.',
+        sources: {
+          prometheus: {
+            expr: 'increase(label_replace(velero_backup_success_total{%(queriesSelector)s}, "schedule", "none", "schedule", "^$")[$__rate_interval:]) / clamp_min(increase(label_replace(velero_backup_attempt_total{%(queriesSelector)s}, "schedule", "none", "schedule", "^$")[$__rate_interval:]),1)',
```
Dasomeone: Minor thing that I just caught here, we need to use `[$__interval:] offset -$__interval` for any `increase()`

This one must have slipped through the cracks in the original mixin. 

If you see any `increase()` calls use `$__rate_interval` please swap those over in the signal migration

Dasomeone (changes requested):
Couple small comments, but overall very close! Thanks @postnati !

velero-2-mixin/signals/overview.libsonnet
postnati: Good catch. Added.

velero-2-mixin/main.libsonnet
postnati: Fixed

velero-2-mixin/links.libsonnet
postnati: Added. Followed the pattern of all the other mixins and used withIncludeVars

Dasomeone: Perfect, I got the name wrong by memory :D

Dasomeone (approved):
Perfect, thank you for iterating on this so fast from yesterdays discussion @postnati ! 

This is exactly the kind of migration we want to see for adopting signals, just proper use of signals -> panels -> rows -> dashboards, and fixing any minor smells like the interval and link vars throughout the rest.

PR #1691, opened 2026-08-14
chore: Modernize openstack mixin

postnati:
Modernized the openstack mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


openstack-mixin/config.libsonnet
```diff
@@ -1,5 +1,7 @@
 {
-  filteringSelector: '',
+  local this = self,
+  // Static selector applied to all dashboard variable queries and alerts.
+  filteringSelector: 'job="integrations/openstack"',
```
Dasomeone: Please leave it blank in the mixin, we'll inject it in the integration

openstack-mixin/panels.libsonnet
```diff
@@ -1,22 +1,18 @@
...
 {
   new(this): {
-    local t = this.grafana.targets,
+    local signals = this.signals,
+
     local alertList = g.panel.alertList,
     local stat = g.panel.stat,
     local timeSeries = g.panel.timeSeries,
     local table = g.panel.table,
     local gauge = g.panel.gauge,
```
Dasomeone: These all need to be replaced with common-libs base variants instead. We don't want to invoke these unless strictly required.

As a result we don't want to hide calls to `g.panel.` behind helper utils. Needs to be obvious when used.

Not as bad as it sounds as they aren't used too frequently, but just to make sure, i.e. througout the file if there's a `timeSeries.new()` or `table.new()` that should be using common-lib or be very visible as a `g.panel.table` call with a comment explaining why

openstack-mixin/panels.libsonnet
```diff
@@ -162,12 +138,12 @@ local utils = commonlib.utils;
     totalResources:
       table.new('Total resources')
```
Dasomeone: I.e. this one needs to use the common base rather than `g.panel.table`

openstack-mixin/panels.libsonnet
```diff
@@ -310,40 +275,26 @@ local utils = commonlib.utils;
...
-        targets=[t.users],
-        description='The number of users for the OpenStack cloud.',
-      )
+      signals.identity.identity_users.asTimeSeries()
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + timeSeries.queryOptions.withInterval('1m')
       + timeSeries.standardOptions.withDecimals(0),
 
     projectDetails:
       table.new('Project details')
```
Dasomeone: +1 here

openstack-mixin/panels.libsonnet
```diff
@@ -530,32 +488,33 @@ local utils = commonlib.utils;
...
-        targets=[t.subnets],
-        description='The number of networks managed by Neutron.'
-      )
+      signals.neutron.neutron_subnets.asTimeSeries()
+      + commonlib.panels.generic.timeSeries.base.stylize()
+      + timeSeries.queryOptions.withInterval('1m')
       + timeSeries.standardOptions.withDecimals(0),
 
     routers:
       commonlib.panels.generic.timeSeries.base.new(
```
Dasomeone: this would be fine to chuck behind one of the helper variables if you wanted

openstack-mixin/panels.libsonnet
Dasomeone, line 1: Already commented on a few, but all the `table.new` usages need to be replaced with common-lib table.base.new

Dasomeone (commented):
Signals all look great, just a few structural changes to common-lib usage in panels and then similar to what you had on the other PR, can you adjust the output of dashboards to include `.json` filetypes please?

postnati:
Hopefully better. Ready for a re-review

Dasomeone (approved):
Thanks @postnati this covers all my feedback, looks great!

PR #1702, opened 2026-09-10
chore: Modernize pgbouncer mixin

postnati:
Modernized the pgbouncer mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


pgbouncer-mixin/variables.libsonnet
```diff
@@ -1,99 +1,70 @@
...
+      prometheusDatasourceLabel='Prometheus data source',
+    );
+
+    // The overview dashboard is a per-instance drill-down, so instance is single-select
+    // while database stays multi-select. commonlib's singleInstance would make both
+    // single-select, so narrow just the instance variable here.
+    local singleSelectInstance(variables) =
+      std.map(
+        function(v)
+          if std.get(v, 'name', '') == 'instance'
```
mariaalons, line 46: For this kind of variable override, you can use grafana-cloud-integration-utils/util.libsonnet patch_variables(dashboard, patch) instead of matching by name

pgbouncer-mixin/panels.libsonnet
```diff
@@ -1,90 +1,74 @@
...
-        + stat.standardOptions.color.withMode('thresholds')
-        + stat.standardOptions.thresholds.withSteps([
-          stat.thresholdStep.withColor('super-light-green')
-          + stat.thresholdStep.withValue(0),
-          stat.thresholdStep.withColor('super-light-orange')
-          + stat.thresholdStep.withValue(10),
-          stat.thresholdStep.withColor('super-light-red')
-          + stat.thresholdStep.withValue(20),
+        signals.connections.pools_client_waiting_connections_total.asStat()
+        + commonlib.panels.generic.stat.base.stylize()
```
mariaalons, line 10: I think you can drop the stylize call here since the colours are set right after and would override the styles added by this call

mariaalons (commented):
I left only a couple of nits this one looks good

pgbouncer-mixin/panels.libsonnet
postnati: Done. Dropped stylize call

pgbouncer-mixin/variables.libsonnet
postnati: Updated

mariaalons (approved):
Thank you for addressing the comments this one also LGTM! 🙇‍♀️

PR #1703, opened 2026-09-10
chore: Modernize catchpoint mixin

postnati:
Modernized the catchpoint mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


catchpoint-mixin/links.libsonnet
```diff
@@ -4,13 +4,16 @@ local g = import './g.libsonnet';
...
-        + link.link.options.withKeepTime(true),
+        link.link.new('Catchpoint overview', '/d/' + this.grafana.dashboards['catchpoint-overview.json'].uid)
+        + link.link.options.withKeepTime(true)
+        + link.link.options.withIncludeVars(true),
       catchpointTestNameOverview:
-        link.link.new('Catchpoint web performance by tests', '/d/' + this.grafana.dashboards.testNameOverview.uid)
-        + link.link.options.withKeepTime(true),
+        link.link.new('Catchpoint web performance by tests', '/d/' + this.grafana.dashboards['catchpoint-testname-overview.json'].uid)
+        + link.link.options.withKeepTime(true)
+        + link.link.options.withIncludeVars(true),
```
mariaalons, line 13: Here and on line 17 :
`testname-overview` and `nodename-overview` both pin instance and one of `test_name/node_name` to single-select, carrying over a multi-select/All value via withIncludeVars(true) could land on a variable that's only built to hold one value. Would it make sense to drop it on just these two?

mariaalons (approved):
Looks great overall. Left just one comment on the dashboard links 🙇‍♀️

catchpoint-mixin/links.libsonnet
postnati: Agreed. Dropped. Kept on catchpointOverview.

mariaalons (approved):
This one LGTM! thank you

PR #1704, opened 2026-09-10
chore: Modernize istio-2 mixin

postnati:
Modernized the istio-2 mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


istio-2-mixin/alerts.libsonnet
```diff
@@ -1,5 +1,10 @@
 {
   new(this): {
+    // %(filteringSelector)s is intentionally LAST in every matcher list: it is blank
+    // in config.libsonnet and only injected by the integration, and a blank selector
+    // in any earlier position would render a leading comma, which is a fatal PromQL
+    // parse error. A trailing comma is valid PromQL.
+    local istiodPodFilter = 'pod=~"istiod.*"',
     local reporterSourceFilter = 'reporter="source"',
```
mariaalons, line 8: These filters reporterSourceFilter, grpcResponseStatusErrorFilter are already in signals/selectors.libsonnet was there any reason not to import that instead of redeclaring them here?

mariaalons (approved):
Looks great overall I left just one non-blocking comment 🙇‍♀️

istio-2-mixin/alerts.libsonnet
postnati: Good catch. These came over from the original alerts file. Alerts now import signals/selectors.libsonnet instead. istiodPodFilter stays local because the selectors version uses a dashboard variable, and alert rules can't use that. The rendered alerts are unchanged.

mariaalons (commented):
Thank you for addressing the comments 🙇  

I didn't saw this on the previous review @postnati Similar to what came up on the vsphere PR (#1706) there are a few hand-written by() clauses hardcode job, cluster instead of using groupLabels. 
All six rules in alerts.libsonnet, plus the aggLevel: 'none' signals in services.libsonnet, workloads.libsonnet and controlplane.libsonnet. Could you build these from this.groupLabels like on #1706 ?

postnati:
Good catch. Done.

PR #1705, opened 2026-09-10
chore: Modernize apache-solr mixin

postnati:
Modernized the apache-solr mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


apache-solr-mixin/signals/jvm.libsonnet
```diff
@@ -0,0 +1,229 @@
...
+      },
+      garbageCollectionTime: {
+        name: 'Garbage collection time / $__interval',
+        nameShort: 'GC time',
+        type: 'raw',
+        description: 'Total time spent in garbage collection.',
+        unit: 's',
+        sources: {
+          prometheus: {
+            expr: 'avg by (job, solr_cluster, base_url, item) (increase(solr_metrics_jvm_gc_seconds_total{%(queriesSelector)s}[$__interval:] offset -$__interval) / clamp_min(increase(solr_metrics_jvm_gc_total{%(queriesSelector)s}[$__interval:] offset -$__interval), 1)) > 0',
```
Dasomeone, line 39: For these handwritten `by ()` expressions, please remember to use `groupLabels` so it properly respects the given job/cluster/etc labels when enabled (i.e. when using `enableMultiCluster` in config

apache-solr-mixin/config.libsonnet
```diff
@@ -1,26 +1,44 @@
...
-    alertsWarningMemoryUsage: 85,
-    alertsCriticalMemoryUsage: 75,
-    alertsWarningCacheUsage: 75,
-    alertsWarningCoreErrors: 15,
-    alertsWarningDocumentIndexing: 30,
+  // alerts thresholds
+  alertsCriticalCPUUsage: 85,  // %, ApacheSolrHighCPUUsageCritical fires when the 5m average system CPU load across the cluster's nodes stays above this for 5m.
+  alertsWarningCPUUsage: 75,  // %, ApacheSolrHighCPUUsageWarning fires when the 5m average system CPU load across the cluster's nodes stays above this for 5m.
+  alertsWarningMemoryUsage: 85,  // %, ApacheSolrHighHeapMemoryUsageWarning fires when JVM heap used over heap max stays above this for 5m.
+  alertsCriticalMemoryUsage: 75,  // %, ApacheSolrHighHeapMemoryUsageCritical fires when JVM heap used over heap max stays above this for 5m. NOTE: inherited from the legacy mixin, this sits below the warning threshold, so critical fires before warning.
```
Dasomeone, line 19: Looks like an inherited bug here. Surely we'd want critical to fire at 85, and warning at 75?

apache-solr-mixin/panels.libsonnet
```diff
@@ -0,0 +1,464 @@
...
+        // Not .percentage: its stylize() adds decimals=1 and gradientMode='scheme'
+        // and would fight the explicit threshold steps below.
+        + g.panel.timeSeries.standardOptions.withUnit('percent')
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withMax(100)
+        + g.panel.timeSeries.standardOptions.color.withMode('continuous-BlYlRd')
+        + g.panel.timeSeries.standardOptions.thresholds.withSteps([
+          { color: 'green' },
+          { color: 'yellow', value: 90 },
+          { color: 'red', value: 80 },
```
Dasomeone, line 105: Flipped values

apache-solr-mixin/panels.libsonnet
```diff
@@ -0,0 +1,464 @@
...
+
+      topCPULoadByNode:
+        signalsCluster.jvm.topCpuLoad.withTopK('$k').asTimeSeries()
+        + commonlib.panels.generic.timeSeries.base.stylize()
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withMax(100)
+        + g.panel.timeSeries.standardOptions.color.withMode('continuous-BlYlRd')
+        + g.panel.timeSeries.standardOptions.thresholds.withSteps([
+          { color: 'blue' },
+          { color: 'yellow', value: 90 },
```
Dasomeone, line 323: Another flipped threshold

apache-solr-mixin/panels.libsonnet
```diff
@@ -0,0 +1,464 @@
...
+
+      topHeapMemoryUsageByNode:
+        signalsCluster.jvm.heapMemoryUsage.withTopK('$k').asTimeSeries()
+        + commonlib.panels.generic.timeSeries.base.stylize()
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withMax(100)
+        + g.panel.timeSeries.standardOptions.color.withMode('continuous-BlYlRd')
+        + g.panel.timeSeries.standardOptions.thresholds.withSteps([
+          { color: 'blue' },
+          { color: 'yellow', value: 90 },
```
Dasomeone, line 335: +flipped

apache-solr-mixin/panels.libsonnet
```diff
@@ -0,0 +1,464 @@
...
+        + commonlib.panels.generic.timeSeries.base.stylize(),
+
+      topCacheHitRatioByNode:
+        signalsCluster.query.topCacheHitRatio.withFilteringSelectorMixin(coreFilter).withExprWrappersMixin(['bottomk($k,', ')']).asTimeSeries()
+        + commonlib.panels.generic.timeSeries.base.stylize()
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withMax(100)
+        + g.panel.timeSeries.standardOptions.thresholds.withSteps([
+          { color: 'green' },
+          { color: 'yellow', value: 90 },
```
Dasomeone, line 360: This one is actually worse as 100% hit rate should be green?

apache-solr-mixin/panels.libsonnet
```diff
@@ -0,0 +1,464 @@
...
+          'Cache hit ratio',
+          targets=[signals.query.cacheHitRatio.withFilteringSelectorMixin(coreFilter).asTarget()],
+          description='The cache hit ratio for various cache activities.',
+        )
+        + g.panel.timeSeries.standardOptions.withUnit('percent')
+        + g.panel.timeSeries.standardOptions.withMin(0)
+        + g.panel.timeSeries.standardOptions.withMax(100)
+        + g.panel.timeSeries.standardOptions.thresholds.withSteps([
+          { color: 'green' },
+          { color: 'red', value: 80 },
```
Dasomeone, line 244: Same thing here, cache hit ratio so the thresholds should be flipped with red at the bottom surely

apache-solr-mixin/panels.libsonnet
Dasomeone, line 1: I think some of the ones I've called out here are leftovers from the old mixin, specifically around cache hit thresholds where they're flipped. I didn't comment on all of them, but would appreciate if you could give a second pass on all the thresholds and if they correctly represent the signal in question

apache-solr-mixin/alerts.libsonnet
Dasomeone, line 1: Double checked here, and five of the alerts duplicate the signals. We have `asRuleExpression()` that can be used here.

E.g. `heapMemoryUsage`, `cpuLoad`, `cacheHitRatio`, `coreErrors` and `updateHandlerAdds`

You'll get more safety by using `asRuleExpression` as it doesn't risk drift

apache-solr-mixin/signals/jvm.libsonnet
```diff
@@ -0,0 +1,229 @@
...
+        aggLevel: 'group',
+        aggFunction: 'avg',
+        sources: {
+          prometheus: {
+            expr: '100 * solr_metrics_jvm_os_cpu_load{%(queriesSelector)s, item="systemCpuLoad"}',
+            aggKeepLabels: ['base_url', 'solr_cluster'],
+            legendCustomTemplate: '{{base_url}}',
+          },
+        },
+      },
```
Dasomeone, line 214: Directly clones `cpuLoad` and does the same as it. `cpuLoad` already retains the cluster variable, and you can optionally define the filter with `withExprWrappersMixin` in the panels (already used a few places)

Claude flagged the following as also being duplicates:
`jvm.heapMemoryUsage` vs `jvm.memoryHeapUsed`, `query.topUpdateHandlerAdds` vs `query.updateHandlerAdds`, `query.topCacheHitRatio` vs `query.cacheHitRatio`.

Dasomeone (changes requested):
Couple comments on this one, and I noticed that there's a typo in the `.lint` file that renders the `target-instance-rule` non-functional.

Can you update `.lint` line 43 to `entries:` instead of `entires:` please?

apache-solr-mixin/signals/jvm.libsonnet
postnati: Fixed. Both raw signals now use avg by (%(agg)s). Same class of bug in cluster.shardState/replicaState (no by-clause at all), fixed those too.

apache-solr-mixin/config.libsonnet
postnati: Hahaha. Wow. Great catch. Thanks AI.  Agreed, swapped to warning 75 / critical 85

apache-solr-mixin/panels.libsonnet
postnati: Fixed

apache-solr-mixin/panels.libsonnet
postnati: Fixed

apache-solr-mixin/panels.libsonnet
postnati: Fixed

apache-solr-mixin/panels.libsonnet
postnati: Done. The five you flagged were the five that were wrong. Live nodes, Running shards and Running replicas were already correct.

apache-solr-mixin/alerts.libsonnet
postnati: Adopted for both heap alerts. Left the other four hand-written, they're not equivalent to the signals. Noticed signal exprs had the selector first, all moved to last.

apache-solr-mixin/panels.libsonnet
postnati: Fixed

apache-solr-mixin/panels.libsonnet
postnati: Fixed

apache-solr-mixin/signals/jvm.libsonnet
postnati: Removed duplicates.

solr_cluster moved to groupLabels, which was the reason the variants existed.

heapMemoryUsage is is bytes used and memoryHeapUsed is used/max percentage. Kept both.

apache-solr-mixin/alerts.libsonnet
```diff
@@ -57,33 +57,31 @@
           },
           {
             alert: 'ApacheSolrHighHeapMemoryUsageCritical',
-            expr: |||
-              100 * sum without(item, base_url)(solr_metrics_jvm_memory_heap_bytes{item="used", %(filteringSelector)s}) / clamp_min(sum without(item, base_url)(solr_metrics_jvm_memory_heap_bytes{item="max", %(filteringSelector)s}), 1) > %(alertsCriticalMemoryUsage)s
-            ||| % this.config,
+            expr: this.signals.jvm.heapMemoryUsage.asRuleExpression()
```
Dasomeone, line 60: Apologies, you'll have to revert this as the `aggLevel` setting is disregarded when the `by()` statement on the signal is manually authored, so the aggregations make it through.

E.g. here where both `job` and `solr_cluster` make it through
```
expr: 100 * avg by (job,solr_cluster,base_url) (sum without(item)(solr_metrics_jvm_memory_heap_bytes{item="used", }) / clamp_min(sum without(item)(solr_metrics_jvm_memory_heap_bytes{item="max", }), 1)) > 85```

File: apache-solr-mixin/prometheus_rules_out/prometheus_alerts.yaml [hunk 10]
```diff
@@ -31,20 +31,18 @@ groups:
...
             severity: critical
         - alert: ApacheSolrHighHeapMemoryUsageWarning
           annotations:
             description: |
-                {{$labels.instance}} on cluster {{$labels.solr_cluster}} has had high memory usage of {{ printf "%.0f" $value }}%, which is above the thresold of 85.
+                {{$labels.base_url}} on cluster {{$labels.solr_cluster}} has had high memory usage of {{ printf "%.0f" $value }}%, which is above the thresold of 75.
             summary: High heap memory usage can lead to garbage collection issues, out-of-memory errors, and overall system instability.
-          expr: |
-            100 * sum without(item, base_url)(solr_metrics_jvm_memory_heap_bytes{item="used", }) / clamp_min(sum without(item, base_url)(solr_metrics_jvm_memory_heap_bytes{item="max", }), 1) > 85
+          expr: 100 * avg by (job,solr_cluster,base_url) (sum without(item)(solr_metrics_jvm_memory_heap_bytes{item="used", }) / clamp_min(sum without(item)(solr_metrics_jvm_memory_heap_bytes{item="max", }), 1)) > 75
```
Dasomeone, line 45: job is making it through unfortunately due to the aggLevel setting not getting respected with manually authored aggregations in signals.

Apologies, hadn't realised my two comments would collide on this, so this one will need to be manually authored.

apache-solr-mixin/alerts.libsonnet
Dasomeone, line 27: Claude flagged this, and I tend to agree, especially given we've mentioned this previously in the `.lint`: ""base_url is used instead of instance because of how cluster metrics are returned.""

Leftover issue from base mixin, because the instance represents the exporter, rather than the nodes, you end up summing by a factor of 1, so the critical might get tripped early.

Not 100% up to date here, so if you disagree or find it wrong please let me know, but might be worth considering the proper calculation here?

Dasomeone (commented):
Two related comments and one new finding, otherwise LGTM here

apache-solr-mixin/alerts.libsonnet
postnati: Did some digging. Definitely looks like a latent bug where the panel use base_url to differentiate nodes. sum without (base_url, item) collapsed every node into one group.

apache-solr-mixin/alerts.libsonnet
postnati: Alright. Reverted.

apache-solr-mixin/prometheus_rules_out/prometheus_alerts.yaml
postnati: Alright. Reverted.

apache-solr-mixin/prometheus_rules_out/prometheus_alerts.yaml
```diff
@@ -12,37 +12,39 @@ groups:
...
-            100 * sum without (base_url, item) (avg_over_time(solr_metrics_jvm_os_cpu_load{item="systemCpuLoad", }[5m])) > 75
+            100 * sum without (item) (avg_over_time(solr_metrics_jvm_os_cpu_load{item="systemCpuLoad", }[5m])) > 75
           for: 5m
           labels:
             severity: warning
         - alert: ApacheSolrHighHeapMemoryUsageCritical
           annotations:
             description: |
-                {{$labels.base_url}} on cluster {{$labels.solr_cluster}} has had high memory usage of {{ printf "%.0f" $value }}%, which is above the thresold of 85.
+                {{$labels.instance}} on cluster {{$labels.solr_cluster}} has had high memory usage of {{ printf "%.0f" $value }}%, which is above the thresold of 85.
```
Dasomeone: You're flipping the instance/base_url here, is this intended? (I.e. is instance the right denominator here?)

Dasomeone (commented):
One question, but otherwise good to go

apache-solr-mixin/prometheus_rules_out/prometheus_alerts.yaml
postnati: Yeah. I'm stupid and get it backwards in my head. Too much withouts and such. Needed to remove the without base_url on that alert but KEEP it in the label. Got a warning that the label wasn't rendering and took the backwards action. Fixed.

PR #1706, opened 2026-09-10
chore: Modernize vsphere mixin

postnati:
Modernized the vsphere mixin. Added signals architecture, moved panel layout into rows, updated to grafonnet v11.4.0, and used commonlib variables to build dashboards. Did not add or remove any panels or dashboards.


vsphere-mixin/signals/selectors.libsonnet
```diff
@@ -0,0 +1,78 @@
...
+  local virtualMachineOptionalLabels = ['vcenter_cluster_name', 'vcenter_resource_pool_inventory_path', 'vcenter_virtual_app_inventory_path'];
+  // config.libsonnet promises filteringSelector applies to panel queries too, not just to
+  // the dashboard variables. Every selector below is built from a non-empty label list, so
+  // filteringSelector is appended LAST, with the blank default it contributes nothing and
+  // can never produce a leading or doubled comma.
+  local withFilter(selector) =
+    if this.filteringSelector != '' then selector + ',' + this.filteringSelector else selector;
+  {
+    queriesSelector:
+      withFilter(utils.labelsToPromQLSelector(groupLabels + datacenterLabels)),
```
Dasomeone, line 49: This is non-functional, as the `queriesSelector` is now only used for templated variables, and doesn't get used at all in queries due to all the overrides here. 

Verified it locally, so can you please have a look and ensure that filteringselector is *always* correctly injected if set?

vsphere-mixin/dashboards.libsonnet
Dasomeone, line 118: Minor nit here, annotations aren't actually applied and get trapped. Not a huge thing, as it's not used currently but could you update the comment to call out annotations not currently being piped (or add the piping)

vsphere-mixin/panels.libsonnet
```diff
@@ -1,236 +1,183 @@
...
-          fieldOverride.byName.new('Cluster')
-          + table.fieldOverride.byName.withProperty('links', [
+        + g.panel.table.standardOptions.withNoValue('NA')
+        + g.panel.table.standardOptions.withOverridesMixin([
+          g.panel.table.fieldOverride.byName.new('Cluster')
+          + g.panel.table.fieldOverride.byName.withProperty('links', [
             {
               title: '',
-              url: 'd/vsphere-clusters?var-datasource=${datasource}&${__all_variables}&var-vcenter_cluster_name=${__value.raw}&${__url_time_range}',
+              url: 'd/' + uid + '-clusters?var-datasource=${datasource}&${__all_variables}&var-vcenter_cluster_name=${__value.raw}&${__url_time_range}',
```
Dasomeone, line 107: I think your links like this throughout are broken? They don't match the format we use for the links setup for the logs-lib I.e.

```
link.link.new('vSphere overview', '/d/' + this.grafana.dashboards[uid + '-overview.json'].uid)
```

So even if this works right now, it risks becoming broken in the future with any rename, so could you please reuse the dynamic generation?

Dasomeone (commented):
Couple comments, mostly with the complex query selectors which ignore the config and panel/dashboard setup. Output otherwise looks fine

vsphere-mixin/signals/selectors.libsonnet
postnati: Not sure what is going on. I'm unable to reproduce. Maybe you tested on a slightly old version (or more likely, I'm missing something)? This was broken prior to commit c759423, but on the latest, if I put in a filteringSelector, I'm seeing it get appended to every one of selectors in selector.libsonnet

Output I'm getting with no filteringSelector
"expr": "vcenter_cluster_vm_count{power_state=\"on\", job=~\"$job\", … ,vcenter_cluster_name=~\"$vcenter_cluster_name\"}"

Output I'm getting with a filteringSelector
"expr": "vcenter_cluster_vm_count{power_state=\"on\", job=~\"$job\", … ,vcenter_cluster_name=~\"$vcenter_cluster_name\",job=\"integrations/vsphere\"}"

I verified the filter reaches 98/98 panel targets and 5/5 alert rules. Restructured so it's applied once over the whole object rather than per-field, so a new selector can't miss it.

vsphere-mixin/dashboards.libsonnet
postnati: Good catch. Updated.

vsphere-mixin/panels.libsonnet
postnati: Agreed. All four now resolve via this.grafana.dashboards. Much cleaner. Rendered output unchanged.

vsphere-mixin/signals/selectors.libsonnet
Dasomeone: I'll re-run locally to try this again

Dasomeone: Yep works now! Must've had an out of date branch or messed something up last I tried

Dasomeone (commented):
All open Qs have been answered thank you!

One new thing that shouldn't be too bad, I hope, i noticed there's a couple manually authored group `by()`s throughout (specifically cluster queries like clusterCPUUtilization) that don't use the `groupLabels`. Could you do a small update like you did for apache solr for those? 🙏 

Otherwise LGTM

postnati:
Updated the by()s

postnati:
the by() clauses now build from groupLabels plus the datacenter/cluster/host lists. Avoiding doing anything else because of the apach-solr revert. Should I dig deeper?

Dasomeone (approved):
Thanks @postnati , that's perfect!

